Issue #32, opened 2022-09-18
Introduce a higher speed serializer option


jeremydmiller:
Do some research. Should do before 1.0 to verify the abstraction

jeremydmiller:
Honestly, this is going to be a documentation effort only. Not shipping anything right now.

jeremydmiller:
I want to try to use [MemoryPack](https://github.com/Cysharp/MemoryPack) here. Looks easy to use, perf numbers are off the charts. It'd need to be done in a separate extension Nuget to avoid taking the dependency.

## Tasks

- [ ] Implementation of a MemoryPack version of Wolverine's `IMessageSerializer`
- [ ] An extension method on `WolverineOptions` to register that serializer globally
- [ ] Extension methods to add that serializer to listeners and/or subscribers?
- [ ] Test it end to end where it has to go over the wire. Could be just over TCP

jeremydmiller:
Done with the MemoryPack option!

PR #141, opened 2023-01-05
issue #32 - implement MemoryPack serialization 

Vertygo:
[Introduce a higher speed serializer option #32](https://github.com/JasperFx/wolverine/issues/32) implementation. This PR contains :
*  MemoryPack version of Wolverine's IMessageSerializer
* Created an extension method on WolverineOptions to register that serializer globally
* Created end to end tests over the TCP
* Added nuget publishing definition to build.cs


jeremydmiller:
@Vertygo I like the look of things so far. I would ask you to just rename the new package to "Wolverine.MemoryPack" though. No need for the deeply nested namespace.

The call to `Serialize(Envelope)` needs to delegate to serializing the `Envelope.Message`. There were some special cases for MassTransit or NServiceBus interop that had to do something special, but not here.

Biggest thing left is just having an end to end test. I'd say you can even spin up a host that publishes with the TCP transport and have it go over the wire to itself.

jeremydmiller:
@Vertygo I know we talked about the per endpoint configuration options, but I'm good to pull this in any time if you want.

Vertygo:
Go ahead and merge it.  I can add per endpoint configuration in separate PR.

PR #164, opened 2023-01-30
Transactional attribute no longer required when using EF tx

wojteksuwala:
According to the docs, the [Transactional] attribute should not be required to enroll in transactional middleware. I added a call to options.Handlers.AutoApplyTransactions() to UseEntityFrameworkCoreTransactions. I adjusted the unit tests and sample code.
I modified auto apply to skip sagas as I am not sure if it should be treated this way (if not omitted all EF Integration Saga) tests fail.


jeremydmiller:
@wojteksuwala I think I want to keep this as an opt in behavior for now, and I adjusted the docs today to reflect that.

Is it okay if we don't take this one in?

wojteksuwala:
Sure no problem

Issue #195, opened 2023-02-13
The CLR type Wolverine.Envelope isn't natively supported by Npgsql or your PostgreSQL.


mus-h:
I'm trying to add marten support for a durable outbox, using the below configuration (wolverine V0.9.8):
```
                    services.AddMarten(opts =>
                    {
                        opts.Connection(connectionString!);
                    })
                   .IntegrateWithWolverine();
```
```
builder.UseWolverine(opts =>
            {
                opts.PersistMessagesWithPostgresql(connectionString!);
                opts.PublishAllMessages().ToPort(5581).UseDurableOutbox();
            });
```

The messages persisted in the DB but: 
- the "deliver_by" is null.
- the attempts counters do not change.
- when the connection is restored, resend of pending messages is not working unless the application is restarted.

ERROR message:
System.NotSupportedException: The CLR type Wolverine.Envelope isn't natively supported by Npgsql or your PostgreSQL. To use it with a PostgreSQL composite you need to specify DataTypeName or to map it, please refer to the documentation.
   at Npgsql.TypeMapping.TypeMapper.ResolveByClrType(Type type)
   at Npgsql.TypeMapping.TypeMapper.ResolveByClrType(Type type)
   at Npgsql.NpgsqlParameter.ResolveHandler(TypeMapper typeMapper)
   at Npgsql.NpgsqlParameterCollection.ProcessParameters(TypeMapper typeMapper, Boolean validateValues, CommandType commandType)
   at Npgsql.NpgsqlCommand.ExecuteReader(CommandBehavior behavior, Boolean async, CancellationToken cancellationToken)
   at Npgsql.NpgsqlCommand.ExecuteReader(CommandBehavior behavior, Boolean async, CancellationToken cancellationToken)
   at Npgsql.NpgsqlCommand.ExecuteNonQuery(Boolean async, CancellationToken cancellationToken)
   at Weasel.Core.CommandExtensions.ExecuteOnce(DbCommand command, CancellationToken cancellation)
   at Weasel.Core.CommandExtensions.ExecuteOnce(DbCommand command, CancellationToken cancellation)
   at Wolverine.Persistence.Durability.DurableSendingAgent.<>c__DisplayClass14_0.<<enqueueForRetryAsync>b__4>d.MoveNext()
--- End of stack trace from previous location ---
   at Wolverine.Transports.Sending.SendingAgent.executeWithRetriesAsync(Func`1 action)

jeremydmiller:
[...]

PR #196, opened 2023-02-13
Fix the CLR type issue in PostgreSQL

aliseyedali:
#195 
for all ids in query should be used WithEnvelopeIds


jeremydmiller:
@aliseyedali I had just barely managed to track that down and added a test for the specific fix you made. **Really** with I'd checked the PR tab first:-).

PR #201, opened 2023-02-19
Changes name of 'parsed' variable in CodeGen to account for multiple params

Shield1739:
If you use multiple nullable parameters like

`
    [WolverineGet("/todoitems/{id}")]
    public static Task<Todo?> GetTodo(
        int id,
        int? int1,
        int? int2,
        IQuerySession session, 
        CancellationToken cancellation) 
        => session.LoadAsync<Todo>(id, cancellation);
`

You get the error that the 'parsed' variable already exists in this context. I've changed the name of the parsed variable to include the name of the variable it is parsing to ensure that the name is not repeated.


jeremydmiller:
Nice fix, I'll have this up on Nuget within a couple days

PR #257, opened 2023-03-20
WIP: Adds Postgres as a Transport

PascalSenn:


jeremydmiller:
@PascalSenn Hey, can you slow down a bit? I'm absolutely buried in other work right now, but I can tell you from a quick glance that I'm going to request several significant changes. This duplicates a **lot** of existing Wolverine code, does other things like schema setup in a very different way than the rest of Wolverine, and won't work OOTB with Wolverine's durable outbox.

PascalSenn:
@jeremydmiller  for context: this pull request was implemented the postgres transport in a similar way to the Azure Service Bus implementation just with some postgres specficis. It uses the same API surface, so it might look like duplicate code. I'm not sure what you mean, otherwise with duplicated code.
 
In any case, there's no rush to get this pull request in. I'll hold off until we've had a chance to chat about how we want to proceed.

PascalSenn:
I will come back to this

Issue #342, opened 2023-04-29
Describe command fails


Richard87:
Hi, I'm just playing with Wolverine (and Martendb), trying to get WolverineGet to work, while running the `describe` command, this happens:

<details><summary>Details</summary>
<p>

```
dotnet run -- describe
Building...
/Users/.../WeatherForecast.cs(12,19): warning CS8618: Non-nullable property 'City' must contain a non-null value when exiting constructor. Consider declaring the property as nullable. [/Users/r.../example.csproj]
Searching 'JasperFx.CodeGeneration.Commands, Version=0.0.0.0, Culture=neutral, PublicKeyToken=null' for commands
Searching 'Weasel.CommandLine, Version=6.0.0.0, Culture=neutral, PublicKeyToken=null' for commands
Searching 'Marten.CommandLine, Version=0.0.0.0, Culture=neutral, PublicKeyToken=null' for commands
Searching 'Wolverine, Version=0.9.17.0, Culture=neutral, PublicKeyToken=null' for commands

No Wolverine extensions are detected

── Wolverine Options ──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────

Service Name: example
├── Assemblies
│   └── example (application)
├── Extensions
│   └── Wolverine.Postgresql.PostgresqlBackedPersistence
├── Serializers
│   └── ┌──────────────────┬──────────────────────┐
│       │ Content Type     │ Serializer           │
│       ├──────────────────┼──────────────────────┤
│       │ binary/envelope  │ EnvelopeReaderWriter │
│       │ application/json │ NewtonsoftSerializer │
│       └──────────────────┴──────────────────────┘
├── DefaultExecutionTimeout
│   └── 00:01:00
├── AutoBuildEnvelopeStorageOnStartup
│   └── False
├── TypeLoadMode
│   └── Dynamic
└── ExternalTransportsAreStubbed
    └── False


── Wolverine Handlers ─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
Handler Discovery Rules
├── Assemblies
│   └── example
├── Handler Type Rules
│   ├── Include:
│   │   ├── Name ends with 'Handler'
│   │   ├── Name ends with 'Consumer'
│   │   ├── Inherits from Wolverine.Saga
│   │   ├── Implements Wolverine.IWolverineHandler
│   │   ├── Has attribute Wolverine.Attributes.WolverineHandlerAttribute
[...]

PR #343, opened 2023-04-29
Update WolverineRuntime.Descriptions.cs

Richard87:
changes [/gray] to [/]

Fixes https://github.com/JasperFx/wolverine/issues/342

Untested, not sure how to test it, any guidance would be amazing!


jeremydmiller:
Honestly? This is the very rare occasion where I'm perfectly good with just pulling it in. I don't think this is anything that'd be likely to be broken later.

And thank you!

PR #366, opened 2023-05-30
Fixes broken Rabbit exchange checks.

ThickPropheT:
Fixes https://github.com/JasperFx/wolverine/issues/367.

Adds test coverage for `RabbitMqExchange` environment checks.

Changes `exchangeName` in Rabbit MQ `end_to_end.use_fan_out_exchange()` test from `"fanout"` to `"fanout1"` to avoid future collisions with the `fanout` exchange type.

Modifies `RabbitMqExchange` to reference exchanges by name rather than by type.


jeremydmiller:
Got it, thanks!

PR #428, opened 2023-06-23
Docs Spelling Fix

ardalis:
Fix spelling issue


ardalis:
Also why is there only one command listed when it says there will be two?

image

PR #437, opened 2023-06-26
Listen events from proper connection in RabbitMqTransport

### LINKED ISSUE #434: RabbitMq transport. Possibly _listenerConnection listen twice
Hi,
Could you look at this part of RabbitMqTransprort?
https://github.com/JasperFx/wolverine/blob/main/src/Transports/RabbitMQ/Wolverine.RabbitMQ/Internal/RabbitMqTransport.cs#L84
```
if (_listenerConnection == null)
{
    _listenerConnection = BuildConnection();
    listenToEvents("Listener", _listenerConnection, logger);
}
if (_sendingConnection == null)
{
    _sendingConnection = BuildConnection();
    listenToEvents("Sender", _listenerConnection, logger);
}
```

Looks like it should be `_sendingConnection` as a parameter for listenToEvents in second if statement.

swimtver:
Fixed #434


jeremydmiller:
Good catch. This will be in 1.0.1 soon (maybe today?)

PR #456, opened 2023-07-04
Fix log message of BrokerResource Setup() errors

Xzelsius:
Fixes wrong log message of `BrokerResource.Setup()`  errors


jeremydmiller:
Good catch, and thank you!

PR #459, opened 2023-07-04
Fix Azure Service Bus topic and subscription auto provisioning

Xzelsius:
Adds support for `AutoProvision` on `AzureServiceBusTopic` and `AzureServiceBusSubscription` endpoints.

### DIFF
--- a/src/Transports/Azure/Wolverine.AzureServiceBus/Internal/AzureServiceBusSubscription.cs
+++ b/src/Transports/Azure/Wolverine.AzureServiceBus/Internal/AzureServiceBusSubscription.cs
@@ -146,11 +146,13 @@ public class AzureServiceBusSubscription : AzureServiceBusEndpoint, IBrokerQueue
         _hasInitialized = true;
     }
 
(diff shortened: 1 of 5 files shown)


src/Transports/Azure/Wolverine.AzureServiceBus/Internal/AzureServiceBusSubscription.cs
```diff
@@ -125,4 +127,19 @@ public async ValueTask PurgeAsync(ILogger logger)
...
+        if (_hasInitialized)
+        {
+            return;
+        }
+
+        if (Parent.AutoProvision)
+        {
+            await SetupAsync(logger);
+        }
+
```
Xzelsius, line 142: @jeremydmiller Not sure whether you want support for `AutoPurge` in here too?

e.g.
```
if (Parent.AutoPurgeAllQueues)
{
    await PurgeAsync(logger);
}
```

jeremydmiller: I think we do. I'll add that real quick after I pull this in.

PR #460, opened 2023-07-05
added target clean to build

BWalti:
As I had a stale version of this git repository on my machine, just pulling the newest version and trying to build resulted in some strange errors, therefore I wanted to clean out all the bin/obj directories - which I manually did.

Afterwards everything was working.

Thought maybe a clean target might be worth as well?


jeremydmiller:
I go draconian and use "git clean -xfd", but that wipes out the ignored connection.txt files for some of the test harnesses I think

PR #482, opened 2023-07-19
Azure Service Bus use topics with conventional routing

solvingproblemswithtechnology:
Hi!

This PR brings a new RoutingConvention for ASB, bringing the ability of discovering and using Broadcast through Topics.

It adds AzureServiceBusQueueAndTopicMessageRoutingConvention and the method  UsePublishingBroadcastFor to activate it (otherwise, is completely retrocompatible).

Also a new overload of UseConventionalRouting, making the previous one Obsolete with the idea of removing it in a following Major version. By default, the existing one is the one being used unless you explicitily use the new overload.

Last, I duplicated the tests to have them for the existing and new RoutingConvention, also with the idea of deleting them in the next Major.

I followed the Expand and Contract principle (also known as Parallel Changes) so that's why I prefer to duplicate, add the new functionality and later on remove the previous implementations instead of refactoring the existing code in one go.


jeremydmiller:
@solvingproblemswithtechnology Couple comments:


* I was a little surprised to see 30 files changing. I was expecting maybe 3. The new convention class, a test fixture for the same, and an addition to the fluent interface for Azure Service Bus registration. 
* Please don't change naming conventions in pull requests. The purposeful naming is to use Pascal casing for `public` or `internal` members, but Camel casing for `private` or `protected`. Please change the naming back the way it was before.
* I'm not a fan of introducing the separate namespaces for "New" and "Existing". That's not going into main that way. The way I see this, we're creating a completely separate alternative here. We will not be retiring the existing, purely queue based conventional approach

"Also a new overload of UseConventionalRouting, making the previous one Obsolete with the idea of removing it in a following Major version. By default, the existing one is the one being used unless you explicitily use the new overload." -- I don't think we communicated very well, because I thought I explicitly said that I didn't want to do this.

jeremydmiller:
@solvingproblemswithtechnology,

Thanks for the changes, sorry I was slow getting back to you.

How are you feeling about this going in? I can help drive the tests to completion & the doc updates.

Oh, and are you still tackling the "execute the conventions before doing resource setup"? If you aren't I can knock that out. I have a client who's going to need that soon.

solvingproblemswithtechnology:
Yes, if you could help with the tests fine, I couldn't discover what happened with the last ones, the discovery wasn't working for reasons. I can do the docs if you tell me how they work or where they are.

jeremydmiller:
Hey, I'm going to cut yet another little 1.5 release today, but absolutely get this into 1.6. Let's call Monday? I'll pull this in and push through the tests, then ask you for a separate PR for docs maybe. Sorry this has lingered so long.

solvingproblemswithtechnology:
No prob! I'll do the docs. Also, If you can tell me what was failing, I'm all ears ^^

Issue #451, opened 2023-06-30
Filter OpenTelemetry events


agross:
This is a screenshot of the Jaeger UI after starting the app with a previously empty Jaeger database.

![image](https://github.com/JasperFx/wolverine/assets/16245/eb169aab-57c5-4322-b5f0-80dcb22b3f68)

After a couple of minutes the outlier to the left contains around 150 spans related to background operations. I could get rid of them by filtering spans with the opentelemetry-collector:

```yaml
processors:
  filter/wolverine-internal:
    traces:
      span:
        - 'IsMatch(attributes["messaging.message_type"], "^Wolverine\\.RDBMS\\..*")'
        - 'IsMatch(attributes["messaging.message_type"], "^Wolverine\\.Runtime\\.Agents\\..*")'
        - 'IsMatch(attributes["db.statement"], "wolverine_messages")' # These are Npgsql statements issued by the background job.
```

Perhaps it's not necessary to log those spans in the first place.

jeremydmiller:
@agross I'm going to have to push this one off until maybe next week. Or at least it's not going in 1.1.3 today.

jeremydmiller:
And I'm pushing this off yet again since there is a workaround.

jeremydmiller:
## Notes

* This PR has a solid start to filtering: https://github.com/JasperFx/wolverine/pull/541
* Might be more mechanically suitable to do the filtering at the endpoint level rather than by message type
* Latch at endpoint for sending/receiving. Can latch at message type for sending too.
* Could latch activity tracing in HandlerPipeline. That handles execution
* Latch in Receiver too
* Use different IExecutor or IExecutor decorators for message type filtering on execution. Remove the tracing from InvokeInlineAsync? -- *Nah, just use a boolean flag*
* Consider latching the audit logging on internal messages too

jeremydmiller:
## Tasks

- [ ] Configure telemetry enabled by endpoint
- [ ] Enable/disable telemetry on all listener types
- [ ] Enable/disable telemetry on all sending agent types
- [ ] Disable telemetry on database control queues
- [ ] Disable telemetry on all agent queues and internal actions

PR #541, opened 2023-09-11
Added configurable activity tracing

mfuqua3:
This pull request addresses issue #451 

It adds configuration support for Wolverine tracing, and adds an additional package for OpenTelemetry instrumentation.

- Adds configuration support for activity filtering predicates for the three supported Wolverine activity types
- Adds configuration support for global activity filtering, regardless of origin type
- Add support of enrichment of Wolverine generated activities
- Adds first class support for filtering out internal message types
- Adds additional package for OpenTelemetry instrumentation

```csharp
Host.CreateDefaultBuilder()
            .UseWolverine(options =>
            {
                // Suppress activity generation for IInternalMessage implementations
                options.Tracing.SuppressInternalMessageTypes = true; 
                options.Tracing.Enrich = (activity, eventType, envelope) =>
                {
                    // enrich the activity externally
                };
                options.Tracing.SendEnvelopeFilter = (envelope) =>
                {
                    //specify a condition in which activities shouldn't be traced for sent messages
                    return true;
                };
                options.Tracing.GlobalFilter = (envelope) =>
                {
                    //Specify a condition in which activities shouldn't be traced for any message
                    return true;
                };
            })
```

-- OR --

```csharp
Host.CreateDefaultBuilder()
            .UseWolverine()
            .ConfigureServices(services =>
            {
                services.AddOpenTelemetry()
                    .WithTracing(builder =>
                    {
                        builder
                            .SetResourceBuilder(ResourceBuilder.CreateDefault().AddService("MY SERVICE"))
                            .AddWolverineInstrumentation(options =>
                            {
                               // Suppress activity generation for IInternalMessage implementations
                                options.SuppressInternalMessageTypes = true; 
                                options.Enrich = (activity, eventType, envelope) =>
                                {
                                    // enrich the activity
                                };
                                options.SendEnvelopeFilter = (envelope) =>
                                {
                                    //specify a condition in which activities shouldn't be traced for sent messages
                                    return true;
                                };
                                options.GlobalFilter = (envelope) =>
                                {
                                    //specify a condition in which activities shouldn't be traced for any messages
                                    return true;
                                };
                            });
                    });
            });
```

### DIFF
--- a/build/build.cs
+++ b/build/build.cs
@@ -151,6 +151,7 @@ namespace build
                 "./src/Extensions/Wolverine.FluentValidation",
                 "./src/Extensions/Wolverine.MemoryPack",
                 "./src/Extensions/Wolverine.MessagePack",
+                "./src/Extensions/Wolverine.OpenTelemetry",
                 "./src/Http/Wolverine.Http",
                 "./src/Http/Wolverine.Http.FluentValidation",
             };
--- a/src/Extensions/Wolverine.OpenTelemetry/OpenTelemetryInstrumentationServiceCollectionExtensions.cs
+++ b/src/Extensions/Wolverine.OpenTelemetry/OpenTelemetryInstrumentationServiceCollectionExtensions.cs
@@ -1,0 +1,27 @@
+﻿using Microsoft.Extensions.DependencyInjection;
+using Wolverine.Runtime.Tracing;
+
+namespace Wolverine.OpenTelemetry;
+
+internal static class OpenTelemetryInstrumentationServiceCollectionExtensions
+{
+    public static void AddTelemetryExtension(this IServiceCollection services, Action<WolverineTracingOptions> configure)
+    {
+        checkConfigMarker(services);
+        var telemetryExtension = new TelemetryWolverineExtension(configure);
+        services.Add(new ServiceDescriptor(typeof(IWolverineExtension), telemetryExtension));
+    }
+
+    private static void checkConfigMarker(IServiceCollection services)
+    {
+        var marker = services.FirstOrDefault(s => s.ServiceType == typeof(TelemetryExtensionConfigMarker));
+        if (marker == null)
+        {
+            services.AddSingleton(new TelemetryExtensionConfigMarker());
+            return;
+        }
+
+        throw new InvalidOperationException(
+            "Wolverine instrumentation has already been registered and doesn't support multiple registrations.");
+    }
+}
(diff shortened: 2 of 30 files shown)


mfuqua3:
resolves #451

jeremydmiller:
@mfuqua3 Hey, Matt, I'm going to get this in soon, but with a couple minor changes. I'm going to introduce a "composite filter" pattern so you can register separate includes/excludes. Experience says you'd want that and those one size fits all Func<something, bool> is never perfectly usable.

The other minor things I'm going to do is pull this:

```
            if (!WolverineActivitySource.Options.GetFilterForActivity(spanName).Invoke(envelope))
            {
                logger.LogTrace("Execution of envelope {EnvelopeId} is filtered out from activity tracing", envelope.Id);
                return null;
            }
```

into each individual Start**** method just to eliminate the lookup. It's a micro-optimization, but you work under the assumption that everything a framework does is the hot path.

Question for you, how do you see the `Enrich()` being used? My intention was that folks would mostly use middlware or even explicit code in handlers for that (plus that's already supported in no small part because of repetitive code I saw in your giant monolith).

mfuqua3:
> The other minor things I'm going to do is pull this into each individual Start**** method just to eliminate the lookup. It's a micro-optimization, but you work under the assumption that everything a framework does is the hot path.

Makes sense. I actually went back and forth initially on whether the code reuse was worth it, so I'm with you there.

> Question for you, how do you see the Enrich() being used?

So I do not necessarily have an *immediate* need for this capability. I actually borrowed this concept (and most of the contract for this options class) from another OpenTelemetry instrumentation package -- the WCF one to be specific. I do not anticipate I would use this feature right away, if ever. Although I can anticipate use-cases for it:

1. I figure it provides an extensibility hook for Activity tagging from the Core library that could be utilized by other Wolverine extensions, especially if they aren't first-party extensions that might have access to the `internal` API.
2. I think it provides an easy way to add interop tags as a Wolverine configuration, or to rename tags to match legacy names so Wolverine generated traces can be included in existing trace queries (maybe I need the tenant tag to be named `tenancy` instead of `tenant.id` for example)

jeremydmiller:
@mfuqua3 Hey Matt, I did take a very close look at this, but decided to do the Otel filtering (and more logging filtering in general) a bit different way -- mostly out of micro-optimization thinking. I will revisit this in 1.8 and use the new extension that adds Otel to Wolverine. That one's a winner. I'll also try to do 1.8 much sooner than the gap was between 1.6 and 1.7

mfuqua3:
Roger that, sounds good!

Issue #554, opened 2023-09-22
Sending in addition to publishing Messages from HTTP Endpoints


erdtsieck:

jeremydmiller:
Closed by https://github.com/JasperFx/wolverine/issues/554

PR #555, opened 2023-09-22
#554 Sending in addition to publishing messages from http endpoints

erdtsieck:
[#554](https://github.com/JasperFx/wolverine/issues/554)


jeremydmiller:
I'll do the rebase locally in just a second

PR #576, opened 2023-10-02
Add conditional in status code setting and logging service, with new …

carcer:
…endpoint and tests

Introduced a check in EmptyBody204Policy.cs to ensure that the http response has not started before setting the status code to 204. This prevents any modifications from being attempted on a response that has already begun and can reduce unexpected behavior.

Included a logging service in the Program.cs file to allow for better tracking of application messages and help in debugging.

Two new files were added, 'using_write_by_id.cs' and 'WriteToEndpoints.cs', to create a new Marten-based endpoint ('/write-to/{id}') along with its corresponding integration test. This aims to enhance the API's functionality by allowing retrieval of issues by their unique ID.

refs #575


jeremydmiller:
Cool, good addition. This will be in 1.8.0 soon-ish

PR #615, opened 2023-10-26
docs: fix info blocks for asp net core integration

timdeschryver:
It seems like the blocks are case sensitive.

BEFORE:

image

AFTER:

image

The other changes are auto-formatted, but I can revert them if needed.


jeremydmiller:
Thanks!

PR #624, opened 2023-11-04
feat: add auditing to http endpoints

timdeschryver:
Sorry for not creating an issue first, feel free to close this PR if you want me to do that first or if the current behavior is intended.

I noticed audited members are not added to `Activity` when using endpoint handlers.
This PR uses the existing `AuditToActivityFrame` to append the audited tags to the `Activity` (that's why the access modifier is changed from `internal` to `public`). 

Side note: I had to comment the following lines to run the API and the tests, because these resulted in `NullReference` exceptions (not included in this PR) https://github.com/JasperFx/wolverine/blob/main/src/Http/Wolverine.Http.Marten/DocumentAttribute.cs#L32 and  https://github.com/JasperFx/wolverine/blob/main/src/Http/Wolverine.Http.Marten/AggregateAttribute.cs#L59


src/Http/Wolverine.Http.Tests/endpoint_adds_requesttype_audit_tags_to_activity.cs
```diff
@@ -0,0 +1,29 @@
...
+        
+        chain.AuditedMembers.Single()
+            .MemberName.ShouldBe(nameof(AuditablePostBody.Id));
+    }
+    
+    [Fact]
+    public void adds_auditable_members_to_activity()
+    {
+        var chain = HttpChains.ChainFor("POST", "/auditable/post");
+        var lines = chain.SourceCode.ReadLines();
```
timdeschryver, line 26: This test currently fails because I can't get the `SourceCode` to be initialized.

PR #631, opened 2023-11-13
Tweaks test to better test if IEvent is handled

carcer:
I couldn't seem to get the changes made in https://github.com/JasperFx/wolverine/commit/c70402ea047e91e9e791bd56e2ef2c681c34c64e work as expected.  I extended the test case to check if the IEvent handler method was being executed (filthy check, i know), and it appears that it may not actually be called.


jeremydmiller:
I think I left that out on purpose for the sake of time to be honest, but I've got more time today to wrestle with that now

carcer:
Ah my mistake, with the commit including an IEvent<T> handler method, thought support had been added.  Understood

Issue #649, opened 2023-12-07
Dynamic code generation fails occasionaly on a list messages that use a Marten aggregate and thus trigger code generation in parallel


erdtsieck:
**Describe the bug**
I have handlers A B C on aggregate one and D handling on aggregate two. That gives me errors such as:

```cs

System.AggregateException
One or more errors occurred. (Compilation failures!

CS0103: The name 'CreateDefault' does not exist in the current context
CS0103: The name 'Apply' does not exist in the current context

Code:

// <auto-generated/>
#pragma warning disable
using Marten;
using Marten.Events.Aggregation;
using Marten.Internal.Storage;
using System;
using System.Linq;

namespace Marten.Generated.EventStore
{
    // START: SingleStreamProjectionLiveAggregation68506063
    public class SingleStreamProjectionLiveAggregation68506063 : Marten.Events.Aggregation.SyncLiveAggregatorBase<Consulteren.Features.AutorisatiesOpPatienten.GebruikerPatientConsultatieToegang>
    {
        private readonly Marten.Events.Aggregation.SingleStreamProjection<Consulteren.Features.AutorisatiesOpPatienten.GebruikerPatientConsultatieToegang> _singleStreamProjection;

        public SingleStreamProjectionLiveAggregation68506063(Marten.Events.Aggregation.SingleStreamProjection<Consulteren.Features.AutorisatiesOpPatienten.GebruikerPatientConsultatieToegang> singleStreamProjection)
        {
            _singleStreamProjection = singleStreamProjection;
        }



        public override Consulteren.Features.AutorisatiesOpPatienten.GebruikerPatientConsultatieToegang Build(System.Collections.Generic.IReadOnlyList<Marten.Events.IEvent> events, Marten.IQuerySession session, Consulteren.Features.AutorisatiesOpPatienten.GebruikerPatientConsultatieToegang snapshot)
        {
            if (!events.Any()) return null;
            Consulteren.Features.AutorisatiesOpPatienten.GebruikerPatientConsultatieToegang gebruikerPatientConsultatieToegang = null;
            var usedEventOnCreate = snapshot is null;
            snapshot ??= Create(events[0], session);;
            if (snapshot is null)
            {
                usedEventOnCreate = false;
                snapshot = CreateDefault(events[0]);
            }

            foreach (var @event in events.Skip(usedEventOnCreate ? 1 : 0))
            {
                snapshot = Apply(@event, snapshot, session);
            }

            return snapshot;
        }


[...]

PR #657, opened 2023-12-13
#649 reproduction compilation failures on cascading aggregate handler

erdtsieck:
This is not the fix. Bug only occurs when you run locally on a multi-core processor


erdtsieck:
@jeremydmiller do you think you could try if you get this reproduced locally by running this test a few times? 

I got a workaround running everything sequentially in my tests, but that does make my integration tests a lot slower :-(

erdtsieck:
Created a PR in the Marten codebase https://github.com/JasperFx/marten/issues/2900

PR #678, opened 2024-01-10
use char instead of strings in GetPrettyName

SimonCropp:


jeremydmiller:
@SimonCropp Thank you, not sure this method is actually used anywhere though. There's a more robust `FullNameInCode()` extension method in JasperFx.Core that probably gets used much more

PR #682, opened 2024-01-11
fix case of OpenApiDemonstrator.csproj path

SimonCropp:


SimonCropp:
seems it is still failing since the build is hard coded to target net7 and OpenApiDemonstrator.csproj only targets net8

```
<Project Sdk="Microsoft.NET.Sdk.Web">

    <PropertyGroup>
        <!-- Leave this alone -->
        <TargetFrameworks>net8.0</TargetFrameworks>
```

SimonCropp:
should the build to changed to target net8? or can OpenApiDemonstrator.csproj be changed to also target net7?

PR #693, opened 2024-01-19
Fix incorrect Http verb in WolverineOptionsAttribute

haefele:
I just randomly stumbled over this and thought "Hey, I can fix this real quick".


jeremydmiller:
@haefele Thanks!

PR #699, opened 2024-01-21
Removed duplicates in Executor

swimtver:
Removed `LogError` call to avoid logging error twice & removed duplicates in ctors logic.


jeremydmiller:
Thank you!

PR #724, opened 2024-02-11
Add support for `IEndpointConventionBuilder.Finally` and fix status-code 200 in `PublishingEndpoint`

haefele:
During testing, the `PublishingEndpoints` still displayed the OpenAPI documentation for status-code 200, although it should have been removed in `PublishingEndpoint<T>.Configure`.

The issue seems to come from the timing of the `Configure` method; it is called before the Wolverine default metadata is added  to the builder.

Specifically, `Configure` is executed within the `HttpChain` constructor in `applyAttributesAndConfigureMethods`.
However, the default metadata is added later during endpoint construction in `establishResourceTypeMetadata`. 
Because of that, it's not possible to remove status-codes 200 or 404 in the `Configure` method because they are added afterward.

It seems that the ASP.NET Core team ran into the same problem and introduced a `IEndpointConventionBuilder.Finally` method in .NET 7 that allows metadata modifications after the standard processing.

This pull request implements this `Finally` method, and fixes the `PublishingEndpoint` class to correctly remove status-code 200.

This also allows users to remove status-code 404 if they want to.

### DIFF
--- a/docs/guide/http/metadata.md
+++ b/docs/guide/http/metadata.md
@@ -44,13 +44,18 @@ public static void Configure(HttpChain chain)
     // a request directly to a Wolverine messaging endpoint for later processing
     chain.Metadata.Add(builder =>
     {
-        // Adding and modifying data
+        // Adding metadata
         builder.Metadata.Add(new WolverineProducesResponseTypeMetadata { StatusCode = 202, Type = null });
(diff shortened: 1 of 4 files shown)


jeremydmiller:
Sweet!

PR #734, opened 2024-02-19
Fixed PollForScheduledMessages in SqlServerMssageStore

swimtver:
Hi,
There is a execution of wrong command in `PollForScheduledMessagesAsync` method.

Could you help me with test?  
It seems that instead of call `PollForScheduledMessagesAsync` I should wait when durability agent execute it in background.
I didn't figure out how to do it in a proper way.


jeremydmiller:
@swimtver I'll take the test today. Not 100% sure w/o some thought about how to do that, but I'll get this in regardless

PR #737, opened 2024-02-19
Update RabbitMqListener.cs to include the ClientProvidedName

gizmohd:
Send the ClientProvidedName as provided in the connection to the consumer so that consumer connections show the ClientProvidedName as well.


jeremydmiller:
@gizmohd Cool, and thank you! This will be in 1.19 today I think

PR #757, opened 2024-02-29
Fix optional side effect

woksin:
Adds the possibility to return "optional" side effect from message handlers


woksin:
Was thinking whether the test should actually check that the side effect execute method was called when the side effect was returned

PR #847, opened 2024-04-27
Fix some usages of HttpChain.RequestType after filling it with System.Void as default value

haefele:
I noticed that [this change](https://github.com/JasperFx/wolverine/blob/844cd97810a391b3cff882e27dbc8d2d680dfce5/src/Http/Wolverine.Http/HttpChain.cs#L178-L180) lead to some changes in the OpenAPI generation (see my change in `HttpChain.ApiDescription`).
It leads to endpoints without a request having a "made up" request with type `System.Void`, which in turn trips up Swashbuckle.

To me, it seems like that wasn't the intention with the `RequestType ??= typeof(void)` so I made these adjustments.

I also checked the other usages of the `RequestType` property and tried to adjust them if needed.

### DIFF
--- a/src/Http/Wolverine.Http/HttpChain.ApiDescription.cs
+++ b/src/Http/Wolverine.Http/HttpChain.ApiDescription.cs
@@ -174,7 +174,7 @@ public partial class HttpChain
 
     private void fillRequestType(ApiDescription apiDescription)
     {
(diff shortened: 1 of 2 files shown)


jeremydmiller:
@haefele Sigh, and thank you!

PR #852, opened 2024-05-06
add generic WolverineModule

SimonCropp:
thoughts on a generic (type safe) WolverineModuleAttribute?


jeremydmiller:
I think it makes sense. Thanks Simon!

PR #862, opened 2024-05-08
avoid linq .Count() when property can be used

SimonCropp:


jeremydmiller:
Got it w/ a manual rebase

Issue #868, opened 2024-05-12
 Postgres Transport using wrong schemas for message storage and transport queues when configured with marten integration


Shield1739:
**Describe the bug**

I'm not sure if the title explained it, but here are some code samples of how i found the error.

I was using my app with the marten integration, when i decided to start publishing to postgres queues.

```
builder.UseWolvrine((context, options) =>
{
    // only publish no listening
    options.PublishMessage<RandomMessage>()
        .ToPostgresqlQueue(RandomQueueName)
        .UseDurableOutbox();
}

services.AddMarten(options =>
    {
        options.Connection(repriSyncConnectionString);
        options.DatabaseSchemaName = "messages"
    }).IntegrateWithWolverine()
```

This creates tables on 2 schemas.

1. 'messages': With the message storage (control nodes, incoming envelopes, etc..,)
2. 'wolverine_queues': With the tables for the "RandomQueueName" queue

This was working fine for me while i was only publishing, but when i started to listen to the queue by addding:

```
options.ListenToPostgresqlQueue(RandomQueueName)
        .UseDurableInbox()
        .MaximumParallelMessages(5);

```

I got the following error:
```
Error while trying to retrieve messages from PostgreSQL Queue random_queue
Npgsql.PostgresException (0x80004005): 42P01: relation "wolverine_queues.wolverine_incoming_envelopes" does not exist
```

![image](https://github.com/JasperFx/wolverine/assets/18101480/b29ca9f3-cdf4-41b2-b45d-7a297db60821)

I tried a couple other configurations using the IntegrateWithWolverine **schemaName** and **transportSchemaName** paremetrs, but to no avail.

I was able to change the schema where the transport created the queues with **transportSchemaName**, but it would still try to search for the wolverine_incoming_envelopes tables in that same schema, when in fact it is created under the name passed by **schemaName**. (hopefully u got that).

if you use the 

`options.UsePostgresqlPersistenceAndTransport(connectionString, schemaName)`

There isnt an issue because it uses the same schema for both things, I would prefer to be able to use 2 separate schemas to keep things a bit more tidy.

I've made a (possible) fix for this in a PR. ( https://github.com/JasperFx/wolverine/pull/869 )

PR #869, opened 2024-05-12
Separates postgre transport and message storage schemas

Shield1739:
The problem is better explained in the issue it fixes https://github.com/JasperFx/wolverine/issues/868 

But the solution i got was to store the transport and message storage schemas into 2 different fields. That way you are able to configure them separately via the wolverine integration. 

I also added an extra parameter to the use postgresql transport option to be able to configure it through that as well.

I would appreciate if anyone could double check the changes.


jeremydmiller:
It's technically a breaking API change (new optional arguments are a breaking change to a public API), but I'm going to be naughty and allow it. I don't think there's enough people using this that this would be an issue yet.

PR #873, opened 2024-05-13
Add optional RouteUrl parameter to Wolverine endpoint attributes

flew2bits:
This update adds an optional second parameter to all of the Wolverine* endpoint attributes which is used to add a RouteNameMetadata instance to the endpoint definition. This allows applications to find the route path by name through the use of the IUrlHelper.RouteUrl() method.


jeremydmiller:
Got it. I'm going to convert the constructor arg to a property though. It's a breaking change the way it is now even as an optional argument

PR #905, opened 2024-05-24
Decoupling the AutoCreate in Message Databases from Marten StoreOptions

Xzelsius:
With this change it's possible to have Wolverine automatically create the message storage for dynamically added tenants, irrelevant of the `AutoCreate` setting in Marten.

Previously, if one used `AutoCreateSchemaObjects = AutoCreate.None` Wolverine did start up the agents for the new tenant, but they just failed because the required schema objects for the message storage were not available.

The default remains `AutoCreate.CreateOrUpdate` (see `DatabaseSettings`) for all `MessageDatabase` implementations.
But for Marten integration, you now can specify `autoCreate: <mode>` on the call to `IntegrateWithWolverine(...)`. If not specified, it has a fallback to `StoreOptions.AutoCreateSchemaObjects`.


Xzelsius:
@jeremydmiller Maybe you have a better idea to test these changes :)

PR #915, opened 2024-05-30
Implement an indefinite scheduled retry policy

woksin:
Implements a policy for indefinitively retrying message processing

### DIFF
--- a/src/Testing/CoreTests/ErrorHandling/FailureRuleCollectionTests.cs
+++ b/src/Testing/CoreTests/ErrorHandling/FailureRuleCollectionTests.cs
@@ -259,6 +259,32 @@ public class FailureRuleCollectionTests
         theHandlers.Failures.DetermineExecutionContinuation(new BadImageFormatException(), theEnvelope)
             .ShouldBeOfType<MoveToErrorQueue>();
     }
+    
+    [Fact]
+    public void schedule_indefinite_retries()
+    {
+        theHandlers.OnException<BadImageFormatException>().ScheduleIndefiniteRetry(1.Minutes(), 5.Minutes(), 10.Minutes());
+
+        theEnvelope.Attempts = 0;
+        theHandlers.Failures.DetermineExecutionContinuation(new BadImageFormatException(), theEnvelope)
+            .ShouldBeOfType<ScheduledRetryContinuation>().Delay.ShouldBe(1.Minutes());
+
+        theEnvelope.Attempts = 1;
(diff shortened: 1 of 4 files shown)


src/Testing/CoreTests/ErrorHandling/FailureRuleCollectionTests.cs
```diff
@@ -259,6 +259,32 @@ public void schedule_retry_n_times()
...
+        theHandlers.Failures.DetermineExecutionContinuation(new BadImageFormatException(), theEnvelope)
+            .ShouldBeOfType<ScheduledRetryContinuation>().Delay.ShouldBe(5.Minutes());
+
+        theEnvelope.Attempts = 3;
+        theHandlers.Failures.DetermineExecutionContinuation(new BadImageFormatException(), theEnvelope)
+            .ShouldBeOfType<IndefiniteScheduledRetryContinuation>().Delay.ShouldBe(10.Minutes());
+
+        theEnvelope.Attempts = 4;
+        theHandlers.Failures.DetermineExecutionContinuation(new BadImageFormatException(), theEnvelope)
+            .ShouldBeOfType<MoveToErrorQueue>();
```
woksin, line 286: Just to comment, this section here is a bit confusing because the failure slots list will be expanded when the IndefiniteSchedluledRetryContinuation is executed, thus the scenario when `attempts == 4` could be a bit misleading in terms of the text, maybe :)

src/Wolverine/ErrorHandling/PolicyExpression.cs
```diff
@@ -169,6 +169,26 @@ public IAdditionalActions ScheduleRetry(params TimeSpan[] delays)
...
+        {
+            throw new InvalidOperationException("You must specify at least one delay time");
+        }
+        cancellationToken.ThrowIfCancellationRequested();
+        
+        for (var i = 0; i < delays.Length; i++)
+        {
+            var continuation = new ScheduledRetryContinuation(delays[i]);
+            var slot = i == delays.Length - 1 ? _rule.AddSlot(new IndefiniteScheduledRetryContinuation(continuation, _rule, cancellationToken)) : _rule.AddSlot(continuation);
+            _slots.Add(slot);
```
woksin, line 186: Question to you, @jeremydmiller. If this solution with continuously adding itself as a slot to the rule is an ok solution, would it make sense to add it this _slots list as well? I've had look how this FailureActions thing works, but I don't think that I understand it fully. So there probably is a better way to implement this :)

PR #916, opened 2024-05-30
Docs: fix broken link

AlexZeitler:


jeremydmiller:
Doh! That's an old find and replace fail

PR #918, opened 2024-05-31
Docs: Add white background on wolverine-subscriptions.png

joelving:
Exceptionally tiny contribution, but nonetheless: The diagram on https://wolverine.netlify.app/guide/durability/marten/subscriptions.html is currently all but unreadable in dark mode due to a transparent background.
This PR fixes that by adding a white background to the image.

For good measure, I ran the file through PNGGauntlet reducing file size to almost half.


jeremydmiller:
@joelving Dude, I'm not good at this kind of thing, so it's very much appreciated from my side. Thank you!

Issue #872, opened 2024-05-13
Enhance `DocumentAttribute` to Respect Soft Deleted Documents


StephenStrickland:
**Is your feature request related to a problem? Please describe.**

Yes, currently, the `DocumentAttribute` in Wolverine does not differentiate between active and soft deleted documents. When using Marten's soft delete feature, `DocumentAttribute` utilizes `IDocumentSession.LoadAsync()`, which loads documents regardless of deletion status. This behavior is problematic in scenarios where we need to exclude soft deleted documents from being loaded, leading to potential issues in handling and displaying data that should be considered inactive.

**Describe the solution you'd like**

I propose adding a new parameter, `bool MaybeDeleted`, to the `DocumentAttribute`. When set to `false`, the `DocumentAttribute` should modify its behavior to exclude soft deleted documents. Specifically, instead of using `IDocumentSession.LoadAsync()`, it should utilize `IDocumentSession.QueryAsync()` so that the soft deletion status is respected, such as:

```csharp
if (MaybeDeleted)
{
    return session.LoadAsync<T>(id);
}
else
{
    return session.Query<T>().Where(x => x.Id == id).FirstOrDefaultAsync();
}
```
This change ensures that only active documents are loaded when `MaybeDeleted` is set to `false`, providing better control over data integrity and visibility.

**Describe alternatives you've considered**

An alternative to modifying the `DocumentAttribute` would be to manually implement the logic to check for soft deleted documents in every handler that needs this behavior. This could be achieved by creating a method like the following:

```csharp
public static async Task<(MyDocument? myDoc, IResult result)> LoadAsync([FromRoute] Guid id, IDocumentSession session)
{
    var myDoc = await session.LoadAsync<MyDocument>(id);
    
    return myDoc != null && !myDoc.Deleted
            ? (myDoc, WolverineContinue.Result()) 
            : (myDoc, Results.NotFound()); 
}
```

This method would individually handle document loading and checking the Deleted status before deciding the result to return. Although this ensures that soft deleted documents are appropriately managed, it requires duplication of code across various parts of the application where document loading is necessary. This approach increases the risk of inconsistencies and errors, as each implementation might slightly vary or could be mistakenly omitted in new developments.

**Additional context**

This enhancement aligns with best practices in handling soft deleted records, ensuring that our APIs can differentiate between active and deleted entities without additional overhead for developers.

PR #923, opened 2024-06-04
Enhance [Document] attribute with support for soft-deletes

haefele:
Added DocumentAttribute.MaybeSoftDeleted flag as a way to indicate whether soft-deleted documents should be handled as NULL or not.

The naming of the property might not be the best, but it's the best succinct one I could come up with.
In the discord-discussion linked from #872 it seemed like an enum was suggested, but that would make it very long to type.
If you want to use it with `Required = true` it already gets quite long.
Updated the docs to include this new property.

Also added a test for a unexpected side-effect of using `[Document][Required]` instead of `[Document(Required = true)]`.
Added a warning for this to the docs as well.

While I was already editing some docs, I also documented the "new" before-methods introduced in #855 (they are amazing).

### DIFF
--- a/docs/guide/handlers/middleware.md
+++ b/docs/guide/handlers/middleware.md
@@ -129,7 +129,7 @@ Here's the conventions:
 
 | Lifecycle                                                | Method Names                |
 |----------------------------------------------------------|-----------------------------|
-| Before the Handler(s)                                    | `Before`, `BeforeAsync`, `Load`, `LoadAsync` |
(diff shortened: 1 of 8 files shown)


haefele:
Oh, it's the first time I wrote a code-gen `Frame` in `SetVariableToNullIfSoftDeletedFrame` - maybe some good eyes on that. 
I wasn't 100 % sure what I was doing.

Issue #926, opened 2024-06-06
Using `IFormFile` in both an endpoint and its middleware causes code generation error


forrestab:
**Describe the bug**
Running into an issue creating a file upload endpoint with file validation middleware. If both the endpoint and middleware have an `IFormFile` parameter the code generation fails with the following error.

>CS0128: A local variable or function named 'file' is already defined in this scope

**To Reproduce**
From a simple web api project I have the following endpoint and middleware classes:

```csharp
public static class FileUploadEndpoint
{
    [Middleware(
        typeof(FileLengthValidationMiddleware), 
        typeof(FileExtensionValidationMiddleware)
    )]
    [WolverinePost("/api/files/upload")]
    public static async Task<Ok> HandleAsync(IFormFile file)
    {
        // todo, generate filename, write mapping to table
        // todo, create sideeffect to write file with new filename

        return TypedResults.Ok(); // return new filename
    }
}

public static class FileLengthValidationMiddleware
{
    public static void Before(IFormFile file, IOptions<FileUploadOptions> options)
    {
        // todo, return ProblemDetail if validation fails
    }
}

public static class FileExtensionValidationMiddleware
{
    public static void Before(IFormFile file, IOptions<FileUploadOptions> options)
    {
        // todo, return ProblemDetail if validation fails
    }
}
```

Just in case its relavent, this is my wolverine configuration (nothing special):

```csharp
// ConfigureServices
builder.Host.UseWolverine(options =>
{
    options.Durability.Mode = DurabilityMode.MediatorOnly;
    options.OptimizeArtifactWorkflow();
});

// ConfigurePipeline
app.MapWolverineEndpoints();
```

**Expected behavior**
Be able to use the `IFormFile` request parameter in both middleware leading up to and in the endpoint.

**Additional context**
For the above repro classes this is the code that is generated (seeing it returned with a 500 when the error is thrown):

```csharp
// START: POST_api_files_upload
public class POST_api_files_upload : Wolverine.Http.HttpHandler
{
    private readonly Wolverine.Http.WolverineHttpOptions _wolverineHttpOptions;
    private readonly Microsoft.Extensions.Options.IOptions<FileUpload.Server.FileUploadOptions> _options;

    public POST_api_files_upload(Wolverine.Http.WolverineHttpOptions wolverineHttpOptions, Microsoft.Extensions.Options.IOptions<FileUpload.Server.FileUploadOptions> options) : base(wolverineHttpOptions)
    {
[...]

PR #942, opened 2024-06-19
Bug 926 also fix OpenAPI generation

StellaAlexis:
Noticed after the fix of bug [926](https://github.com/JasperFx/wolverine/issues/926) that the code generation works fine but the OpenAPI generation is broken. 
Based on the tests this change should fix it.


jeremydmiller:
OpenAPI is just a plot to torture anyone who is foolish enough to try to work on a custom http api framework.

Thanks, and I'll get this out tonight with the next release!

Issue #818, opened 2024-04-10
Balanced durability mode doesn't work with RabbitMQ (attempt 2)


ArkadiuszChorian:
This issue is direct continuation of #788.

Verified wolverine v2.2.0 which have included PR #791 aim to fix this but it still doesn't work on [repro repo](https://github.com/ArkadiuszChorian/wolverine-balanced-durability-issue-788).

I would appreciate further investigation on this @jeremydmiller

jeremydmiller:
@ArkadiuszChorian I'm sorry, but your make script didn't work for me, and I just don't have enough bandwidth to trouble shoot whatever is going on with your docker setup. You'd neglected to mention -- unless I completely missed that -- that your issue came from publishing an app to docker because this works all day long just running the process. 

I'll take one more shot at it if you can feed me the exact stack trace after Wolverine 2.4 hits later this evening with just a little more in the exception message that might explain what I'm guessing is a race condition of some sort.

And Wolverine is a community OSS project that accepts pull requests, so feel free to dig in if you can. Or my bandwidth weirdly goes up if you engage w/ JasperFx Software.

ArkadiuszChorian:
@jeremydmiller Unfortunately it's not the happy ending for my project but I understand your limited bandwidth. Despite that I'll try to follow up your last comment.

First of all
> your issue came from publishing an app to docker because this works all day long just running the process

That would be good clue but unfortunately it's not the case here. We prepared repro repo using `docker` and `make` only to automate whole setup for you but this issue is still present without docker (aka native development).

Now this part
> I'll take one more shot at it if you can feed me the exact stack trace after Wolverine 2.4 hits later this evening with just a little more in the exception message that might explain what I'm guessing is a race condition of some sort.

I updated Wolverine to 2.5 and below are the logs (app run without docker). Maybe it'll somehow trigger an idea in your mind.
```
info: Wolverine.Runtime.Agents.NodeAgentController[0]
      Starting agents for Node d9b91381-d801-48ef-a88e-51644065d494 with assigned node id 1
info: Wolverine.Runtime.Agents.NodeAgentController[0]
      Found no other existing nodes, deciding to assume leadership in node 1
fail: Wolverine.Runtime.WolverineRuntime[0]
      Failed to start the Wolverine messaging
      System.Reflection.TargetInvocationException: Exception has been thrown by the target of an invocation.
[...]

Issue #788, opened 2024-03-25
Balanced durability mode doesn't work with RabbitMQ


ArkadiuszChorian:
**Describe the bug**
When using RabbitMQ with default `DurabilityMode = Balanced` app is unable to start.
All Wolverine and Marten packages are latest as for now: Wolverine 2.1.2 and Marten 7.4.0

There are two other issues strictly related to that and giving necessary context:
#645 and #746 

**Firstly** this exception is thrown:
```
fail: Wolverine.Runtime.Agents.StartLocalAgentProcessing[0]
      Invocation failed!
      System.Reflection.TargetInvocationException: Exception has been thrown by the target of an invocation.
       ---> System.ArgumentNullException: Value cannot be null. (Parameter 'serializer')
         at Wolverine.Runtime.Routing.MessageRoute..ctor(IMessageSerializer serializer, ISendingAgent sender, IEnumerable`1 rules, IReplyTracker replyTracker)
         at Wolverine.Runtime.Routing.MessageRoute..ctor(Type messageType, Endpoint endpoint, IReplyTracker replies)
         at Wolverine.Runtime.Routing.MessageRouterBase`1.<>c__DisplayClass2_0.<.ctor>b__1(Endpoint endpoint)
         at System.Linq.Enumerable.WhereSelectArrayIterator`2.ToArray()
         at Wolverine.Runtime.Routing.MessageRouterBase`1..ctor(WolverineRuntime runtime)
         at Wolverine.Runtime.Routing.MessageRouter`1..ctor(WolverineRuntime runtime, IEnumerable`1 routes)
         at System.RuntimeMethodHandle.InvokeMethod(Object target, Void** arguments, Signature sig, Boolean isConstructor)
         at System.Reflection.MethodBaseInvoker.InvokeDirectByRefWithFewArgs(Object obj, Span`1 copyOfArgs, BindingFlags invokeAttr)
         --- End of inner exception stack trace ---
         at System.Reflection.MethodBaseInvoker.InvokeDirectByRefWithFewArgs(Object obj, Span`1 copyOfArgs, BindingFlags invokeAttr)
         at System.Reflection.MethodBaseInvoker.InvokeWithFewArgs(Object obj, BindingFlags invokeAttr, Binder binder, Object[] parameters, CultureInfo culture)
         at System.Reflection.RuntimeConstructorInfo.Invoke(BindingFlags invokeAttr, Binder binder, Object[] parameters, CultureInfo culture)
[...]

PR #952, opened 2024-06-28
Fix RabbitMQ not able to start in Balanced Mode

doctorspider42:
start node agent workflow after starting messaging transport to avoid null agent exception, when rabbit mq is used in balanced mode


doctorspider42:
Solves:
https://github.com/JasperFx/wolverine/issues/818
https://github.com/JasperFx/wolverine/issues/788

PR #969, opened 2024-07-09
Publishing a message without a handler doesn't add it to the tracked session

AlexZeitler:
This PR exposes all states of messages represented via `MessageEventType` as record collections on `ITrackedSession` and it's `TrackedSession` implementation.

It also extends the related docs.

The property `Executed` is obsoleted in favor of two new properties `ExecutionStarted` and `ExecutionFinished` where the latter is replacing the existing `Executed` collection.

### DIFF
--- a/docs/guide/testing.md
+++ b/docs/guide/testing.md
@@ -419,6 +419,11 @@ that is tracked.
 Note that you'll probably *mostly* *invoke* messages in these tests, but there are additional extension
 methods on `IHost` for other `IMessageBus` operations.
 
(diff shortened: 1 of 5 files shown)


AlexZeitler:
I also obsoleted `Executed` as we now provide two `Execution*` properties which are more explicit.

https://github.com/JasperFx/wolverine/pull/969/files#diff-3a23b6e5971bd662b66f03aa14aaa562373fcb6f7d7eb69446d2b91083d84523R67

PR #1013, opened 2024-08-17
Use default postgres public schema when no schema is provided.

NilsLattek:
When setting up postgres using `UsePostgresqlPersistenceAndTransport` without providing a schema name then an error is thrown during startup, because the schema is empty and the table name then starts with a dot `.wolverine_node_records`: https://github.com/JasperFx/wolverine/blob/d478269/src/Persistence/Wolverine.RDBMS/Durability/PersistNodeRecord.cs#L27


jeremydmiller:
I'm sorry this got lost, but we solved this problem a little differently. 2 PRs came in at about the same time, then the mechanics changed in 4.0. Thank you for your contribution!

PR #1080, opened 2024-10-15
GCP Pub/Sub draft

jay-zahiri:


jay-zahiri:
Here's the output, you can see the `Received` record has no message

image

jay-zahiri:
It's weird because I don't do anything that different from Azure in terms of envelope mapping, but I assume its e2e send and receive test worked. Maybe I just write a test envelope mapper that maps out `Message` as well.

PR #1091, opened 2024-10-17
Adds support for tenantId to MartenOps

carcer:
As mentioned in Discord.

**I have left the 2 different ways of applying the tenantId for feedback on which is the prefered way.** I will update the PR accordingly, hence the PR being draft

### DIFF
--- a/src/Persistence/MartenTests/handler_actions_with_implied_marten_operations_with_tenant_switching.cs
+++ b/src/Persistence/MartenTests/handler_actions_with_implied_marten_operations_with_tenant_switching.cs
@@ -1,0 +1,150 @@
+using IntegrationTests;
+using Marten;
+using Marten.Events;
+using Marten.Exceptions;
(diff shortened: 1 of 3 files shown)


DurinMusicspear:
This is a great feature, i have been missing this several times already! Mostly because we are stuck with a system with both Wolverine and non-Woverine services.

jeremydmiller:
@carcer Hey Chris, the code itself looks good, but because I didn't get this in *before* the 3.0 release, I need you to make overload methods instead of the optional parameters. Adding an optional parameter is a breaking API change even though it doesn't change anyone's code because it requires a re-compilation. Picky rules, but SemVer

carcer:
Fair, no problem, I'll get the change made soon 😊 thanks

jeremydmiller:
Closing this PR — thank you for the idea @carcer! We're creating a new issue to implement tenant ID support across all IMartenOp types with a slightly different approach (new overloads rather than changing existing signatures).

PR #1135, opened 2024-11-19
Fix inconsistent tenant/activity test

richardszalay:
The test relied on accessing IHttpActivityFeature after the request completed in order to extract the Activity to assert against.

However, the runtime ([HostingApplication](https://github.com/dotnet/aspnetcore/blob/main/src/Hosting/Hosting/src/Internal/HostingApplication.cs)) actually sets the value of HttpActivityFeature.Activity to null after each request so that the HttpActivityFeature can be recycled.

This commit works around this behavior by copying the Activity into a custom feature.


jeremydmiller:
Nice!

PR #1150, opened 2024-11-23
Add support for routing messages based on attribute decoration

### LINKED ISSUE #1149: Add support for routing messages based on attribute decoration
I'd like to decorate messages I want to wind up in a certain queue with an attribute, rather than use a marker interface.

Something like the following
```csharp
options
    .Publish(rule =>
    {
        rule.MessagesDecoratedWith<MySpecificQueueAttribute>()
            .ToPostgresqlQueue("my-specific-queue");
    });
```

Marker interfaces offer essentially the same functionality as what I'm looking for.

patchoulish:
Closes #1149

I've run the CoreTests unit tests on .NET 7.0/8.0/9.0. All pass.

Otherwise let me know if you want anything altered.

### DIFF
--- a/src/Testing/CoreTests/Runtime/Green/Messages.cs
+++ b/src/Testing/CoreTests/Runtime/Green/Messages.cs
@@ -5,3 +5,5 @@ public class GreenMessage1;
 public class GreenMessage2;
 
 public class GreenMessage3;
+
+public class GreenAttribute : Attribute;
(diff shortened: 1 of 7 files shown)


jeremydmiller:
@patchoulish Looked good, but if there's a next time, could you add an example of that to the docs? I'm doing so now, so no worries if you see this

jeremydmiller:
@patchoulish Actually, sorry, I reverted this. The general approach is fine, but I can't take this without an actual end to end test and preferably some documentation

PR #1259, opened 2025-01-31
Remove log failure to Console from ExtensionLoader

### LINKED ISSUE #1258: Scanning for extensions may write dll paths to console
When my application, which uses SqlClient, starts, I'm getting these mysterious dll paths as the first output on the console:
```
[redacted]\runtimes\win-arm\native\Microsoft.Data.SqlClient.SNI.dll
[redacted]\runtimes\win-arm64\native\Microsoft.Data.SqlClient.SNI.dll
[redacted]\runtimes\win-x64\native\Microsoft.Data.SqlClient.SNI.dll
[redacted]\runtimes\win-x86\native\Microsoft.Data.SqlClient.SNI.dll
```

I think this comes from this code:
https://github.com/JasperFx/wolverine/blob/9bd4175a8cb076e2413b770d26b8f85e45b14fb5/src/Wolverine/ExtensionLoader.cs#L20-L25

After reading the code, I understand why these specific dlls are logged, but I don't think this console output is productive in its current form or at all. As a user of Wolverine, this did not tell me anything helpful and only lead to questioning what was happening. 

I suggest this output is removed.

ArturDorochowicz:
Fixes #1258

### DIFF
--- a/src/Wolverine/ExtensionLoader.cs
+++ b/src/Wolverine/ExtensionLoader.cs
@@ -18,7 +18,7 @@ internal static class ExtensionLoader
         }
 
         _extensions = AssemblyFinder
-            .FindAssemblies(msg => Console.WriteLine(msg), a => a.HasAttribute<WolverineModuleAttribute>(), false)
+            .FindAssemblies(a => a.HasAttribute<WolverineModuleAttribute>(), false)
             .Concat(AppDomain.CurrentDomain.GetAssemblies())
             .Distinct()
             .Where(a => a.HasAttribute<WolverineModuleAttribute>())


jeremydmiller:
@ArturDorochowicz Hey man, I'm sorry, but this is the very epitome of the "solves my tactical problem, doesn't address a strategic concern" pull request. I didn't just put the console output on failures there for fun. What you did is essentially swallowing a potential problem. Like I told you before, you can simply opt out of the automatic extension finding altogether. Or come up with some kind of "suppress assembly warnings" flag I guess.

jeremydmiller:
If you wanna try this a different way that just latches the logging on assembly failures, I'm good to take that. But this just treats the symptoms.

jeremydmiller:
Closing for now — thank you for the contribution @ArturDorochowicz!

PR #1287, opened 2025-02-25
Minor documentation updates

elexisvenator:
Don't use WolverineGet on methods with a request payload - reduce confusion.


src/Http/Wolverine.Http.Tests/DocumentationSamples.cs
```diff
@@ -47,4 +47,4 @@ public IResult Redirect(GoToColor request)
     }
 
     #endregion
-}
\ No newline at end of file
+}
```
elexisvenator, line 50:
```suggestion
}

```

PR #1310, opened 2025-03-15
docs: commit 'mdsnippets' changes

hsnabszhdn:
I intended to contribute to docs, but even before I do anything, I ended up with 66 changes by running `npm run docs`:

With this many changes, it's really hard to focus on the actual change I want to make/suggest.

It could be that `mdsnippets` has released a new version that is case sensitive now :shrug: 

```
$ mdsnippets --version
mdsnippets 27.0.2+1f22752e2764bfbe50a24b4bb424a91bf6dc82c9
Finished 44ms
```

>[!NOTE]
>It would've been better if the version of `mdsnippets` was mentioned in README.md instead of [implying the "latest" version](https://github.com/JasperFx/wolverine/blob/883833c2cf1bb499dd7da3d233f951d0260fdd36/README.md?plain=1#L49C48-L50C90). Otherwise, even if this PR gets merged, someone else with an older version of `mdsnippets` installed would raise another PR reverting all this.


jeremydmiller:
I'll do this independently against the master branch in a bit. Long story, but I don't like to have the mdsnippets changes committed because that's caused loss of content in the past on other projects.

Issue #58, opened 2022-10-17
SNS topic support for the SQS transport


jeremydmiller:
Breaking this out for now.

jeremydmiller:
Wrong issue, this is still open.

kurtwarwick-new:
@jeremydmiller - I would just like to find out about the nearness to delivery for this topic.

As you know from one of my other issues, I am currently using RabbitMq. Our systems, however, call for AWS platforms and thus, AWS SNS. 

Since they rely on event publishing, which is currently not (fully) supported with the SQS-only implementation, and rabbit is so expensive to run in AWS, I am eagerly awating this feature! :) It would so amazing to see this come to a release soon.

Thanks.

Kurt

igor-toporet:
From the article https://jeremydmiller.com/2025/01/06/critter-stack-roadmap-for-2025/ I'm getting AWS SNS support isn't planned for this year.

Shield1739:
Having SNS support would be very beneficial for my team and after looking into this for a bit I've got a question @jeremydmiller.  Do you think the extra effort to keep SNS and SQS under the same transport is worth it?

The `AWSSDK.SimpleNotificationService` package has its own thing going and the current `AmazonSqsTransport` doesn't look flexible to just add the SNS client inside of it without some extra rewriting effort.

Whatever you decide this looks fun so I'll give it a shot

jeremydmiller:
Closed by https://github.com/JasperFx/wolverine/pull/1314

PR #1314, opened 2025-03-17
Adds Amazon SNS Support

Shield1739:
* WIP
This PR is to adds support for Amazon Simple Notification Service  (SNS). Mentioned in the issue #58 

So far my plan is to create a separate transport for SNS instead of adding it to the current SQS support.

So far here are my cons and pros of creating a separate transport:

Pros
* Less abstractions. Some of the code could be reused between the SQS/SNS transports, but due to the AWS SDKs having 2 different packages for SQS and SNS we would need to create some sort of abstraction between the SQS queue and the SNS topic. 

If we were to merge it all into the `AmazonSqsTransport` class, we would need to create a separate field for the Client (`IAmazonSQS` / `IAmazonSimpleNotificationService`), Config (`AmazonSQSConfig` / `AmazonSimpleNotificationServiceConfig`). The queue/topic creation configuration and attributes would also be different for each endpoint.

* Only load the transport that you are actually going to use. Just because you are using SNS means you are using SQS or vice versa.

Cons
* Harder to integrate the two. SNS and SQS can be used completely independently from each other, but of course the do work well together. Mainly SNS give you the ability to subscribe an SQS queue to a SNS topic, so all messages you send to the topic will be then published to this queue. Automating this setup is something I think we could easily implement from within wolverine. BUT by having 2 different transports it makes it a little more complicated to achieve (at least to my peanut brain). 

I had envisioned the ability to create a SNS topic, configure a subscription to a SQS queue and then configure a listener on that same queue all within wolverine. Storing the queues and topics in 2 different transports, just makes this a little more difficult.


Shield1739:
The last pushed commits implements a separate transport for the SNS support.

So far I'm able to successfully create an SNS topic, publish a message to it, create a SQS queue subscription to that topic and listen to the SQS queue. That last part is done using the SQS transport.

The publishing to SNS is very simple to implement, but there are some things to consider. 

### Topic Subscriptions. 

Right now the SNS transport can only "Publish" as there is no listening to an SNS topic. Nevertheless, we can create subscribe an SQS queue to forward the messages sent to the topic. 

For this I added the method:  `SubscribeSqsQueue(string queueName, bool rawMessageDelivery = false)`

Which will automatically create the Topic Subscriptions if _AutoProvision_ is _true_
This is working fine in LocalStack, but for this to work out-of-the-box in AWS we would also need to create a policy in the SQS queue to allow the SNS Topic to send messages to it. 

I added the subscriptions support since doing the compliance testing without the listening part seems to be impossible. 

### SNS Metadata Template

When SNS forwards a message to SQS it adds wraps the message body with some metadata. See more [here](https://docs.aws.amazon.com/sns/latest/dg/sns-large-payload-raw-message-delivery.html). Fortunately you can toggle this behavior in the subscription and send the raw message that was received in the Topic. That's what the _rawMessageDelivery_ option is for in the method I mentioned above. 

We can also add a feature to the SQS transport to parse this SNS Metadata. I did a quick POC and something quick JSON parsing worked. Example:
`var body = JObject.Parse(message.Body)["Message"].ToString()`

### Auto Provisioning

Just a small note, the SNS transport only auto-creates the Topics and Subscriptions, but for a SQS Queue subscriptions to be created the Queue needs to exist first, if someone were to create everything straight from wolverine the queues would need to be created first or the system will throw an exception while trying to create the subscriptions.

Shield1739:
Things added on last commit

* Added `SnsTopicEnvelopeMapper` for the SQS Transport.
* * This mapper is a little unique as it has an "inner" mapper. You can see the **SNS Metadata Template** section on my previous [comment](https://github.com/JasperFx/wolverine/pull/1314#issuecomment-2764657086).
* Adds Batch publish support for the SNS transport. It has the same 10 messages limitation as SQS
* Added more transport tests. The test uses both the SQS and SNS transports. It creates an SNS topic, an SQS queue and a subscription between the two. Then publishes a messages to the SNS topic and listens for it in the SQS queue. I use both the new SnsTopicMapper and the raw message delivery method in the tests.

Finally I'm also working in the compliance testing... but seeing as how this transport doesn't have a listener some tests are not really working

Shield1739:
Finally added some compliance tests with the latest commits. Since the SNS transport doesn't have listeners we had to add some workaround for the some tests and don't work without listeners. 

* Added the `IsSenderOnlyTransport` field in the `TransportComplianceFixture` class. Also added short circuit to the following test when this field is _null_: `can_send_and_wait`, `can_request_reply` & `explicit_respond_to_sender`

The remaining failing compliance tests are: `can_apply_requeue_mechanics`, `requested_response`, `will_move_to_dead_letter_queue_without_any_exception_match`, `will_requeue_and_increment_attempts`.

Shield1739:
Latest commit adds the setting of a policy in the SQS queue to be able to receive messages from the SNS topic after a subcription has been created. I used the same policy as shown in [this example](https://docs.aws.amazon.com/sns/latest/dg/subscribe-sqs-queue-to-sns-topic.html).

One quick note tho, I noticed that the Topic subscriptions are not deleted right away after calling `DeleteTopicAsync`. So if someone were to create a topic with a subscription, then call `AmazonSnsTopic.TeardownAsync` and then recreate the topic and subscription again you would end up with 1 topic and 2 subscriptions.

It seems that these subscriptions will be deleted automatically after 3 days. (see [here](https://stackoverflow.com/questions/71763300/aws-snsclient-deletetopic-doesnt-delete-subscriptions)). This does feel like a small edge case, since I would assume people would delete things manually and not through the Endpoint Teardown methods. So we would leave it like that or add a call to `UnsubscribeAsync` for each subscription inside `TeardownAsync`

I do think this needs a little more work and refining before we can merge it, but I'll be marking it as _Ready for review_ since I believe enough of it is complete that it's worth it for people to start looking into it.

Issue #1320, opened 2025-03-20
WolverineFx.Pulsar Topic subscription types


punxrok:
Hi,

is it possible to extend the library to define subscription type on Pulsar's topics in similar fashion to: https://pulsar.apache.org/docs/4.0.x/client-libraries-consumers/#shared
 
``` csharp
    opts.ListenToPulsarTopic("persistent://public/default/message")
        .SubscriptionName("round-robin")
        .SubscriptionType(SubscriptionType.Shared) // like this, currently the subscription type is fixed as Exclusive
        .MaximumParallelMessages(10, ProcessingOrder.UnOrdered);
```

jeremydmiller:
Closed by #1322

PR #1322, opened 2025-03-21
#1320 Added option to define subscription type on Pulsar's topics and…

punxrok:
… subscription name

#1320 Added option to define subscription type on Pulsar's topics and subscription name


jeremydmiller:
This will be in Wolverine 3.11 later today. Thank you!

PR #1345, opened 2025-04-05
Fixes broken link for envelope in docs

jerriep:
There was a broken link for the "Envelope" text in the Cascading Messages docs. Looking through the docs, I could not see anything obvious that it should be linking to, so I simply turned it into inline code. If there is a specific place in the docs it ought to be linking to, please let me know and I will adjust this PR.


jeremydmiller:
@jerriep Thank you! That's some kind of leftover from when we converted the old Jasper documentation using a homegrown documentation website building tool to VitePress & markdownsnippets. That's been there awhile:)

PR #1346, opened 2025-04-05
Configure Kafka builders

igor-toporet:
### Motivation

We're using [AWS MSK](https://aws.amazon.com/msk/) (Amazon Managed Streaming for Apache Kafka), and our containerized applications authenticate via [IAM](https://aws.amazon.com/iam/) (AWS Identity and Access Management) using https://github.com/aws/aws-msk-iam-sasl-signer-net library.

As described in that repository's [Getting Started](https://github.com/aws/aws-msk-iam-sasl-signer-net?tab=readme-ov-file#getting-started) guide, specific properties on the resulting product—producer, consumer, or admin client—must be set via the corresponding builder—specifically, [Set OAuthBearerTokenRefreshHandler](https://github.com/search?q=repo%3Aconfluentinc%2Fconfluent-kafka-dotnet%20SetOAuthBearerTokenRefreshHandler&type=code).

Currently, Wolverine does not offer this capability.

In addition to the use case above, many other circumstances might require leveraging the full potential of the underlying [Confluent.Kafka](https://github.com/confluentinc/confluent-kafka-dotnet) library.

One of the baseline scenarios is troubleshooting Kafka connectivity in an environment using [SetLogHandler](https://github.com/confluentinc/confluent-kafka-dotnet/blob/6b281a85de2e4ee836f3be2435961f336a638e1b/src/Confluent.Kafka/ProducerBuilder.cs#L252) or [SetErrorHandler](https://github.com/confluentinc/confluent-kafka-dotnet/blob/6b281a85de2e4ee836f3be2435961f336a638e1b/src/Confluent.Kafka/ProducerBuilder.cs#L223).

### Proposal

I'm proposing the code I used to extend Wolverine's Kafka transport. This approach was verified in the AWS environment, and being able to customize `Confluent.Kafka`'s logging was very helpful in troubleshooting.

For e.g., I used the following values (or a subset):

- `"broker,msg,security"` for `ProducerConfig.Debug`
- "broker,admin,security"` for `AdminClientConfig.Debug`
- `"consumer,cgrp,topic,fetch,broker,security" for `ConsumerConfig.Debug`

But those would only work if `Set[Log|Error]Handler` was used on a corresponding builder.

### Notable workaround

If this does not get accepted, the workaround I used initially was to use [OutgoingMessages](https://wolverinefx.net/guide/handlers/cascading.html#using-outgoingmessages) and handle them by explicit [handlers](https://wolverinefx.net/guide/handlers/) with Producer/Client/etc. injected via DI and published to Kafka by hand.

### In closing

I tried to adhere to the existing code style and declaration order. I also used the `internal` modifier where appropriate. 

Let me know if I need to change anything.

As always, a HUGE THANK YOU for the fantastic JasperFx libraries!

### DIFF
--- a/src/Transports/Kafka/Wolverine.Kafka/Internals/InlineKafkaSender.cs
+++ b/src/Transports/Kafka/Wolverine.Kafka/Internals/InlineKafkaSender.cs
@@ -12,7 +12,7 @@ public class InlineKafkaSender : ISender, IDisposable
     {
         _topic = topic;
         Destination = topic.Uri;
-        _producer = new ProducerBuilder<string, string>(_topic.Parent.ProducerConfig).Build();
+        _producer = _topic.Parent.CreateProducer();
         
     }
 
--- a/src/Transports/Kafka/Wolverine.Kafka/Internals/KafkaListener.cs
+++ b/src/Transports/Kafka/Wolverine.Kafka/Internals/KafkaListener.cs
@@ -16,11 +16,12 @@ internal class KafkaListener : IListener, IDisposable
     private readonly string? _messageTypeName;
     private readonly QualityOfService _qualityOfService;
 
-    public KafkaListener(KafkaTopic topic, ConsumerConfig config, IReceiver receiver,
+    public KafkaListener(KafkaTopic topic, ConsumerConfig config,
+        IConsumer<string, string> consumer, IReceiver receiver,
         ILogger<KafkaListener> logger)
     {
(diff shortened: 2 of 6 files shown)


rkim1999 (approved):

igor-toporet:
100% agree. I started looking into adding tests right after submitting the PR, but I  stuck on some of the existing ones failing on my machine.

PR #1347, opened 2025-04-05
Add support for DateOnly in query parameters

PerfectlyNormal:
Our application, which I'm experimenting adding Wolverine to, uses a lot of `DateOnly` where we don't care about the time.

This PR adds support for getting a `DateOnly` resolved as a query parameter instead of requiring a `DateTime` which then have to be converted to `DateOnly` before being passed along.

Couldn't find any tests for `DateTime` as query parameters, so I added some of those as well


jeremydmiller:
Nice! This will be in the next Wolverine release by the end of Monday

PR #1348, opened 2025-04-05
Update docs add warning about keyed services

AnnaSasDev:
Adds a callout warning to the docs, which explains that Wolverine doesnt (yet) support keyed services.
Hit this issue myself, and thus want to warn other people who are switching from MediatR to Wolverine.


jeremydmiller:
I'm going to rewrite that a bit, but thanks for taking the time here.

And Wolverine does far, far, far more than MediatR does over all. What you're talking about is an IoC feature.

Issue #1360, opened 2025-04-09
AutoProvision breaks when using tenants with RabbitMQ


b-nijhuis-innovadis:
**Describe the bug**
It seems Wolverine can't create any queues for the set tenants in RabbitMQ when calling AutoProvision, although the exchanges (including those for the dead letter queues) are created successfully. The queues for the main setup are created successfully though.

We are currently looking for a solution as alternative to MassTransit, so if this can be quickly fixed that would be great :)

**To Reproduce**
```csharp
builder.UseWolverine(wolverine =>
{
    var policies = wolverine.Policies;

    wolverine.CodeGeneration.TypeLoadMode = builder.Environment.IsDevelopment()
        ? TypeLoadMode.Dynamic
        : TypeLoadMode.Static;
    policies.UseDurableInboxOnAllListeners();
    policies.UseDurableOutboxOnAllSendingEndpoints();
    policies.DisableConventionalLocalRouting();

    wolverine
        .UseRabbitMq(rabbitMq =>
        {
            rabbitMq.HostName = "localhost";
            rabbitMq.Port = 5672;
            rabbitMq.UserName = "guest";
            rabbitMq.Password = "guest";
            rabbitMq.VirtualHost = "messageexamples";
        }) // outside tenancy works fine
        .DisableDeadLetterQueueing()
        .UseQuorumQueues()
        .AutoProvision()
        // This makes it break
        .AddTenant(
            "test",
            rabbitMq =>
            {
                rabbitMq.VirtualHost = "messageexamples2";
            }
        );

    foreach (var messageType in messageTypes)
    {
        var exchangeName = $"messageexamples.exchange.fanout.{messageType.Name}";
        var queueName = $"messageexamples.queue.{messageType.Name}.service-name";

        wolverine
            .Publish()
            .Message(messageType)
            .ToRabbitExchange(
                exchangeName,
                exchange =>
                {
                    exchange.ExchangeType = ExchangeType.Fanout;
                    exchange.IsDurable = true;
                    exchange.AutoDelete = false;
                }
            );
        var consumerTypes = AssemblyReference
            .Assembly.GetTypes()
            .Where(t =>
                t.IsAssignableTo(typeof(IBusMessageConsumer<>).MakeGenericType(messageType))
                && t is { IsInterface: false, IsAbstract: false }
            );
        if (consumerTypes.Any())
        {
            // NOTE: even with this commented out it breaks, because of the 'wolverine.response.xxx' queues not being able to exist
[...]

PR #1362, opened 2025-04-09
bug(rabbitmq): fix issue where RabbitMQ queues couldn't be declared for multiple tenants

b-nijhuis-innovadis:
Fixes https://github.com/JasperFx/wolverine/issues/1360

I removed a condition which prevented queue creation for multiple tenants, I tested this and found this to be working for my use cases. I compared this with how exchanges were declared, which also didn't check for it beforehand.


jeremydmiller:
Okay, but it's going to take a little more to get this exactly "right". And a test in the codebase. I'll handle that in a bit before this release.

PR #1364, opened 2025-04-09
Fix query string enum binding and alternate parameter names

esond:
- Makes enums parsed in query string values case-insensitive. This mirrors the behavior in ASP.NET Core ([ref](https://github.com/dotnet/aspnetcore/issues/3450)).
- When creating query parameters from an object bound using [FromQuery], uses the name set by [FromQuery] on any properties of that object. This mirrors existing query string binding behaviours.

Also updates tests to use lower-camel-cased query string values to catch any similar issues.


jeremydmiller:
@esond Nice! Will be in 3.13 very soon.

PR #1367, opened 2025-04-10
Add support for topics in RMQ conventional routing overrides

### LINKED ISSUE #1366: Support topic routing when using conventional routing
I have a use case where a subset of my messages (all of which implement a specific interface) need to be routed to specific consumers, which I would like to do using RabbitMQ topic exchange routing.

I would also like to keep using conventional routing, since there are other messages as well.

Currently, I need to explicitly call `PublishMessagesToRabbitMqExchange` for the affected message types, and recreate the exchange naming rules from the conventional routing conventions.

On the listening side, I would like to route all implementers of the this same interface to a topic exchange with a specific routing key, like this:

```csharp
opts.UseConventionalRouting(routing =>
  routing.ConfigureListeners((listener, ctx) =>
  {
      if (ctx.MessageType.IsAssignableTo(typeof(ILineSpecificMessage)))
      {
          // bind to topic exchange using ctx.MessageType for the generic T
          // this won't work, so either duplicate the if statements with concrete types
          // or do some unholy reflection to instantiate the generic method
          // listener.BindToExchange<T>(ExchangeType.Topic, lineId);
      }
  })
);
```

Specifically, two things would solve the issues I have:

- Something like this to configure topic publishing, while leaving the rest up to conventions
    ```csharp
    opts.UseConventionalRouting(routing =>
        routing.ConfigureSending((exchange, ctx) =>
        {
            if (ctx.MessageType.IsAssignableTo(typeof(ILineSpecificMessage)))
            {
                // not sure on the types here
                exchange.PublishMessagesToTopic<ILineSpecificMessage>(topicSource: msg => msg.LineId);
            }
        }));
    ```
- Something like this to configure consuming a topic for messages that implement a specific interface
   ```csharp
    opts.UseConventionalRouting(routing =>
        routing.ConfigureListeners((listener, ctx) =>
        {
            if (ctx.MessageType.IsAssignableTo(typeof(ILineSpecificMessage)))
            {
                // type parameter as a value rather than a generic type argument
                listener.BindToExchange(ctx.MessageType, ExchangeType.Topic, routingKey: lineId);
            }
        }));
    ```

Since header exchanges are supported by conventional routing, I guess the same could be implemented routing based on headers.

**Additional context**
This sparked from a short discussion on discord: https://discord.com/channels/1074998995086225460/1074999099855753317/1359848737601617930

Hawxy:
Closes #1366 

Needs tests - DNM

### DIFF
--- a/src/Transports/RabbitMQ/Wolverine.RabbitMQ/Internal/RabbitMqTransportExpression.cs
+++ b/src/Transports/RabbitMQ/Wolverine.RabbitMQ/Internal/RabbitMqTransportExpression.cs
@@ -106,7 +106,7 @@ public class RabbitMqTransportExpression : BrokerExpression<RabbitMqTransport, R
     public RabbitMqTransportExpression UseConventionalRouting(
         Action<RabbitMqMessageRoutingConvention>? configure = null)
     {
-        var convention = new RabbitMqMessageRoutingConvention();
+        var convention = new RabbitMqMessageRoutingConvention(Options);
         configure?.Invoke(convention);
         Options.RoutingConventions.Add(convention);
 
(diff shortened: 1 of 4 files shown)


Hawxy:
This aint it chief, lemme rethink this.

PR #1385, opened 2025-04-24
Swap rule.Modify and options.Override order

klagowski:
This change allows to apply different rules based on options specified for example in the header.


jeremydmiller:
@klagowski I hear you, but the values in DeliveryOptions have to win out over any rules. I'm sorry, but we can't take this in, this breaks functionality.

PR #1393, opened 2025-04-28
Add Pulsar configuration to unsubscribe on close

aeons:
This PR adds an `UnsubscribeOnClose(bool)` to `PulsarTransportExtension` to set whether the Pulsar listener should unsubscribe when it's closed.

The default is `true` to keep the current behavior.

Ideally, I would prefer this (and enable requeue) to default to false and be opt-in, but that would be a breaking change.

### Motivation
We have a setup where subscriptions are created by an external system, and due to access rights limitations, our consumer can't create subscriptions but can delete them.
So we need a way to have the subscription outlive the listener.

### DIFF
--- a/docs/guide/messaging/transports/pulsar.md
+++ b/docs/guide/messaging/transports/pulsar.md
@@ -72,8 +72,6 @@ builder.UseWolverine(opts =>
     {
         var pulsarUri = builder.Configuration.GetValue<Uri>("pulsar");
         c.ServiceUrl(pulsarUri);
-        
-        
     });
 
     // Listen for incoming messages from a Pulsar topic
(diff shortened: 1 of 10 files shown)


jeremydmiller:
@aeons Looks good. Couple points though:

* You could add another "all Pulsar listeners should be auto-closed" off of WolverineOptions.Policies w/ an extension method
* Be nice to have a little bit in the docs
* Even if it's just a smoke test, add some kind of test first?

aeons:
>You could add another "all Pulsar listeners should be auto-closed" off of WolverineOptions.Policies w/ an extension method

Yeah, that makes sense. I'll see if I can figure it out.

Should I add one for `EnableRequeue` as well?

> Be nice to have a little bit in the docs

Sure.

> Even if it's just a smoke test, add some kind of test first?

I'll try :)

aeons:
Re: testing, I can't really figure out how to verify that the subscription is not unsubscribed at the pulsar level.

I can add some tests that the configuration extensions do apply as expected.

aeons:
I've added `IPolicies` extension methods, added some tests that the configuration is applied as expected and updated the docs.

I cannot figure out a good way to test the actual functionality, so if you have any ideas, please let me know.

Issue #1409, opened 2025-05-08
List<T> properties in a FromQuery class generates uncompilable code


richardszalay:
**Describe the bug**
If a property defined in a FromQuery class is of type List<string> the generated code doesn't compile.

**To Reproduce**

```
public class ExampleQuery
{
  public List<string> Values { get; set; }
}

public class ExampleEndpoint
{
  [WolverineGet("/")]
  public static string Handle([FromQuery]ExampleQuery query)
  {
    return "";
  }
}
```

Generates:

```
var exampleQuery.Values = new System.Collections.Generic.List<string>();
exampleQuery.Values.AddRange(httpContext.Request.Query["exampleQuery.Values"]);
```

**Expected behavior**
Similar to if the property was an array (which does work)

jeremydmiller:
Closed by https://github.com/JasperFx/wolverine/pull/1411

PR #1411, opened 2025-05-08
Fix a few bugs with collections and FromQuery/FromForm

jakobt:


jakobt:
Closes https://github.com/JasperFx/wolverine/issues/1409

PR #1421, opened 2025-05-13
[Repro] Add test to show-case bug with nullable query string params

Xzelsius:
This is a reproduction to showcase a bug that occurs with Wolverine > `3.11.0`
On `3.11.0` it works fine.

Before upgrading to later Wolverine versions, the code gen create correct variables.
After upgrading, the code gen seems to create a variable named `nullable` multiple times which is not allowed/possible.


jeremydmiller:
@richardszalay got your test in and the fix. Wolverine 3.13.3 coming shortly w/ the fixes

PR #1426, opened 2025-05-15
Fix nullable querystrings (#1421)

richardszalay:


richardszalay:
(NOTE: This PR was done _on top_ of the #1421 branch so the original commit is not lost)

PR #1428, opened 2025-05-16
Allow many brokers for Kafka

dominikjeske:
This PR is not complete because it needs some additional changes. Please read comments in files.


src/Transports/Kafka/Wolverine.Kafka/KafkaTransportExtensions.cs
```diff
@@ -14,11 +14,11 @@ public static class KafkaTransportExtensions
     /// </summary>
     /// <param name="endpoints"></param>
     /// <returns></returns>
-    internal static KafkaTransport KafkaTransport(this WolverineOptions endpoints)
+    internal static KafkaTransport KafkaTransport(this WolverineOptions endpoints, BrokerName? name = null)
```
dominikjeske, line 17: BrokerName is to passed to GetOrCreate so I added this here

src/Wolverine/Configuration/PublishingExpression.cs
```diff
@@ -26,9 +26,17 @@ internal PublishingExpression(WolverineOptions parent)
     /// </summary>
     /// <param name="uri"></param>
     /// <returns></returns>
-    public ISubscriberConfiguration To(Uri uri)
+    public ISubscriberConfiguration To(Uri uri, BrokerName? brokerName = null)
     {
-        var endpoint = Parent.Transports.GetOrCreateEndpoint(uri);
+        Endpoint? endpoint = null;
+        if (brokerName is null)
```
dominikjeske, line 32: When using BrokerName Parent.Transports.GetOrCreateEndpoint is not working because we have more then one with same scheme. So I add condition. ForScheme has misleading name because it is looking transport name that could be other than scheme.

src/Wolverine/Runtime/WolverineRuntime.HostService.cs
```diff
@@ -234,6 +234,7 @@ private async Task startMessagingTransportsAsync()
             {
                 // There are a couple other places where senders might be getting
                 // started before this point, so latch to avoid double creations
+                // TODO: this will not work for two brokers
                 if (_endpoints.HasSender(endpoint.Uri)) continue;
```
dominikjeske, line 238: This code is problematic because when having more than one kafka broker we have more endpoints with 'wolverine.topics' so for second broker agent will not be created. I'm not sure how to fix this and not broke something else. (For testing purposes when commented this line of code I could work with success but it may break in other cases).

src/Wolverine/TransportCollection.cs
```diff
@@ -60,7 +61,7 @@ internal void AddPolicy(IEndpointPolicy policy)
 
     public ITransport? ForScheme(string scheme)
     {
-        return _transports.TryGetValue(scheme.ToLowerInvariant(), out var transport)
```
dominikjeske, line 63: When transport has not only schemas but broker name it could have names with capitals and when using to lower they are not found.

src/Wolverine/TransportCollection.cs
```diff
@@ -85,14 +86,13 @@ public void Add(ITransport transport)
         }
         else
         {
-            var transport = _transports.Values.OfType<T>().FirstOrDefault(x => x.Protocol == name.Name);
-            if (transport == null)
+            if (!_transports.TryGetValue(name.Name, out var transport))
             {
                 transport = (T)Activator.CreateInstance(typeof(T), name.Name);
```
dominikjeske, line 91: This code is fragile - not every transport has one parameter constructor

jeremydmiller:
I'm going to aim for this going by July 20th. Didn't make the cut for today, but will soon. I'm out next week though.

jeremydmiller:
@dominikjeske Sorry for how long this took, but I pushed through this feature today using the existing Rabbit MQ usage as the guide. Doesn't end up taking much code.

PR #1440, opened 2025-05-21
Binary producer/consumer for kafka

dominikjeske:
Why we need this change - in current implementation Wolverine is serializing message using IMessageSerializer to bytes but in kafka ProducerBuilder<string, string> is used to we are transforming this to string with Encoding.Default.GetString and inside IProducer<string, string>.ProduceAsync() another method is used to change this to bytes again with Encoding.Default.GetBytes (https://github.com/confluentinc/confluent-kafka-dotnet/blob/3f10fc8f829e7540e135fe8fe343371f1d4f5f2b/src/Confluent.Kafka/Serializers.cs#L42). So there is additional overhead that could be simplified by using ProducerBuilder<string, byte[]>.
Additionally in our case we implemented custom serializer using confluent avro and during tests we saw that Encoding.Default is not enough and byte array we get when deserializing are different. After investigation we find out that bytes produced by avro can't be change to UTF8 - it works only when we changed woverine code to Encoding.Unicode.

### DIFF
--- a/src/Transports/Kafka/Wolverine.Kafka/IKafkaEnvelopeMapper.cs
+++ b/src/Transports/Kafka/Wolverine.Kafka/IKafkaEnvelopeMapper.cs
@@ -6,7 +6,7 @@ using Wolverine.Util;
 
 namespace Wolverine.Kafka;
 
-public interface IKafkaEnvelopeMapper : IEnvelopeMapper<Message<string, string>, Message<string, string>>;
+public interface IKafkaEnvelopeMapper : IEnvelopeMapper<Message<string, byte[]>, Message<string, byte[]>>;
 
 /// <summary>
 /// Option to publish or receive raw JSON from or to Kafka topics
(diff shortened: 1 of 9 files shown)


src/Transports/Kafka/Wolverine.Kafka/IKafkaEnvelopeMapper.cs
```diff
@@ -23,17 +23,17 @@ public JsonOnlyMapper(KafkaTopic topic, JsonSerializerOptions options)
...
     }
 
-    public void MapEnvelopeToOutgoing(Envelope envelope, Message<string, string> outgoing)
+    public void MapEnvelopeToOutgoing(Envelope envelope, Message<string, byte[]> outgoing)
     {
         outgoing.Key = envelope.GroupId;
 
         if (envelope.Data != null && envelope.Data.Any())
         {
-            outgoing.Value = Encoding.Default.GetString(envelope.Data);
```
dominikjeske, line 32: We can skip this conversion

jeremydmiller:
Dammit, I should have gotten this in before 4.0. This is a breaking change, but I'm going to cheat and get it in anyway

Issue #1430, opened 2025-05-19
Outbox messages are not dispatched after RabbitMQ connection is restored unless application is restarted


hamedsabzian:
**Describe the bug**
When we have a network hiccup with RabbitMQ connection, the persisted messages in the outbox (`wolverine_outgoing_envelopes`) remain unsent till the next restart of the application.
Once the network issue is resolved, new messages are dispatched as expected, but the previously persisted messages from the disconnection period are never sent.

**To Reproduce**
1. Add marten to persisting the messages:
```csharp
builder.ConfigureServices((hostContext, services) =>
{
    services.AddMarten(options =>
        {
            options.DatabaseSchemaName = "durability";
            options.Connection(hostContext.Configuration.GetConnectionString("DefaultConnection")!);
        })
        .IntegrateWithWolverine();
});
```
3. Use `UseDurableOutbox` to outbox the published messages:
```csharp
    opts.PublishMessage<PingMessage>()
        .ToRabbitExchange("pings")
        .UseDurableOutbox(
```
5. Try to make RabbitMQ unavailable temporarily.
6. Make RabbitMQ available.
7. Check the rows of `wolverine_outgoing_envelopes`.

**Expected behavior**
I expected Wolverine to retry sending these messages after a while, and that they would be dispatched and removed from the outbox table automatically.

**Additional context**
I cloned the latest version of the code and worked on it.
This is my edited `Program.cs` content of Pinger project:
```csharp
using Marten;
using Oakton;
using Oakton.Resources;
using Pinger;
using Wolverine;
using Wolverine.Marten;
using Wolverine.RabbitMQ;

#region sample_bootstrapping_rabbitmq

var builder = Host.CreateDefaultBuilder(args);

builder.ConfigureServices((hostContext, services) =>
{
    services.AddMarten(options =>
        {
            options.DatabaseSchemaName = "durability";
            options.Connection(hostContext.Configuration.GetConnectionString("DefaultConnection")!);
        })
        .IntegrateWithWolverine();
});

builder.UseWolverine(opts =>
{
    // Listen for messages coming into the pongs queue
    opts.ListenToRabbitQueue("pongs")
        .UseDurableInbox()
        .DefaultIncomingMessage<PongMessage>();

    // Publish messages to the pings queue
    opts.PublishMessage<PingMessage>()
        .ToRabbitExchange("pings")
        .UseDurableOutbox();

    // Configure Rabbit MQ connection to the connection string
[...]

PR #1446, opened 2025-05-25
Fix remaining messages in the outbox when the transportation connection isn't stable

hamedsabzian:
To fix the reported problems in this issue: #1430

Fixed issues:
- Reset failure count after reconnecting and modifying it thread-safe in the sending agent (to ensure latching exactly after `FailuresBeforeCircuitBreaks` retries and reset after reconnecting).
- Make the in-memory queue thread-safe in durable outboxes.
- Fix issue in `RabbitMqChannelAgent` connection state management (rely on the RabbitMQ.Client connection recovery feature and recreate channels only after reconnecting).

Other improvements:
- Use PeriodicTimer in `CircuitWatcher` instead of using the `while` loop in
- Ensure RabbitMQ channels and consumptions will be recovered after connection recovery by creating a new channel and re-consumption.

### DIFF
--- a/src/Transports/RabbitMQ/Wolverine.RabbitMQ/Internal/ConnectionMonitor.cs
+++ b/src/Transports/RabbitMQ/Wolverine.RabbitMQ/Internal/ConnectionMonitor.cs
@@ -2,7 +2,6 @@ using JasperFx.Core;
 using Microsoft.Extensions.Logging;
 using RabbitMQ.Client;
 using RabbitMQ.Client.Events;
-using Wolverine.Runtime;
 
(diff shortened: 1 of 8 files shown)


jeremydmiller:
@hamedsabzian This is a big win. I'm pulling it in now

luisfelipediaz:
Sorry guys, for writing about a closed issue. But this issue is merged in Wolverine v4.3.0? 
I'm experiencing the same issue. I have 3 messages pending to send. When rabbitmq is up again, only the first message is sent, and the remaining messages are sent only if I restart the service.

PR #1451, opened 2025-05-28
Features/date bug

dominikjeske:


src/Wolverine/Transports/EnvelopeMapper.cs
```diff
@@ -431,7 +431,7 @@ protected DateTimeOffset readDateTimeOffset(TIncoming incoming, string key)
     {
         if (tryReadIncomingHeader(incoming, key, out var raw))
         {
-            if (DateTimeOffset.TryParse(raw, out var flag))
+            if (DateTimeOffset.TryParseExact(raw, DateTimeOffsetFormat, null, DateTimeStyles.AssumeUniversal,  out var flag))
```
dominikjeske, line 434: Header like 'SenAt' is not parsed because format is not maching

PR #1466, opened 2025-06-02
Use .NET 8 or .NET 9 with TargetFramework condition

dominikjeske:


src/Wolverine/Wolverine.csproj
```diff
@@ -9,18 +9,32 @@
...
+
+	<ItemGroup Condition=" '$(TargetFramework)' == 'net8.0' ">
+		<PackageReference Include="Microsoft.Extensions.Configuration" Version="[8.0.0,9.0.0)"/>
+		<PackageReference Include="Microsoft.Extensions.Caching.Memory" Version="[8.0.1,9.0.0)"/>
+		<PackageReference Include="Microsoft.Extensions.Configuration.Abstractions" Version="[8.0.0,9.0.0)"/>
+		<PackageReference Include="Microsoft.Extensions.Configuration.Binder" Version="[8.0.2,9.0.0)"/>
+		<PackageReference Include="Microsoft.Extensions.Configuration.Json" Version="[8.0.1,9.0.0)"/>
+		<PackageReference Include="Microsoft.Extensions.Hosting" Version="[8.0.1,9.0.0)"/>
+		<PackageReference Include="Microsoft.Extensions.Hosting.Abstractions" Version="[8.0.1,9.0.0)"/>
+		<PackageReference Include="Microsoft.Extensions.ObjectPool" Version="[8.0.0,9.0.0)"/>
```
dominikjeske, line 36: I copied this from previous version but I'm not sure it is what I have excepted. From this version [8.0.0,9.0.0) I would except 8.0.16 to be resolved (the newest) but 8.0.0 is resolved  "smallest acceptable stable version" from-> https://learn.microsoft.com/en-us/nuget/concepts/package-versioning?tabs=semver20sort. To always have newest maybe it would be better to use 8.0.* or just use specific version and update it manually?

PR #1507, opened 2025-06-19
fix: api description contain invalid request type for GET requests

uniquelau:
Tiny adjustment, the ensures that when generating the open api documentation, GET endpoints do not have request bodies and adds a couple of tests.

Currently, if the GET endpoint uses `[FromQuery] ComplexT` the type will shows up as the request body and is required.
This is incorrect, as these values are expected on the query string, not the request body.

This means that the swagger tooling can not be used to submit a request to the API and the generated JSON does not match the api specification.

In an ideal standard compliant world, the GET would support a request body (even though it's weird), but from looking at how RequestType works within the system, it would be a larger change, and from looking at some of the code, I'm not sure Wolverine supports this anyway.

+ Added test to replicate issue
+ Added fix to ensure generated API valid
+ Ran tests


uniquelau:
Just got a notification that the build has failed, I did check on my local using the docker setup, so apologies if I have missed something. I'll re-run locally and see what is happening.

Issue #1514, opened 2025-06-20
Outbox messages are not dispatched after RabbitMQ connection is restored unless application is restarted, combining event store and outbox


luisfelipediaz:
**Describe the bug**
When there's a temporary network disconnection from RabbitMQ, Wolverine stops dispatching messages that are persisted in the wolverine_outgoing_envelopes table. Once the RabbitMQ connection is re-established, Wolverine resumes dispatching new messages, but the persisted messages from the disconnection period are not sent — except for the first message in the queue. The remaining messages are only dispatched after restarting the service.

**To Reproduce**
1. Use Wolverine with Marten event sourcing and RabbitMQ, as shown in the example below.
2. Configure a command (PongMessage) that:
     - Appends an event to a Marten event stream.
     - Publishes a message (PingMessage) to a RabbitMQ exchange.
4. Enable UseDurableOutbox() for the outgoing messages.
5. Simulate a RabbitMQ disconnection (e.g., stop RabbitMQ service).
6. Send multiple commands (/send endpoint), which result in multiple events and outgoing messages.
7. Restart RabbitMQ (recover connection).
8. Observe that only one message is dispatched to the exchange.
9. Remaining messages are not dispatched until the app is restarted.

Code example:

```c#

var builder = WebApplication.CreateBuilder(args);

builder.Services.AddMarten(options =>
    {
        options.Events.StreamIdentity = StreamIdentity.AsString;
        options.Connection(builder.Configuration.GetConnectionString("MartenEventStore")!);
    })
    .UseLightweightSessions()
    .IntegrateWithWolverine();

builder.UseWolverine(opts =>
{
    // Publish messages to the pings queue
    opts.PublishMessage<PingMessage>()
        .ToRabbitExchange("pings")
        .UseDurableOutbox();

    // Enable auto apply transactions for Wolverine
    opts.Policies.AutoApplyTransactions();

    // Configure Rabbit MQ connection to the connection string
    // named "rabbit" from IConfiguration. This is *a* way to use
    // Wolverine + Rabbit MQ using Aspire
    opts.UseRabbitMqUsingNamedConnection("RabbitMQ")
        .AutoProvision();
});

// Add services to the container.
// Learn more about configuring OpenAPI at https://aka.ms/aspnet/openapi
builder.Services.AddOpenApi();

var app = builder.Build();

app.MapGet("/send", async (IMessageBus bus) =>
{
    var command = new PongMessage("Ping", "Pong");

[...]

PR #1515, opened 2025-06-20
Improved handling of latched senders in `QuickSendAsync`

luisfelipediaz:
To fix the reported problems in this issue: #1514 
This pull request introduces an enhancement to the `QuickSendAsync` method in `src/Wolverine/Envelope.Internals.cs` to improve handling of latched senders. The change ensures that messages are not sent immediately when the sender is latched, while maintaining durability and retry logic.

Key change:

* **Improved handling of latched senders in `QuickSendAsync`:**
  - Added logic to detect when the sender is latched (e.g., due to network disconnection or transport failure).
  - If the sender is a `SendingAgent`, the envelope is explicitly marked as failed using `MarkProcessingFailureAsync`, enabling retries when the connection is restored.

### DIFF
--- a/src/Wolverine/Envelope.Internals.cs
+++ b/src/Wolverine/Envelope.Internals.cs
@@ -240,7 +240,24 @@ public partial class Envelope
 
         _enqueued = true;
 
-        if (Sender.Latched) return;
+        if (Sender.Latched)
+        {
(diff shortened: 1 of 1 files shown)


jeremydmiller:
Oh, good catch!

PR #1520, opened 2025-06-23
Note to "Command Bus" broken link

donniehale-tl:
I couldn't find any obvious page that corresponds to "Command Bus". Perhaps the link text should be `[Wolverine as Message Bus]` and the link should be `(https://wolverinefx.net/guide/messaging/introduction.html#getting-started-with-wolverine-as-message-bus)`.


jeremydmiller:
See how the day and weekend goes, but I think it's time for an all new tutorial on this that I'll link to

jeremydmiller:
@donniehale-tl Somebody had gotten to this, and thanks for pointing it out!

PR #1523, opened 2025-06-23
Correct attribute name in explanatory text

donniehale-tl:


jeremydmiller:
Good catch, that's *old* docs that hasn't been touched in years

PR #1533, opened 2025-06-27
Add support for Complex QueryString in AsParameters

Damix48:
Fix using `[FromQuery]` in `[AsParameters]` for complex object.

This now works:
```csharp
[WolverinePost("/test5")]
public static Response Handle5([AsParameters] Parameters parameters)
{
    return new Response(Guid.NewGuid(), $"Ciao from {parameters.Query.Name}");
}

public class Parameters
{
    [FromQuery]
    public Query Query { get; set; }
    [FromBody]
    public Body Body { get; set; }
}

public class Query
{
    public string Name { get; set; }
    public int Age { get; set; }
}

public class Body
{
    public string Surname { get; set; }
}
```

### DIFF
--- a/src/Http/Wolverine.Http/CodeGen/AsParametersBindingFrame.cs
+++ b/src/Http/Wolverine.Http/CodeGen/AsParametersBindingFrame.cs
@@ -143,6 +143,11 @@ internal class AsParametersBindingFrame : SyncFrame
             var queryStringName = qatt.Name ?? memberName;
             variable =
                 chain.TryFindOrCreateQuerystringValue(memberType, queryStringName);
+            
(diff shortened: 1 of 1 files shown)


jeremydmiller:
@Damix48 Hey, I'll add the tests, but next time you can't make a PR that changes behavior like this w/o tests backing it up. For regression if nothing else.

Damix48:
> @Damix48 Hey, I'll add the tests, but next time you can't make a PR that changes behavior like this w/o tests backing it up. For regression if nothing else.

You're right, sorry about that. Unfortunately, I'm not used to writing tests, but I'll make sure to include them next time.

PR #1534, opened 2025-06-27
Fix body params in GET request when using complex query string

Damix48:
Now in GET request the `[FromQuery]` complex object is duplicated in the body.

image

I removed `chain.RequestType = parameter.ParameterType;` but i'm not sure if it is used for something else. In that case maybe a solution is creating something similar to `public bool IsFormData { get; internal set; }` in HttpChain.cs


jeremydmiller:
Thank you for spotting that

Issue #1546, opened 2025-07-02
Wolverine Marten integration testing with dynamic database schemes fails on dynamic code generation


erdtsieck:
> Below is all with code pre generated using `dotnet run -- codegen write`

## Setting up Marten using a different schema every run

```cs
webApplication.Services.CritterStackDefaults(x =>
{
    x.Development.ResourceAutoCreate = AutoCreate.All;
    x.Development.GeneratedCodeMode = TypeLoadMode.Dynamic;
});

webApplication.Services.AddMarten(sp =>
{
    var martenSettings = sp.GetRequiredService<IOptions<MartenSettings>>().Value;
    var opts = new StoreOptions
    {
        Events =
        {
            TenancyStyle = TenancyStyle.Single,
            UseIdentityMapForAggregates = true,
            AppendMode = EventAppendMode.Quick,
            UseArchivedStreamPartitioning = true,
            UseMandatoryStreamTypeDeclaration = true
        },
        DisableNpgsqlLogging = true
    };

    opts.UseSystemTextJsonForSerialization(enumStorage: EnumStorage.AsString);

    if (!string.IsNullOrEmpty(martenSettings.SchemaName)) // different for every test run
    {
        opts.Events.DatabaseSchemaName = martenSettings.SchemaName;
        opts.DatabaseSchemaName = martenSettings.SchemaName;
    }
    return opts;
})
```

```
An unhandled exception has occurred while executing the request.
System.InvalidOperationException: Wolverine may be having trouble with concurrent access to the same route at startup. Set the WolverineHttpOptions.Warmup = Eager to work around this problem
 ---> System.NullReferenceException: Object reference not set to an instance of an object.
   at Wolverine.Runtime.ServiceContainer.QuickBuild(Type concreteType) in /home/runner/work/wolverine/wolverine/src/Wolverine/Runtime/ServiceContainer.cs:line 321
   at Wolverine.Http.HttpChain.<BuildEndpoint>b__139_4() in /home/runner/work/wolverine/wolverine/src/Http/Wolverine.Http/HttpChain.EndpointBuilder.cs:line 78
   --- End of inner exception stack trace ---
   at Wolverine.Http.HttpChain.<BuildEndpoint>b__139_4() in /home/runner/work/wolverine/wolverine/src/Http/Wolverine.Http/HttpChain.EndpointBuilder.cs:line 82
   at System.Lazy`1.ViaFactory(LazyThreadSafetyMode mode)
[...]

PR #1547, opened 2025-07-02
#1546 Wolverine Marten integration testing with dynamic database schemes fails on dynamic code generation

erdtsieck:


erdtsieck:
It only fails locally since it is a parallelization error

PR #1548, opened 2025-07-02
Update getting-started.md to fix the broken link.

voroninp:
Fix the link of the reference `See also: Wolverine as Command Bus` in the "Getting Started" of the documentation.


voroninp:
I'm not 100% sure this is the right url, though =)

jeremydmiller:
It is, and thank you!

At some point it's worth a new tutorial on using Wolverine as an in memory command bus, but that's just not happening anytime soon.

PR #1550, opened 2025-07-03
fix: Use correct IFormFile/IFormFileCollection type in ApiDescription…

### LINKED ISSUE #1549: OpenAPI description treats IFormFileCollection parameter as IFormFile
When using IFormFileCollection with Wolverine.HTTP, the generated code correctly uses `ReadManyFormFileValues` to read all the files instead of just one. But when viewing the generated OpenAPI description, it is treated as a single IFormFile instead of an array of IFormFile.

**To Reproduce**
1. Create a Web Api with a Wolverine.Http endpoint using IFormFileCollection
2. View generated OpenAPI file (`/openapi/v1.json`)
3. Find endpoint and see that request body contains only one IFormFile.

```json
        "requestBody": {
          "content": {
            "application/x-www-form-urlencoded": {
              "schema": {
                "required": [
                  "files"
                ],
                "type": "object",
                "properties": {
                  "files": {
                    "$ref": "#/components/schemas/IFormFile"
                  }
                }
              }
            }
          },
          "required": true
        },
```

**Expected behavior**
IFormFileCollection should be treated as an Array by OpenAPI, as such:

```json
        "requestBody": {
          "content": {
            "application/x-www-form-urlencoded": {
              "schema": {
                "required": [
                  "files"
                ],
                "type": "object",
                "properties": {
                  "files": {
                    "$ref": "#/components/schemas/IFormFileCollection"
                  }
                }
              }
            }
          },
          "required": true
        },
```

akselkvitberg:
… for FileParameters

fixes #1549

### DIFF
--- a/src/Http/Wolverine.Http/HttpChain.ApiDescription.cs
+++ b/src/Http/Wolverine.Http/HttpChain.ApiDescription.cs
@@ -97,7 +97,7 @@ public partial class HttpChain
                     Name = parameter.Name,
                     ParameterType = parameter.ParameterType
                 },
-                Type = typeof(IFormFile),
+                Type = parameter.ParameterType,
                 IsRequired = true
             };
 


jeremydmiller:
Good catch, and thank you!

PR #1552, opened 2025-07-04
Improve Mapper Extensibility

vzam:
**Overview**
As part of migrating a production system to Wolverine, we encountered scenarios where we need to process mixed message formats through a single listener, particularly with Kafka and MQTT. This PR introduces targeted enhancements to support such use cases without breaking existing behavior.

**Changes**
- Allow full configuration of the endpoint during interop setup.
  - Note that previous methods similar to `UseInterop` were not removed to avoid breaking changes - it might be duplicated for some configurations.
- MQTT topic name is now added to incoming envelopes.
- Several base methods in EnvelopeMapper are now virtual, allowing selective overrides.
- Internal mappers were made public so they can be extended or wrapped without copying internal logic.

**Motivation**
- We need to handle versioned messages in Kafka using a single listener.
- Additionally, MQTT messages must be validated based on the topic they were received on.

These changes allow us to plug in custom logic without having to fork or duplicate Wolverine internals.

**Example**
```csharp
public abstract class VersionedKafkaEnvelopeMapper
    : IEnvelopeMapper<Message<string, string>, Message<string, string>>
{
    /// <summary>
    /// The mappers for different versions of Kafka messages.
    /// If messages without a version header are received, a mapper with the key "" will be used.
    /// </summary>
    protected abstract Dictionary<string, IKafkaEnvelopeMapper> Mappers { get; }
    
    public void MapEnvelopeToOutgoing(Envelope envelope, Message<string, string> outgoing)
    {
        var version = envelope.Headers["version"] ?? string.Empty;

        if (!Mappers.TryGetValue(version, out var mapper))
            throw new NotSupportedException("Unsupported Kafka message version: " + version);
        
        mapper.MapEnvelopeToOutgoing(envelope, outgoing);
        outgoing.Headers.Add("version", Encoding.Default.GetBytes(version));
    }

    public void MapIncomingToEnvelope(Envelope envelope, Message<string, string> incoming)
    {
        var version = string.Empty;
        if (incoming.Headers.TryGetLastBytes("version", out var versionBytes))
        {
            version = Encoding.Default.GetString(versionBytes);
        }
        
        if (!Mappers.TryGetValue(version, out var mapper))
            throw new NotSupportedException("Unsupported Kafka message version: " + version);
        
        mapper.MapIncomingToEnvelope(envelope, incoming);
    }

    public IEnumerable<string> AllHeaders()
    {
        return Mappers.SelectMany(x => x.Value.AllHeaders()).ToHashSet();
    }
}

public class MyKafkaEnvelopeMapper(Endpoint endpoint) : VersionedKafkaEnvelopeMapper
{
    protected override Dictionary<string, IKafkaEnvelopeMapper> Mappers { get; } = new()
    {
        { string.Empty, new LegacyKafkaEnvelopeMapper(endpoint) },
        { "v1", new KafkaEnvelopeMapper(endpoint) },
    };
}
```

### DIFF
--- a/src/Transports/Azure/Wolverine.AzureServiceBus/Internal/AzureServiceBusEnvelopeMapper.cs
+++ b/src/Transports/Azure/Wolverine.AzureServiceBus/Internal/AzureServiceBusEnvelopeMapper.cs
@@ -5,7 +5,7 @@ using Wolverine.Transports;
 
 namespace Wolverine.AzureServiceBus.Internal;
 
-internal class AzureServiceBusEnvelopeMapper : EnvelopeMapper<ServiceBusReceivedMessage, ServiceBusMessage>, IAzureServiceBusEnvelopeMapper
+public class AzureServiceBusEnvelopeMapper : EnvelopeMapper<ServiceBusReceivedMessage, ServiceBusMessage>, IAzureServiceBusEnvelopeMapper
 {
     public AzureServiceBusEnvelopeMapper(Endpoint endpoint, IWolverineRuntime runtime) : base(endpoint)
     {
--- a/src/Transports/Kafka/Wolverine.Kafka/Internals/KafkaEnvelopeMapper.cs
+++ b/src/Transports/Kafka/Wolverine.Kafka/Internals/KafkaEnvelopeMapper.cs
@@ -5,11 +5,10 @@ using Wolverine.Transports;
 
 namespace Wolverine.Kafka.Internals;
 
-internal class KafkaEnvelopeMapper : EnvelopeMapper<Message<string, byte[]>, Message<string, byte[]>>, IKafkaEnvelopeMapper
+public class KafkaEnvelopeMapper : EnvelopeMapper<Message<string, byte[]>, Message<string, byte[]>>, IKafkaEnvelopeMapper
 {
     public KafkaEnvelopeMapper(Endpoint endpoint) : base(endpoint)
     {
-
     }
 
     protected override void writeOutgoingHeader(Message<string, byte[]> outgoing, string key, string value)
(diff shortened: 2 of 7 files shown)


jeremydmiller:
Nice, I'll get this in soon!

jeremydmiller:
@vzam I think this mostly looks good, but we'll need to add some documentation after this. In your `UseInterop()` method, I think there already is a method to do raw configuration right to the Endpoint types. Why did you need to full Endpoint and not just the envelope mapper in your case?

vzam:
@jeremydmiller I didn't use the `EnvelopeMapper` directly because it's tied to specific endpoint implementations, whereas `UseInterop()` is intended to work at the generic Listener/Subscriber configuration level. The idea is to keep it consistent unless there's a need to restrict settings at that level.

vzam:
There might be an issue with the envelope mapper in this PR, I am doing some tests and will share my results soon-ish.

jeremydmiller:
I'm going to aim for this going by July 20th. Didn't make the cut for today, but will soon.

jeremydmiller:
@vzam I reviewed this a little closer today. Sorry, but I can't take this **as is**, mostly because there are no tests, no sample usages, and no docs. 

Can you instead break the changes to the individual mappers up a bit? I'd take the MQTT changes independently.

I think I'd rather have a specialized `UseInterop()` merely for different header mappings. I don't think the current UseInterop() signature is right.

jeremydmiller:
Nevermind my last note. I think I'll take this in as is, but probably modify it somewhat as we go. You got me making notes about interop improvements

jeremydmiller:
@vzam Man, I'm sorry how long this has hung out here. This I think will roll into Wolverine 5.0 soon. If the day goes well, I might even incorporate some of this this afternoon

jeremydmiller:
Hey, this all got incorporated into the 5.0 branch, if a touch differently in some cases. Even though this PR didn't land, it inspired a ton of work that will release soon. Thank you for taking the time to do this, and sorry it's taken so long to move through.

jeremydmiller:
@vzam Maybe check this out? https://github.com/JasperFx/wolverine/blob/main/docs/tutorials/interop.md

And when I do make the release, do you want to be in the release notes? It's a major contribution here.

PR #1553, opened 2025-07-05
Http - Include errors from FluentValidation HTTP 400 produced errors

nicholas-brooks:
This will send a HTTP 400 and a ProblemDetails response body for FluentValidation validation errors that contains an errors property.  This errors property contains the list of fields that are in error with a description of each error produced by the validation on that field.


nicholas-brooks:
Completely misread the original issue.  Closing for now until I can add tests to prove the OpenAPI spec includes the errors property.

Issue #1568, opened 2025-07-14
Database backed DLQ not working with RabbitMQ


Joelgullander:
**Describe the bug**
I can't for the world get database backed DLQ's to work with my rabbitmq setup. Would appreciate any help I can get to resolve this. 

Should be able to be reproduced by the repo below, I created a minimal sample where I can myself reproduce it.

**To Reproduce**
https://github.com/Joelgullander/Wolverine.DLQ.Repro

**Expected behavior**
I expect the DLQ:ed messages to be available when asking for them by the DLQ REST API or by query on table wolverine_dead_letters.

**Screenshots**
N/A

jeremydmiller:
I think this is completely addressed by https://github.com/JasperFx/wolverine/pull/1571 and the accompanying fix commits I did for @Joelgullander 's tests

PR #1571, opened 2025-07-15
RabbitMQ Transport with DisableDeadLetterQueueing() not sending to sql DLQ storage

Joelgullander:
# 🐞 Bug Reproduction: RabbitMQ DLQ Not Saved to DB When Using WolverineStorage

## Overview

This PR adds a test to demonstrate a bug where failed RabbitMQ messages are **not written to the SQL Server Dead Letter Queue (DLQ)** when using `.DisableDeadLetterQueueing()` (i.e., when `DeadLetterQueueMode` is set to `WolverineStorage`). Instead, the messages are not persisted to the database as expected.

---

## Key Findings

- **File:** `src/Wolverine/Runtime/WorkerQueues/DurableReceiver.cs`
- The property:
  ```csharp
  public bool NativeDeadLetterQueueEnabled => true;
  ```
  is **hardcoded to always return `true`**, regardless of the actual DLQ mode.
- Even when configuring:
  ```csharp
  opts.UseRabbitMq().DisableDeadLetterQueueing()
  ```
  and `queue.DeadLetterQueue.Mode == DeadLetterQueueMode.WolverineStorage`, Wolverine still treats the queue as if native DLQ is enabled.
- As a result, the code path for writing to the SQL Server DLQ is never taken, and failed messages are not persisted to the database.

---

## Relevant Logic in MessageContext.cs

The decision to use the native DLQ or the database store is made in `MoveToDeadLetterQueueAsync`:

```csharp
if (_channel is ISupportDeadLetterQueue { NativeDeadLetterQueueEnabled: true } c)
{
    // Use native DLQ (e.g., RabbitMQ error queue)
    await c.MoveToErrorsAsync(Envelope, exception);
    return;
}

// Otherwise, use database-backed DLQ
await Storage.Inbox.MoveToDeadLetterStorageAsync(Envelope, exception);
```

Because `NativeDeadLetterQueueEnabled` is always `true` for `DurableReceiver`, the code always takes the native DLQ branch and never posts to the database store, even when the queue is configured for `WolverineStorage`.

---

## Test Details

- **File:** `src/Transports/RabbitMQ/Wolverine.RabbitMQ.Tests/Bugs/Bug_DLQ_NotSavedToDatabase.cs`
- The test:
  - Configures RabbitMQ with `.DisableDeadLetterQueueing()` and SQL Server persistence.
  - Sends messages that always fail in the handler.
  - Asserts that the messages should appear in the SQL Server DLQ table.
  - Prints debug output for DLQ mode and database state.
- **Result:** The test fails, confirming the bug.

---

## Why This Matters

- Users expect that disabling native RabbitMQ DLQ and enabling SQL Server persistence will result in failed messages being stored in the database.

---

## Next Steps

- The likely fix is to make `DurableReceiver.NativeDeadLetterQueueEnabled` reflect the actual DLQ mode of the endpoint/queue, rather than always returning `true`.

---

If you need more context or want to see the test output, please let me know!

---

**Closes/Relates to:** #1568

### DIFF
--- a/src/Transports/RabbitMQ/Wolverine.RabbitMQ.Tests/Bugs/Bug_DLQ_NotSavedToDatabase.cs
+++ b/src/Transports/RabbitMQ/Wolverine.RabbitMQ.Tests/Bugs/Bug_DLQ_NotSavedToDatabase.cs
@@ -15,7 +15,7 @@
 // Investigation: This test demonstrates the current behavior and investigates whether
 // this is the intended design or if there's a configuration issue.
 //
-// See also: https://github.com/JasperFx/wolverine/issues/<your-issue-number>
+// See also: https://github.com/JasperFx/wolverine/issues/1568
 
 using System;
 using System.Threading.Tasks;


jeremydmiller:
"The likely fix is to make DurableReceiver.NativeDeadLetterQueueEnabled reflect the actual DLQ mode of the endpoint/queue, rather than always returning true." -- that logic is 100% correct because it reflects `DurableReceiver` itself.

@Joelgullander Since you're so focused on the details here, use `IMessageStore.Admin` instead of raw SQL to test and access the DB-backed DLQ.

Joelgullander:
> @Joelgullander Since you're so focused on the details here, use `IMessageStore.Admin` instead of raw SQL to test and access the DB-backed DLQ.

I have implemented 3 new testcases, showing that durable fails, but non durable passes. I also switched to use the IMessageStore.Admin instead.

Joelgullander:
@jeremydmiller I get your point about the DurableReceiver logic, but I think there's actually a bug in the RabbitMQ transport implementation. I've been testing this and found something interesting.

The `TryBuildDeadLetterSender` method in RabbitMqQueue **always returns `true`** no matter what:

```csharp
public override bool TryBuildDeadLetterSender(IWolverineRuntime runtime, out ISender? deadLetterSender)
{
    var dlq = _parent.Queues[DeadLetterQueue?.QueueName ?? _parent.DeadLetterQueue.QueueName];
    deadLetterSender = dlq.CreateSender(runtime);
    return true; // Always returns true, never checks DeadLetterQueueMode
}
```

For comparison, here's how Amazon SQS does it:

```csharp
public override bool TryBuildDeadLetterSender(IWolverineRuntime runtime, out ISender? deadLetterSender)
{
    if (DeadLetterQueueName.IsNotEmpty() && !_parent.DisableDeadLetterQueues)
    {
        var dlq = _parent.Queues[DeadLetterQueueName];
        deadLetterSender = new InlineSqsSender(runtime, dlq);
        return true;
    }

    deadLetterSender = default;
    return false; // Returns false when DLQ should not be used
}
```

## Why this prevents SQL Server DLQ

Because `TryBuildDeadLetterSender` always returns `true`, the DurableReceiver `_moveToErrors` always takes the RabbitMQ DLQ path:

```csharp
if (_deadLetterSender != null)  // This is always non-null because TryBuildDeadLetterSender returns true
{
    await _deadLetterSender.SendAsync(envelope); // RabbitMQ DLQ path - ALWAYS taken
    return;
}

// This SQL Server DLQ path is NEVER reached
var report = new ErrorReport(envelope, envelope.Failure!);
await _inbox.MoveToDeadLetterStorageAsync(report.Envelope, report.Exception);
```

Joelgullander:
Like I said in discord, I might be totally wrong here. And apologize if that's the case. Either way this is a great way for me to understand the core mechanics :)

jeremydmiller:
This logic as you said, might not be perfectly catching the exact permutation of things that you're doing w/ Rabbit MQ:

```csharp
public override bool TryBuildDeadLetterSender(IWolverineRuntime runtime, out ISender? deadLetterSender)
{
    if (DeadLetterQueueName.IsNotEmpty() && !_parent.DisableDeadLetterQueues)
    {
        var dlq = _parent.Queues[DeadLetterQueueName];
        deadLetterSender = new InlineSqsSender(runtime, dlq);
        return true;
    }

    deadLetterSender = default;
    return false; // Returns false when DLQ should not be used
}
```
 
That might need to be smarter. The durable endpoint behavior is correct though. 

The combo of Rabbit MQ, not durable, but database backed DLQ isn't common at all. We had a separate issue I addressed for a JasperFx client on that last week. Honestly can't remember if it was pushed, but that was a problem w/ Rabbit MQ ack'ing and not w/ the DLQ mechanics themselves

Joelgullander:
Hey, thanks for the context & explaination.

I’m happy to take a shot at a fix for TryBuildDeadLetterSender in the RabbitMQ transport, but I want to make sure I’m not missing any edge cases or intended behaviors.

From what I see, the logic should probably be:

- If the DLQ mode is Native, create a sender (so RabbitMQ handles the DLQ).
- If the DLQ mode is WolverineStorage (i.e., .DisableDeadLetterQueueing()), don’t create a sender—let Wolverine move failed messages to the DB DLQ.
- If there’s no DLQ configured, don’t create a sender.

This would make it behave more like the SQS implementation we discussed above, and it matches what my tests are expecting.

If that sounds right, I can put together a proposed fix. But if there are other scenarios I should be aware of, or if you’d rather handle it, I’m happy to wait for your fix!

Just let me know how you’d like to proceed. Thanks again for the help and the quick responses! Meanwhile, I will just opt out from durable queues for now, to get the db backed DLQ to work.

jeremydmiller:
@Joelgullander I'm back in action today. This is next up for me.

jeremydmiller:
@Joelgullander Got this in and the fix w/ a rebase. Thanks for doing all the leg work here!

PR #1574, opened 2025-07-18
Add async serialization support

dominikjeske:


src/Wolverine/Envelope.cs
```diff
@@ -120,6 +121,30 @@ public TimeSpan? ScheduleDelay
         get => _scheduleDelay;
     }
 
+    public async ValueTask<byte[]?> GetDataAsync()
```
dominikjeske, line 124: This is additive change GetDataAsync can be executed and it fallback to Data

src/Wolverine/Envelope.cs
```diff
@@ -120,6 +121,30 @@ public TimeSpan? ScheduleDelay
...
+    {
+        if (_data != null)
+        {
+            return _data;
+        }
+        AssertMessage();
+
+        if(Serializer is IAsyncMessageSerializer asyncMessaeSerializer)
```
dominikjeske, line 132: We use only when serializer support async

src/Wolverine/Runtime/Serialization/IMessageSerializer.cs
```diff
@@ -14,4 +14,14 @@ public interface IMessageSerializer
     object ReadFromData(byte[] data);
 
     byte[] WriteMessage(object message);
+}
+
+/// <summary>
+///  Async version of <seealso cref="IMessageSerializer"/>
+/// </summary>
+public interface IAsyncMessageSerializer : IMessageSerializer
```
dominikjeske, line 22: We extend standard serializer so nothing is changing

src/Wolverine/Runtime/IHandlerPipeline.cs
```diff
@@ -7,5 +7,5 @@ public interface IHandlerPipeline
 {
     Task InvokeAsync(Envelope envelope, IChannelCallback channel);
     Task InvokeAsync(Envelope envelope, IChannelCallback channel, Activity activity);
-    bool TryDeserializeEnvelope(Envelope envelope, out IContinuation continuation);
+    ValueTask<IContinuation> TryDeserializeEnvelope(Envelope envelope);
```
dominikjeske, line 10: This is only one potential breaking change but I'm wonder if this interface should be public? Maybe it should be internal or nobody is executing this so event public we don't break anything

jeremydmiller: This should not be part of the public interface. It's just a detail of the HandlerPipeline details

src/Transports/RabbitMQ/Wolverine.RabbitMQ/Internal/RabbitMqSender.cs
```diff
@@ -26,29 +27,40 @@ public RabbitMqSender(RabbitMqEndpoint endpoint, RabbitMqTransport transport,
...
-                }
-                catch (Exception e)
-                {
-                    Logger.LogError(e, "Error trying to deserialize an envelope in order to determine the topic name");
-                }
-            }
-            
-            return new CachedString(TopicRouting.DetermineTopicName(x));
-        };
+            _routingKey = new CachedString(endpoint.RoutingKey());
```
jeremydmiller, line 33: Pretty sure that's there for MassTransit interop. Why are you deleting this?

PR #1596, opened 2025-07-23
Awaiting on send task

dominikjeske:


src/Wolverine/Util/Dataflow/RetryBlock.cs
```diff
@@ -85,19 +85,14 @@ public void Post(T message)
...
+    public Task PostAsync(T message)
     {
-        if (_cancellationToken.IsCancellationRequested) return;
+        if (_cancellationToken.IsCancellationRequested) return Task.CompletedTask;
+
+        Post(message);
+
+        return _block.Completion;
 
-        try
```
dominikjeske, line 92: I'm not sure why first approach is outside ActionBlock - is this way we have some code duplication so this is why I propose combine this in one execution in ActionBlock but this is not the main problem so if this was some perf optimization and you prefer to execute first step outside ActionBlock I can rollback this.

src/Wolverine/Util/Dataflow/RetryBlock.cs
```diff
@@ -85,19 +85,14 @@ public void Post(T message)
...
-        {
-            await _handler.ExecuteAsync(message, _cancellationToken);
-        }
-        catch (Exception e)
-        {
-            _logger.LogError(e, "Error while trying to retry {Item}", message);
```
dominikjeske, line 98: The error massage is not exactly true - this was first approach and not retry error

PR #1598, opened 2025-07-28
Produce validation problems metadata for OpenAPI in FluentValidation middleware

Damix48:
Produce `HttpValidationProblemDetails` metadata for OpenAPI in FluentValidation middleware instead of generic `ProblemDetails`


jeremydmiller:
Nice!

PR #1603, opened 2025-07-30
Protobuf serialization

lukedukeus:
Adds support for protobuf serialization


jeremydmiller:
@lukedukeus I'm pulling this in now, but it won't hit Nuget until 4.9 -- but that will happen within a week.

PR #1608, opened 2025-07-30
Add docs for multi-database migrations process

### LINKED ISSUE #1602: Wolverine db resources not being included in generated migrations
**Describe the bug**

**To Reproduce**

A full reproduction is reproduced here: [WebApiWithMarten](https://github.com/johnbaro/wolverine/tree/bug-migrations/src/Samples/WebApiWithMarten)

1. Setup project with Wolverine integrated with Marten
2. Generate migration script  
```
dotnet run --framework net8.0  -- db-patch -d postgresql://localhost/postgres/orders  ./Migrations/2025-07-30T092620_Initial.sql
```
3. Apply the migration script by running it against the target DB
5. Run the project
6. On startup, wolverine complains about missing relation
```
Npgsql.PostgresException (0x80004005): 42P01: relation "orders.wolverine_nodes" does not exist
```

**Expected behavior**
`wolverine_nodes` (and any other relevant wolverine resources) should be included in the generated script

johnbaro:
Closes #1602

### DIFF
--- a/docs/guide/durability/managing.md
+++ b/docs/guide/durability/managing.md
@@ -122,7 +122,9 @@ can use this command to apply any outstanding database changes necessary to brin
 ```bash
 dotnet run -- db-apply
 ```
- or this option -- but just know that this will also clear out any existing message data:
+> NOTE: See the [Exporting SQL Scripts](#exporting-sql-scripts) section down the page for details of applying migrations when integrating with Marten
+
+or this option -- but just know that this will also clear out any existing message data:
 
 ```bash
 dotnet run -- storage rebuild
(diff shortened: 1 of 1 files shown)


johnbaro (commented):
Please confirm my understanding is correct.
Initially I assumed that the `DatabaseUri` could be passed as well as the `SubjectUri`, on further testing this appears incorrect so I've squashed that change

jeremydmiller:
> Initially I assumed that the `DatabaseUri` could be passed as well as the `SubjectUri`

I think that's something I *meant* to enable, but didn't get around to yet

jeremydmiller:
The build failure isn't related, I'll address that whatever it is

PR #1613, opened 2025-07-31
take capabilites from `WolverineOptions`

lukedukeus:


jeremydmiller:
This isn't going to happen

PR #1614, opened 2025-08-01
Add sync way to retry in InlineSendingAgent

dominikjeske:


src/Wolverine/Transports/Sending/InlineSendingAgent.cs
```diff
@@ -20,16 +20,18 @@ public InlineSendingAgent(ILogger logger, ISender sender, Endpoint endpoint, IMe
         _settings = settings;
         Endpoint = endpoint;
 
-        if (endpoint.TelemetryEnabled)
+        if (settings.UseSyncRetryBlock)
```
dominikjeske, line 23: We use optional property to switch to different behavior so no breaking change is added

src/Wolverine/Util/Dataflow/IRetryBlock.cs
```diff
@@ -0,0 +1,15 @@
+namespace Wolverine.Util.Dataflow;
+
+/// <summary>
+/// Abstract a way we can retry on <typeparamref name="T"/> message
+/// </summary>
+/// <typeparam name="T"></typeparam>
+internal interface IRetryBlock<T> : IDisposable
```
dominikjeske, line 7: Abstraction for retry logic

src/Wolverine/Util/Dataflow/RetryBlockSync.cs
```diff
@@ -0,0 +1,143 @@
...
+        _block.Post(item);
+    }
+
+    public Task PostAsync(T message)
+    {
+        if (_cancellationToken.IsCancellationRequested) return Task.CompletedTask;
+
+        Post(message);
+        
+        return _block.Completion;
```
dominikjeske, line 75: In this version we are returning Task from ActionBlock

src/Wolverine/Util/Dataflow/RetryBlockSync.cs
```diff
@@ -0,0 +1,143 @@
...
+
+            if (item.Attempts > 1)
+            {
+                var pause = DeterminePauseTime(item.Attempts);
+                await Task.Delay(pause, _cancellationToken);
+            }
+            await _handler.ExecuteAsync(item.Message, _cancellationToken);
+            _logger.LogDebug("Completed {Item}", item.Message);
+
+            _block.Complete();
```
dominikjeske, line 104: Success

PR #1623, opened 2025-08-05
Add F# examples demonstrating issues with Wolverine usage

marcpiechura:


jeremydmiller:
@marcpiechura This will make the next Wolverine release. I just got bleary eyed, so I'm cutting a release w/ some other things tonight. I'll get this, and thank you!

marcpiechura:
FYI the Wolverine.Http.Tests are failing because both WolverineWebApi and WolverineWebApiFsharp are now executables and the Alba Host can't decide which Program class to use. I tried a few things to resolve it, but couldn't find a solution. 
If you want to keep the examples and maybe guard them with tests for the future, the easiest way might be to separate the F# part completely by creating a Wolverine.Http.Fsharp.Tests project which uses WolverineWebApiFsharp and removing it from WolverineWebApi.

jeremydmiller:
> FYI the Wolverine.Http.Tests are failing because both WolverineWebApi and WolverineWebApiFsharp are now executables and the Alba Host can't decide which Program class to use. I tried a few things to resolve it, but couldn't find a solution. If you want to keep the examples and maybe guard them with tests for the future, the easiest way might be to separate the F# part completely by creating a Wolverine.Http.Fsharp.Tests project which uses WolverineWebApiFsharp and removing it from WolverineWebApi.

I'll deal with that one some time this week. It's annoying, but there's a couple tricks for dealing w/ that

marcpiechura:
Sounds good. 
I added another example when using Task without returning a value in a handler.

jeremydmiller:
@marcpiechura Man, I'm sorry. Can you remind me what this was about? I let it slide and now I don't remember the context

jeremydmiller:
@marcpiechura Nevermind, I'm on it. Got `Unit` working fine. Let me see about the other stuff. Be nice to say "now better w/ F#"

jeremydmiller:
@marcpiechura So the `Unit` support was easy. JasperFx now also "knows" that `Tuple` is a `Tuple` and separates things out. In your validation sample though, using a `string` is going to confuse the codegen for now. By default Wolverine is going to see any primitive or "simple" type (string, Guid, numbers, DateTime of all sorts, bool, enum values) and assume that it should use a querystring value in the absence of any other explicit way to fill that value.

In other words, the codegen treats "simple" types differently. Annoying, but if you wrapped the validation result inside of a struct of some sort you'd be good to go.

jeremydmiller:
@marcpiechura The side effect issue is because the `StartStream<T>()` requires the generic type argument. There actually isn't an overload that doesn't take a generic argument. That's a quirk of F# I wasn't expecting. Is this method effectively an "open" method?

```fs
[<WolverinePost("start-stream")>]
[<EmptyResponse>]
let post (command: Command) =
    let event: Event = { Id = Guid.NewGuid(); Name = command.Name }
    //this doesn't work
    MartenOps.StartStream(event.Id, box event)
```

jeremydmiller:
The other fixes for Tuple and Unit are coming shortly though

jeremydmiller:
This got pulled in w/ some rebase/squash mechanics. @marcpiechura Thank you for doing this work! Helped a great deal in addressing most of this.

PR #1627, opened 2025-08-06
fix(doc): Add a tip about Compound Handlers restriction

AlexandreBossard:


jeremydmiller:
@AlexandreBossard I added some text just now that's a little bit more accurate:)

Thank you for pitching in though!

PR #1646, opened 2025-08-25
Fix sentence

RustyF:
Missing word or too many, not sure but make it's right.


jeremydmiller:
@RustyF It'd be better if I didn't make grammatical errors, but I'm always happy for these doc fixes. Thank you!

PR #1652, opened 2025-08-28
docs: add signposting around NotFound saga method

uniquelau:
- update code examples
- update text to state that an exception is thrown


jeremydmiller:
I've updated (but not yet re-published) the docs for this. Thanks for bringing it to my attention even though I wasn't in a good place to listen that day on Discord:)

PR #1654, opened 2025-08-28
Improve Kafka header mapping and ensure reliable Envelope propagation

gardasar-code:
This change strengthens Kafka header handling by reliably mapping Kafka record headers to Wolverine Envelope headers and safely reading/writing individual header values. It ensures predictable conversion between Kafka byte[] headers and string values and guards against null header collections.

Why this is needed:
- Reliability: Prevents header loss.
- Predictable decoding: Provides a deterministic conversion from Kafka byte[] header values to strings, improving interoperability with external Kafka producers/consumers.
- Metadata propagation: Ensures application and system metadata (e.g., correlation, message type, and custom attributes) remain available on the Envelope for logging, tracing, and middleware.
- Transport correctness: Aligns behavior within the Kafka transport so that headers are consistently written and read across ingestion and publishing paths.
- Non-breaking: The changes are internal to the Kafka envelope mapping and do not alter public APIs or configuration surfaces.

Impact:
- More reliable header propagation across Kafka-based message flows.
- Reduced risk of runtime errors from missing header collections.
- Negligible performance impact and no API changes.

### DIFF
--- a/src/Transports/Kafka/Wolverine.Kafka/Internals/KafkaEnvelopeMapper.cs
+++ b/src/Transports/Kafka/Wolverine.Kafka/Internals/KafkaEnvelopeMapper.cs
@@ -28,4 +28,15 @@ internal class KafkaEnvelopeMapper : EnvelopeMapper<Message<string, byte[]>, Mes
         value = default!;
         return false;
     }
+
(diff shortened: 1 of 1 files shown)


jeremydmiller:
@gardasar-code Got it. This will be in either 4.10.1 or 4.11. Not sure yet about the next version number.

Issue #1667, opened 2025-09-02
HandlerGraph.RegisterMessageType race condition


migajek:
Under heavy load of concurrent calls, we observed some of the Requests (InvokeAsync) time out, even though the Reply is properly sent from the request handler. 
For each instance started, if any of the types failed initially, they'd fail for the lifetime of the instance.

Long story short - **HandlerGraph.RegisterMessageType is not thread safe.**
When multiple message types are being invoked at the same time (for the first time), the registrations override each other, because of how the immutable collection is used.

Also, the issue is not self-healing since `_replyTypes` and `_messageTypesLock` go out of sync, but only `_replyTypes` is checked.

```
var graph = host.Services.GetRequiredService<HandlerGraph>();
var runtime = host.Services.GetRequiredService<IWolverineRuntime>();
Parallel.ForEach(typesToRegister, t => runtime.RegisterMessageType(t));

var missingTypes = typesToRegister.Select(t => t.ToMessageTypeName())
    .Where(t => graph.TryFindMessageType(t, out var _) is false).ToArray();
```

Please note - I lack ImTools experience, so I'm not sure `ImTools + lock` is still better choice over regular `ConcurrentDictionary`

jeremydmiller:
Closed by https://github.com/JasperFx/wolverine/pull/1668

PR #1668, opened 2025-09-02
#1667 fix

migajek:
Please note - I lack ImTools experience, so I'm not sure ImTools + lock is still better choice over regular ConcurrentDictionary

### DIFF
--- a/src/Testing/CoreTests/Runtime/Handlers/HandlerGraphTests.cs
+++ b/src/Testing/CoreTests/Runtime/Handlers/HandlerGraphTests.cs
@@ -1,8 +1,12 @@
-using System.Diagnostics;
+using ImTools;
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.Hosting;
 using Module1;
+using System.Diagnostics;
+using System.Reflection;
+using JasperFx.Core.Reflection;
 using Wolverine.Attributes;
+using Wolverine.Runtime;
 using Wolverine.Runtime.Handlers;
 using Wolverine.Util;
 using Xunit;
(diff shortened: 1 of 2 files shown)


jeremydmiller:
@migajek I don't think this code does any harm per se, but can you explain a little bit more what that race condition is? How did you get that?

And ImTools is significantly faster than ConcurrentDictionary, hence it's usage all over Wolverine and Marten

migajek:
Understood. I assumed they're still more performant especially since these collections are mostly read.

For the circumstances - you're gonna hate the answer ;) 
It's a bunch of Wolverine.Http endpoints, each injects IMessageBus and calls `InvokeAsync`. The important bit is, each endpoint sends a different message type and expects different response type.
When exposed to a heavy load of http calls immediately after start, occasionally some of the handlers are being called in parallel for the first time.

The call chain:
MessageBus.InvokeAsync -> MessageRoute.InvokeAsync -> MessageRoute.RemoteInvokeAsync -> bus.Runtime.RegisterMessageType(typeof(T));

Obviously the quick remedy for the issue was to register all  response types upfront, but unless uprfront registration is a requirement, I'd go with bullet-proofing the HandlerGraph

jeremydmiller:
Makes sense, but also:

Just so you and anyone who reads this knows, the Wolverine team most certainly does not recommend building an application like this:

```
The call chain:
MessageBus.InvokeAsync -> MessageRoute.InvokeAsync -> MessageRoute.RemoteInvokeAsync -> bus.Runtime.RegisterMessageType(typeof(T));
```

Just don't do that. Way too much indirection.

migajek:
Thanks, but to make things clear - we're not doing anything like `return bus.InvokeAsync<TResponse>(TRequest)`
The handlers are more complex but first they fetch some auxiliary data from other service for further processing. Hope that's not an anti-pattern, too.

jeremydmiller:
@migajek Just potentially slow. Doing request/reply through Wolverine (really through *any* service bus) is going to incur more latency than an HTTP call would. Probably a little more efficient to go purely async if you can.

Just so you know, I'm going to pull this into main *now* that's for 5.0. I'll cherry pick this over to 4.0 in the next day or two for whatever the next 4.* bug fix release is

migajek:
thank you, both for the advices and the cherry-pick. cheers

PR #1675, opened 2025-09-04
Custom error policy indefinite support

punxrok:
When using custom error policies with `CustomAction`, the action would only execute for the first attempt. On subsequent attempts (attempt 2+), messages would automatically move to the error queue instead of continuing to execute the custom logic. This prevented users from implementing their own retry logic with custom termination conditions.

I added a new `CustomActionIndefinitely` method that sets the `InfiniteSource` to allow custom actions to run indefinitely until the user's logic decides to stop.

### New Method CustomActionIndefinitely

```csharp
opts.Policies.OnException<SpecialException>()
    .CustomActionIndefinitely(async (runtime, lifecycle, ex) =>
    {
        if (lifecycle.Envelope.Attempts > 10)
        {
            runtime.MessageTracking.DiscardedEnvelope(lifecycle.Envelope);
            await lifecycle.CompleteAsync();
            return;
        }

        await lifecycle.ReScheduleAsync(DateTimeOffset.Now.AddSeconds(15));
    }, "Handle SpecialException with conditional discard/requeue");
```


jeremydmiller:
Can you add some text to the documentation and rebase this one first?

punxrok:
Hey, @jeremydmiller. Done.

Issue #1648, opened 2025-08-26
Redo on GCP Transport


CharlieDigital:
**Describe the bug**
Pub/Sub transport eventually fails with the error:

```
The service was unable to fulfill your request. Please try again. [code=8a75]
```

**To Reproduce**

Setup Pub/Sub:

```cs
        builder.UseWolverine(options =>
        {
            options
                .UsePubsub("...")
                .AutoProvision()
                .EnableDeadLettering()
                .EnableSystemEndpoints()
                .PrefixIdentifiersWithMachineName()
                .UseConventionalRouting();
        }
```

Start up the application and observe that the Pub/Sub listener works for a period of time.

After some time, the following errors will occur:

```
[20:18:51.455 ERR] pubsub://gcp-motion-sbx/Charless-MacBook-Pro.Motion.Apps.Features.Examples.Echo: Error while trying to retrieve messages from Google Cloud Platform Pub/Sub, attempting to restart listener (5/5)... ()
Grpc.Core.RpcException: Status(StatusCode="Unavailable", Detail="The service was unable to fulfill your request. Please try again. [code=8a75]")
   at Google.Api.Gax.Grpc.AsyncResponseStream`1.MoveNextAsync(CancellationToken cancellationToken)
   at Wolverine.Pubsub.Internal.BatchedPubsubListener.<>c__DisplayClass1_0.<<StartAsync>b__1>d.MoveNext() in /home/runner/work/wolverine/wolverine/src/Transports/GCP/Wolverine.Pubsub/Internal/BatchedPubsubListener.cs:line 50
--- End of stack trace from previous location ---
   at Wolverine.Pubsub.Internal.PubsubListener.listenForMessagesAsync(Func`1 listenAsync) in /home/runner/work/wolverine/wolverine/src/Transports/GCP/Wolverine.Pubsub/Internal/PubsubListener.cs:line 177
```

For some window of time, it is still possible to send + receive messages.

However, after some additional time, the listening/subscribe side fails as well.

Here is a minimum reproduction with a .NET Web API:

```cs
using Wolverine;
using Wolverine.Pubsub;

var builder = WebApplication.CreateBuilder(args);

// Place your credentials file at the root
Environment.SetEnvironmentVariable(
    "GOOGLE_APPLICATION_CREDENTIALS",
    "application_default_credentials.json"
);

builder.UseWolverine(options =>
{
    options
        .UsePubsub("gcp-motion-sbx")
        .AutoProvision()
[...]

PR #1700, opened 2025-09-16
Switch Pub/Sub implementation to use high level API

cchen-motion:
## Purpose

This PR aims to replace the low-level Pub/Sub API using the `SubscriberServiceApiClient` with the higher level API call `SubscriberClient` which behind the scenes manages instance of `SubcriberServiceApiClient`.

See also: https://github.com/JasperFx/wolverine/issues/1648

## Background

I noticed that when connecting to upstream Pub/Sub, the connection would fail around the 90-100 second mark.  This is because the low-level API is typically managed by the higher level API per Jon Skeet at Google: https://github.com/googleapis/google-cloud-dotnet/issues/15057#issuecomment-3226885267

From the docs: https://cloud.google.com/pubsub/docs/pull#streamingpull-api

> The StreamingPull API keeps an open connection. The Pub/Sub servers recurrently close the connection after a time period to avoid a long-running sticky connection. The client library automatically reopens a StreamingPull connection.
> 
>Messages are sent to the connection when they are available. The StreamingPull API thus minimizes latency and maximizes throughput for messages.

The minimum reproduction of this failure is here: https://github.com/CharlieDigital/dn-pubsub-error-repro but requires testing with an upstream Pub/Sub connection to observe the failure.


jeremydmiller:
@cchen-motion

image

CharlieDigital:
Let me spend a bit more time with this to do some testing, but please review and let me know if I've made some incorrect assumptions about how Wolverine internals work.

src/Transports/GCP/Wolverine.Pubsub/Internal/PubsubListener.cs
```diff
@@ -232,61 +165,55 @@ protected async Task listenForMessagesAsync(Func<Task> listenAsync)
...
-                var batched = EnvelopeSerializer.ReadMany(message.Message.Data.ToByteArray());
+                await _receiver.ReceivedAsync(this, batched);
+            }
 
-                if (batched.Any())
-                {
-                    await _receiver.ReceivedAsync(this, batched);
-                }
+            return;
+        }
```
CharlieDigital, line 182: Not certain if this batching piece is still relevant with the higher level API

src/Transports/GCP/Wolverine.Pubsub/Internal/BatchedPubsubListener.cs
```diff
@@ -25,39 +24,47 @@ public override async Task StartAsync()
...
-            await listenForMessagesAsync(async () =>
+            // Start the subscriber and capture the lifetime task
+            var subscriberLifetime = subscriberClient.StartAsync(async (msg, ct) =>
             {
-                while (await stream.MoveNextAsync(_cancellation.Token))
-                {
-                    await handleMessagesAsync(stream.Current.ReceivedMessages);
-                }
+                await handleMessagesAsync(msg);
+                return SubscriberClient.Reply.Ack;
```
CharlieDigital, line 52: Acking should technically happen automatically here in this lambda on the return of `SubscriberClient.Reply.Ack` so it seems like this removes some of the code that was previously managing the Ack at a lower level, but not clear if that code I removed is "structurally required" by Wolverine.

CharlieDigital:
Current state of tests:

image

Are there any tips to get the tests to run faster?  Currently taking a really, really long time to run the suite just for GCP

jeremydmiller:
@cchen-motion Hate to admit that I paid more attention to this today because someone was ripping Wolverine because of the GCP transport today. Do you want me to take it the rest of the way, or just get into the questions?

Some of the test slowness is some compliance tests that wait for draining, and I've never taken any time to look at why that's slow

CharlieDigital:
I'm rethinking the implementation move to the high level lib as it may not mesh well with the overall architecture/design goals of Wolverine.  Unfortunately, it was not easy to deduce the right way that this should work early on.

It may be that the correct move here is to change the error handling logic that was originally implemented which would fail after a certain number of reconnects.  The key here is that these transient disconnects will not happen when testing with the emulator as the behavior here diverges from IRL Pub/Sub upstream in GCP.

I looked at the Brighter implementation of their Pub/Sub transport and it's quite good, actually.

https://github.com/BrighterCommand/Brighter/pull/3645/files#diff-b803fcb7f17ed9235f1e5cb1fcd2f5d3b2838429d4368ae4c57ce4436577f03f

Author implemented using idiomatic parts of the platform (e.g. clever use of Firestore atomic writes to perform distributed locking and using Firestore/Spanner as outbox).

I think the right course of action:

- Determine if the high level client API fits into Wolverine's architecture/design.  Mainly the high level API automatically handles ACKs and connection pooling.  Does this fit into how Wolverine is designed to pull and consume messages?
- If the high level client API does not fit into the intended design, then the question is how to make the intentional server disconnect handling correct in the Wolverine client implementation

The high level client API is quite stable and offers very high throughput in some testing I've done over the last week for different purposes.  I just don't know if it fits well into your design goals.

I have two sandboxes here I was using for testing that you may find helpful:

- https://github.com/CharlieDigital/dn-pubsub-latency (latency+throughput testing for the high level client API)
- https://github.com/CharlieDigital/dn-pubsub-error-repro (minimum repro for the low-level disconnect if you want to understand the mode of failure)

I would also encourage you to give this thread a look: https://github.com/dotnet/aspnetcore/issues/53219#issuecomment-3346666458.  Not sure about your thoughts here, but I think it would be a huge boon for the community and library owners like you if Microsoft separated out the front-side scaffolding for Orleans into a set of abstractions.  Then, for example, if someone implements a NATS messaging transport, it would "just work" for Wolverine assuming Wolverine adopted the abstractions.

jeremydmiller:
@cchen-motion I honestly don't know. This is one of the very few parts of Wolverine where I wasn't terribly involved. 

**YOU** are the guy who was bashing Wolverine on that thread? 

"I think it would be a huge boon for the community and library owners like you if Microsoft separated out the front-side scaffolding for Orleans into a set of abstractions."

I don't want anything at all to do with that. I don't think that effort would provide any real value and probably end up being a huge waste of time for people like me, but I get why that might sound attractive. 

At this point, I would prefer that you just move on, go use something else (Brighter is fine, if much more ceremony than Wolverine), and stay off of both the Wolverine GitHub and Discord. I don't know how much of a priority I feel the GCP transport is or even the Kafka usage you were after.

jeremydmiller:
Hey all, I'll take this on soon in an independent effort.

CharlieDigital:
@jeremydmiller I'm surprised that you would view that as "bashing"; in its current state, the Pub/Sub implementation will only work with the emulator because it does not correctly handle the server disconnect signals.  I'm merely sharing my experience about the state of messaging on .NET.  It seems that individual projects not having to re-invent the wheel with it comes to "read message here, write message there" would be a big boon for platform authors.

I'm not trying to bash or pick on any particular team; I'm merely here exploring the options available as objectively as I can.

I'm very surprised that you see any malicious intent or snide in any of the feedback I've provided.

jeremydmiller:
@cchen-motion Hey man, you're being nothing but overhead and a distraction. If you're unhappy w/ Wolverine's implementation, and you're not able to finish this PR, I'd ask that you go somewhere. I'm irritated at the interactions with you on the Kafka usage as well and I think you're going to be difficult to interact with. I am still asking you to just go use something else.  

I'm very sure that when I have some bandwidth, I can quickly address the GCP transport issues. To be honest with you though, there's very little demand for it and there are more pressing matters.

PR #1707, opened 2025-09-18
EntityFrameworkCore optimistic concurrency support for saga

JMounier:
This PR adds support for optimistic concurrency when a saga is persisted using the EFCore integration (implementation based on concurrency token cf. [Handling Concurrency Conflicts](https://learn.microsoft.com/en-us/ef/core/saving/concurrency)). It patches the code generation to:
- Ask EF if the saga entity has any pending changes and if so increment the internal saga version number.
- Throw a `SagaConcurrencyException` when `SaveChangesAsync` throws a `DbUpdateConcurrencyException` impacting the actual saga entity.
- Documentation was updated to recommend flagging the `Version` property as a concurrency token.

---

P.S. I tried running the `./build.sh` to test my change but I have the whole `HttpTests` suite failing for unclear reasons and `CodegenPreviewCommand` failing to compile with the following error:
> Method 'get_LockReleaseBehavior' in type 'Npgsql.EntityFrameworkCore.PostgreSQL.Migrations.Internal.NpgsqlHistoryRepository' from assembly 'Npgsql.EntityFrameworkCore.PostgreSQL

### DIFF
--- a/docs/guide/durability/efcore/sagas.md
+++ b/docs/guide/durability/efcore/sagas.md
@@ -112,9 +112,12 @@ public class OrdersDbContext : DbContext
         modelBuilder.Entity<Order>(map =>
         {
             map.ToTable("orders", "sample");
-            map.HasKey(x => x.Id);
             map.Property(x => x.OrderStatus)
                 .HasConversion(v => v.ToString(), v => Enum.Parse<OrderStatus>(v));
+
+            // enable optimistic concurrency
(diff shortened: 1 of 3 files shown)


jeremydmiller:
@JMounier Hey, I did the trick of pulling in the PR, *then* running the tests. Next time your shop tries a pull request, just make sure to run the test suite locally please? It's 95% there, but the codegen was off.

PR #1762, opened 2025-10-16
Add support for named Azure Service Bus brokers

WillemCoudron:
Exposes AzureServiceBusTransport constructor as public and extends extension methods to support multiple named brokers via BrokerName. Adds new methods for configuring, listening, and publishing to Azure Service Bus resources on specific brokers, enabling advanced multi-broker scenarios.


jeremydmiller:
@WillemCoudron Got it. I'll try to have this worked in prior to the 5.0 release, so sometime by Monday

jeremydmiller:
@WillemCoudron I'm going to take a run at Azure Service Bus work tomorrow and get this in as part of that

PR #1771, opened 2025-10-19
Add ability to pass arguments to the RabbitMQ consume method

rdm1234:
Add ability to pass arguments to the RabbitMQ channel `BasicConsumeAsync` method. It can be used to customize stream consumption behavior. 

Example use case: `q.ConsumerArguments["x-stream-offset"] = "first";`


jeremydmiller:
@rdm1234 That looks fine, but could you add at least a smoke test and/or docs?

jeremydmiller:
@rdm1234 Don't worry about it, I'll get this

PR #1778, opened 2025-10-23
Add RabbitMqExchange.DisableAutoProvision

mac-michael:


jeremydmiller:
What's the point of this? Just to be able to declare queues or bindings, but not exchanges?

mac-michael:
That's right. 

We've got a multi-host scenario where we don't want to create exchanges that we use as a client. We want only the owner of the given vhost to create them.

mac-michael:
Thanks.

PR #1786, opened 2025-10-27
Allow TCP transport to listen on a specific hostname/address

e-tobi:
Instead of localhost (the default) allow the TCP transport to bind to any hostname / address.


jeremydmiller:
Hey, I'll see about getting a version of this in soon. Sorry, and I know this is annoying, but we'll need to do an overload. A new optional arg is actually a breaking change in .NET world

jeremydmiller:
I'm doing this separately w/ an overload.

PR #1792, opened 2025-10-28
Add NSwag note to OpenAPI Metadata page

benv-nti:
This seemed like the best place to put this.  If you have another spot in mind, or any wordsmithing, let me know.


jeremydmiller:
@benv-nti That's good for now. Think I'd like to take care of that a little more permanently soon though.

Issue #1796, opened 2025-10-29
[Log] Spam of IAgentCommand


MichalNemec:
**Describe the bug**
Upgraded from 4.12.3 to 5.0.0 and suddenly i have full log of IAgentCommand processed message:
```
[19:36:47 INF] Successfully processed message Wolverine.Runtime.Agents.IAgentCommand#08de171a-1e5e-797e-8ee5-e9ecd0eb0000 from dbcontrol://ccb366d3-ca3d-4e44-a11a-61ac857d9769/
[19:36:48 INF] Successfully processed message Wolverine.Runtime.Agents.IAgentCommand#08de171a-1ef7-4ae6-8ee5-e9ecd0eb0000 from dbcontrol://ccb366d3-ca3d-4e44-a11a-61ac857d9769/
[19:36:49 INF] Successfully processed message Wolverine.Runtime.Agents.IAgentCommand#08de171a-1f90-2a7c-8ee5-e9ecd0eb0000 from dbcontrol://ccb366d3-ca3d-4e44-a11a-61ac857d9769/
[19:36:50 INF] Successfully processed message Wolverine.Runtime.Agents.IAgentCommand#08de171a-2028-fa0e-8ee5-e9ecd0eb0000 from dbcontrol://ccb366d3-ca3d-4e44-a11a-61ac857d9769/
[19:36:51 INF] Successfully processed message Wolverine.Runtime.Agents.IAgentCommand#08de171a-20c1-b528-8ee5-e9ecd0eb0000 from dbcontrol://ccb366d3-ca3d-4e44-a11a-61ac857d9769/
[19:36:52 INF] Successfully processed message Wolverine.Runtime.Agents.IAgentCommand#08de171a-215a-84a6-8ee5-e9ecd0eb0000 from dbcontrol://ccb366d3-ca3d-4e44-a11a-61ac857d9769/
[19:36:53 INF] Successfully processed message Wolverine.Runtime.Agents.IAgentCommand#08de171a-21f3-41aa-8ee5-e9ecd0eb0000 from dbcontrol://ccb366d3-ca3d-4e44-a11a-61ac857d9769/
[19:36:54 INF] Successfully processed message Wolverine.Runtime.Agents.IAgentCommand#08de171a-228c-7a5a-8ee5-e9ecd0eb0000 from dbcontrol://ccb366d3-ca3d-4e44-a11a-61ac857d9769/

```

**To Reproduce**
```
[...]

PR #1814, opened 2025-11-05
fix: Prevent AgentCommandHandler Chain from being overwritten

akselkvitberg:
This caused the LogLevel set in AgentCommandHandler to be overwritten.

Fixes: #1796


jeremydmiller:
Doh! Good catch

PR #1868, opened 2025-11-26
Add Discovery Property for DisableConventionalDiscovery call

niekschoemaker:


jeremydmiller:
Meh, there's a helper extension method in the testing projects just like that, so that's why it snuck into the sample code

PR #1891, opened 2025-12-02
Endpoint with [FormData] and TenantId parameter fails with 415 http error

marcpiechura:
When using the FormData attribute and the TenantId parameter, codegen tries to create the TenantId from the json body and therefore requests with form data fail since the endpoint expects json. As a workaround, one can use the [FromServices] attribute on the TenantId parameter. 
Not sure where to fix that, but at least you've a failing test ;)


jeremydmiller:
Failing tests are the best repro steps. Will get this in soon-ish

jeremydmiller:
@marcpiechura That was an awful bug, thanks for the repro. Nothing to do w/ multi-tenancy per se.

Fix will be in 5.7:

https://github.com/JasperFx/wolverine/pull/1914

Issue #1924, opened 2025-12-10
Accepting non-nullable string query parameters violates nullable annotation if no query string is passed


Blackclaws:
**Describe the bug**
When defining a Wolverine handler that takes a string from a query parameter there is a nullability annotation violation that leads to NullReferenceExceptions because StringValues implictly converted to string with an empty array leads to null instead of an empty string.

**To Reproduce**
1. Create a Wolverine handler like this:
```
    [WolverineGet("/test")]
    public static async Task<Ok<string>> Test([FromQuery]    string  ids) {
       return ids.ToString();
    }
```
2. Call it without supplying the ids query parameter
3. NullReferenceException ensues

**Expected behavior**
There is either a bad request response created while the parameters are extracted from the Query, or alternatively empty StringValues are converted to an empty string.

**Additional context**
The reason is that the generated code uses StringValues from the httpContext.Request.Query directly and implicitly converts it to string, which is a nullability violation.

PR #1926, opened 2025-12-10
fix: Non-Nullable String empty query parameter

Blackclaws:
Fixes: https://github.com/JasperFx/wolverine/issues/1924

Substitutes the empty string for non existing query parameters when non-nullable string is requested.


jeremydmiller:
@Blackclaws Thank you!

PR #1957, opened 2025-12-14
Saga identity from attribute

Anton0123:
- **feat: saga identity from attribute**
- **test**


jeremydmiller:
Hey, can you just reuse [SagaIdentity] here instead of an all new attribute?

Anton0123:
@jeremydmiller 
> Hey, can you just reuse [SagaIdentity] here instead of an all new attribute?

I could but I only want the property to be available when the attributes is placed on a parameter (doesn't make sense in other contexts). 

Want me to move it still and should I throw an exception if the context is not a parameter or just ignore it if it's set anyways?

jeremydmiller:
@Anton0123 Nah, let's take it as is. I'll add a smidge of docs for it in a bit

PR #1959, opened 2025-12-14
refactor(durability): add options for disabling or delaying node assignment traces

ross1296:


jeremydmiller:
Got this with a local rebase. Thanks @ross1296!

Issue #1828, opened 2025-11-10
Support DataAnnotations for validation


rigtigeEmil:
**Is your feature request related to a problem? Please describe.**
The Fluent Validation support works great, but it can be tedious ander overkill to write simple validation rules that are supported with DataAnnotations (or through custom ones).

**Describe the solution you'd like**
Support for DataAnnotations similar to what's supported in [.NET 10 for Minimal API's](https://learn.microsoft.com/en-us/aspnet/core/release-notes/aspnetcore-10.0?view=aspnetcore-10.0#validation-support-in-minimal-apis).

LodewijkSioen:
I can contribute that if you're interested. I already have the code laying around.

jeremydmiller:
@LodewijkSioen Would you mind opening a PR for whatever you have?

If I was more ambitious than I feel this week, I'd also pair that w/ a new guide on validation

LodewijkSioen:
Here you go. The documentation is a bit rough.

d-stanescu:
Hello! This is a nice feature I'd like to use. The [DataAnnotations Validation Middleware](https://wolverine.netlify.app/guide/handlers/dataannotations-validation.html) mentions the `WolverineFx.DataAnnotationsValidation` will add support for the built-in and custom attributes via middleware. But the package is not yet published on https://www.nuget.org/packages?q=WolverineFx.DataAnnotationsValidation. Any idea about when the package will be published? Also, there's a broken link in the TIP section to this page https://wolverine.netlify.app/guide/http/dataannotationsvalidation.html.

PR #1966, opened 2025-12-16
Add DataAnnotations Validation middlewares

LodewijkSioen:
See #1828

Implements DataAnnotations Validation. Heavily borrowed from the FluentValidation implementation.

The documentation is a bit rough.


jeremydmiller:
Awesome! I'll deal w/ whatever the build failure thing is. I *might* pull these into the core HTTP lib, but let me think on that a bit.

Thank you!

PR #1971, opened 2025-12-17
Add HTTP transport support with client and server integration

sdesaiLULA:
This pull request introduces a new HTTP transport implementation and associated tests for the Wolverine framework. It adds a public `HttpTransport` class, a `WolverineHttpTransportClient`, and comprehensive test coverage for both. Additionally, a sample external HTTP server is provided to demonstrate usage. The main themes are the implementation of HTTP transport, testing, and sample usage.

**HTTP Transport Implementation:**

* Introduced a public `HttpTransport` class that manages HTTP endpoints, caches them, and supports initialization and endpoint retrieval by URI. The transport now uses HTTPS as its protocol and is derived from `TransportBase<HttpEndpoint>`.
* Added a public `WolverineHttpTransportClient` class that uses `IHttpClientFactory` to send single envelopes or batches of messages over HTTP, properly serializing content and setting content types.

**Testing:**

* Added `HttpTransportTests` to ensure correct protocol usage, endpoint caching, retrieval, and initialization logic for the HTTP transport.
* Added `WolverineHttpTransportClientTests` to verify correct HTTP requests are made for both single envelopes and batches, including a mock HTTP handler for inspection.

**Sample Usage:**

* Added `ExternalHttpServer.cs`, a sample ASP.NET Core application showing how to configure Wolverine with HTTP transport, set up named HTTP clients, and define message handlers and endpoints.

### DIFF
--- a/src/Http/Wolverine.Http.Tests/Samples/ExternalHttpServer.cs
+++ b/src/Http/Wolverine.Http.Tests/Samples/ExternalHttpServer.cs
@@ -1,0 +1,62 @@
+using Microsoft.AspNetCore.Builder;
+using Microsoft.AspNetCore.Http;
+using Microsoft.Extensions.DependencyInjection;
+using Wolverine;
+using Wolverine.Http;
+using Wolverine.Http.Transport;
+
+var builder = WebApplication.CreateBuilder(args);
+const string httpNamedClient = "https://extenal/";
+builder.UseWolverine(opts =>
+{
+    var transport = new HttpTransport();
+    opts.Transports.Add(transport);
+    // Publish all messages to the external http endpoint using the named http client
+    // The .SendInline() method fails, so do not use it here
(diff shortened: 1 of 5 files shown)


sdesaiLULA:
We are looking at [qstash](https://upstash.com/docs/qstash/overall/getstarted) as an alternative to our current Postgresql transport. We use Marten and wolverine, so I thought I'd enhance the existing http transport so we get all the benefits of wolverine and still use qstash.

jeremydmiller:
@sdesaiLULA oh, nice. This has been on my deep TODO list. The IHttpClientFactory usage will allow users to add security, right?

sdesaiLULA:
@jeremydmiller  yes. The users can setup a http named client with any security scheme and pass it to httpTransport.

jeremydmiller:
@sdesaiLULA Bueno, turns out I need to push 5.9 for a client today, so this will get bundled in later today. This will actually get used in CritterWatch down the line.

PR #2008, opened 2026-01-06
Add Azure SignalR to SignalR transport

PerfectlyNormal:
Since there's almost no difference between the two I just added the `Microsoft.Azure.SignalR` nuget directly to `Wolverine.SignalR`. If you'd rather have a separate package that extends `Wolverine.SignalR`, just say the word (and should it then be Wolverine.AzureSignalR, or Wolverine.SignalR.Azure?)

I haven't added any automated tests, since I'm a bit unsure how to test SignalR properly. None of the existing SignalR-tests have broken. Some manual testing shows that messages gets sent correctly between clients.


jeremydmiller:
I don't think it hurts anything to add this, and that's come up a few times.

Thank you!

PR #2009, opened 2026-01-06
Correct parameter name (`complete` instead of `start`)

rdeveen:
This pull request makes a minor fix to the `OrderSagaSample` API endpoint. The change corrects the parameter name in the `/complete` route handler to ensure the correct object is passed to the message bus.

- Fixed the `/complete` endpoint in `Program.cs` to use the correct parameter name (`complete` instead of `start`) when invoking the message bus.


jeremydmiller:
Thank you!

PR #2018, opened 2026-01-12
Return the correct error message when ISideEffect is directly returned by a handler

elexisvenator:
The wrong type check was being done here, leading to a different, less clear method about what is going on (the one about the missing Sync/SyncAsync method). Doesn't change any behaviour for working handlers.


jeremydmiller:
image

Good catch!

PR #2038, opened 2026-01-15
Add global timeout option for remote invocation

codingzerotohero:
Added an option to set global default timeout for remote invocation in `WolverineOptions`, as well as test in `CoreTests.Acceptance.remote_invocation` - to ensure that it's respected.

The global timeout option will be overriden when `InvokeAsync` is called with a timeout parameter, allowing per call overrides if desired.


codingzerotohero:
@jeremydmiller Hey! Any chance of this getting added to upcoming release? I'm currently integrating Wolverine to replace MassTransit for a client and I'd love to NOT have to add a timeout on each of the hundreds of remote `bus.InvokeAsync<T>` calls they've got..

jeremydmiller:
@codingzerotohero Yes. I'm working on Wolverine today. Still needing some client input for a potential change, but they're being slow so maybe that has to wait for 5.12.

PR #2041, opened 2026-01-18
Fix missing transaction commit for multi-tenanted EF Core

### LINKED ISSUE #2040: Cascading IEnumerable<object> starts an EF transaction during EnqueueCascadingAsync but does not commit
When using Wolverine managed EF Core multi-tenancy + durable local queues + InvokeForTenantAsync(request/reply), returning domain events as the second item of:
```
Task<(TResponse, IEnumerable<object>)>
```
causes Wolverine to start an EF Core transaction during `context.EnqueueCascadingAsync(outgoing2)`, but the generated handler code only calls `SaveChangesAsync()` and never commits the started transaction. The DbContext is disposed at the end of the handler, so the transaction is rolled back and the entity insert does not persist.

Additionally, even though the transaction is rolled back (customer row not stored), the cascaded message handler (`CustomerCreatedHandler`) still executes. With `UseDurableLocalQueues()` and transactional outbox, we would expect the cascaded message to not be dispatched if the transaction is not committed.

Versions / environment
- WolverineFx: 5.10.0
- WolverineFx.EntityFrameworkCore: 5.10.0
- WolverineFx.SqlServer: 5.10.0
- EF Core: 10.0.2
- SQL Server
- Target framework: net10.0

Expected behavior
- When handler returns `(result, [new DomainEvent(...)])`:
  1. the customer insert is committed, and
  2. the domain event is stored in the outbox and later dispatched.
- If the transaction rolls back, the domain event should not be dispatched (durable local).

Actual behavior
- When the cascading enumerable contains at least one event:
  - transaction is started during `EnqueueCascadingAsync`
  - generated adapter ends with `SaveChangesAsync()` only, no explicit commit
  - customer insert is rolled back (row missing in DB)
  - but `CustomerCreatedHandler` still fires (unexpected under durable local/outbox)
- When the handler returns an empty enumerable (`[]`), the customer insert commits normally.

Repro project
[MultiTenantMessaging.zip](https://github.com/user-attachments/files/24689487/MultiTenantMessaging.zip)

Steps to reproduce
1. Set two connection strings in appsettings.json:
    - ConnectionStrings:main = Wolverine message store DB
    - ConnectionStrings:tenant1 = tenant DB that contains Customers
2. Run the app.

The app invokes two tenant commands:
- `CreateCustomerNoEvents (returns `[]`)` - persists
- `CreateCustomerWithEvents (returns [new CustomerCreated(...)])` - customer row is NOT persisted, but `CustomerCreatedHandler` still runs.

Minimal handler code
```csharp
[Transactional]
[WolverineHandler]
public sealed class CreateCustomerWithEventsHandler(ICustomerRepository repository)
{
    public async Task<(CreateCustomerResult, IEnumerable<object>)> Handle(
        CreateCustomerWithEvents request,
        CancellationToken ct)
    {
        var customer = new Customer { Id = Guid.NewGuid(), Name = request.Name };

        await repository.AddAsync(customer, ct);

        return (new CreateCustomerResult(customer.Id, customer.Name),
            [new CustomerCreated(customer.Id)]);
    }
}

```

Generated handler adapter
`Internal/Generated/WolverineHandlers/CreateCustomerWithEventsHandler*.cs:`

```csharp
// Outgoing, cascaded message
await context.EnqueueCascadingAsync(outgoing2).ConfigureAwait(false);

// Added by EF Core Transaction Middleware
var result_of_SaveChangesAsync = await tenantDbContext.SaveChangesAsync(cancellation).ConfigureAwait(false);
```

`EnqueueCascadingAsync(outgoing2)` appears to start an EF transaction (via `EfCoreEnvelopeTransaction`), but there is no explicit commit afterward. Because the DbContext is disposed at the end of the generated handler method, the transaction is rolled back and the customer row is not saved.

Despite that, `CustomerCreatedHandler` still fires, seems to violate the durable local/outbox expectation that outgoing messages are only dispatched after successful commit.

PerfectlyNormal:
Fixes #2040, at least in my test and the repro project

This should probably get a thorough review, as I feel I'm deeper in Wolverine than I should be and this could possibly affect other things. Feels a bit like a "works on my machine, now you try".

Mostly copied from `EnrollDbContextInTransaction`, but without the `IDomainEventScraper` bits. Started to look into it, but a bit unsure how it should work

### DIFF
--- a/src/Persistence/EfCoreTests/MultiTenancy/MultiTenancyCompliance.cs
+++ b/src/Persistence/EfCoreTests/MultiTenancy/MultiTenancyCompliance.cs
@@ -179,6 +179,39 @@ public abstract class MultiTenancyCompliance : IAsyncLifetime, IWolverineExtensi
         }
     }
 
+    [Fact]
+    public async Task end_to_end_with_cascading_messages()
+    {
+        var blueId = Guid.NewGuid();
+        var redId = Guid.NewGuid();
+        var greenId = Guid.NewGuid();
+
+        await theHost.InvokeMessageAndWaitAsync(new StartAndTriggerApproval(blueId, "Blue!"), "blue");
+        await theHost.InvokeMessageAndWaitAsync(new StartAndTriggerApproval(redId, "Red!"), "red");
+        await theHost.InvokeMessageAndWaitAsync(new StartAndTriggerApproval(greenId, "Green!"), "green");
+
+        var blueDbContext = await theBuilder.BuildAsync("blue", CancellationToken.None);
+        var greenDbContext = await theBuilder.BuildAsync("green", CancellationToken.None);
+        var redDbContext = await theBuilder.BuildAsync("red", CancellationToken.None);
+
+        var blue = await blueDbContext.Items.FindAsync(blueId);
+        blue.Name.ShouldBe("Blue!");
+        blue.Approved.ShouldBeTrue();
+        (await greenDbContext.Items.FindAsync(blueId)).ShouldBeNull();
+        (await redDbContext.Items.FindAsync(blueId)).ShouldBeNull();
+
+        (await blueDbContext.Items.FindAsync(redId)).ShouldBeNull();
+        (await greenDbContext.Items.FindAsync(redId)).ShouldBeNull();
+        var red = await redDbContext.Items.FindAsync(redId);
+        red.Name.ShouldBe("Red!");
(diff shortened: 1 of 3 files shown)


jeremydmiller:
@PerfectlyNormal I just added the new idempotency check. Need to also add the new domain event scraping to the multi-tenancy path. Missed that when I was adding it before.

Thank you!

PR #2052, opened 2026-01-21
Include custom virtual host setup for RabbitMQ

mj2015:
Added instructions for using a custom virtual host in RabbitMQ.


jeremydmiller:
👍

PR #2070, opened 2026-01-24
Provide extension methods to get Aggregate Id Variable in HTTP Chain Policy

flew2bits:
Add extension methods for HttpChain to retrieve the AggregateId variable for an Wolverine endpoint decorated with AggregateHandlerAttribute.


flew2bits:
I'm not sure why the `CoreTests.Acceptance.middleware_usage.pass_objects_from_and_to_middleware` test is failing. I ran the test locally and it passes for all frameworks.

PR #2072, opened 2026-01-26
Change multiple flush to an information as you have to be opted into …

thechucklingatom:
…with the AllowMultiple setting


BrianVallelunga:
I'd bump this down to Debug actually, since otherwise it is just spamming the logs.

Issue #2079, opened 2026-01-28
MQTT Messages from Azure Event Grid can't be received


t0ms3n:
**Describe the bug**
Messages from an Azure Event Grid MQTT broker can not be received, because UserProperties of the MqttApplicationMessage are null instead of an empty list when a incoming message is mapped in the MqttEnvelopeMapper.

These properties aren't sent and a [document limitation](https://learn.microsoft.com/en-us/azure/event-grid/mqtt-support#mqtt-v5-current-limitations).

**To Reproduce**
1. Connect to an azure event grid namespace with MQTT Mode via WebSocketServer
2. Use OAUTH2-JWT authentication
3. Listen to a topic
    opts.ListenToMqttTopic("space/#")
        .DefaultIncomingMessage<DeviceEdgeMessage>()
        .QualityOfService(MqttQualityOfServiceLevel.AtMostOnce);  

**Expected behavior**
Handle null like an empty list.

**Version**
WolverineFx.MQTT 5.12.1

**Stack Trace**
System.NullReferenceException: Object reference not set to an instance of an object.
   at Wolverine.MQTT.Internals.MqttEnvelopeMapper.MapIncomingToEnvelope(Envelope envelope, MqttApplicationMessage incoming) in /home/runner/work/wolverine/wolverine/src/Transports/MQTT/Wolverine.MQTT/Internals/MqttEnvelopeMapper.cs:line 110
   at Wolverine.MQTT.Internals.MqttListener.ReceiveAsync(MqttApplicationMessageReceivedEventArgs args) in /home/runner/work/wolverine/wolverine/src/Transports/MQTT/Wolverine.MQTT/Internals/MqttListener.cs:line 97

**Screenshots**
<img width="1237" height="556" alt="Image" src="https://github.com/user-attachments/assets/f51f30ad-6b8c-418c-8c1e-9b32a0c00337" />

jeremydmiller:
Hoping this is closed by https://github.com/JasperFx/wolverine/pull/2083

PR #2083, opened 2026-01-28
Add support to refresh JWT tokens for MQTT transport

t0ms3n:
Currently we only can authenticate once by providing the token with the options by using 
```.WithAuthentication("OAUTH2-JWT", Encoding.UTF8.GetBytes(accessToken.Token))```. This change provides another parameter to `.UseMqtt(...)`, where we define a callback function to fetch the token and a timespan when the token should be refreshed.
If the parameter is set, the authentication is automatically set to "OAUTH2-JWT" and an initial token is fetched on transport initialization.

Also #2079 is fixed, by adding a null-check on the user options of incoming messages.


jeremydmiller:
@t0ms3n Hey, I'm struggling to keep up w/ Wolverine prs and fixes right now. Trying to get this in reasonably soon, but it might not make 5.14

PR #2103, opened 2026-02-04
Add support for SignalR hubs requiring authorization

PerfectlyNormal:
Not really happy with the error handling in `SignalRClientEndpoint.BuildListenerAsync`. Since we try connecting immediately any errors prevents us from starting. The most common error will (probably) be unauthorized, so catching and logging that. But there's probably a lot better ways of handling this.

SignalR expects a `Func<Task<string?>>` as a token provider. I wrapped that so we get access to the `IServiceProvider` to get configuration or anything like that. That turns into a `Func<IServiceProvider, Func<Task<string?>>>` which isn't very pretty.

Feel free to edit or suggest improvements


jeremydmiller:
@PerfectlyNormal One thing at a time. Let's go w/ what you've done and just collect feedback from users.

Issue #1708, opened 2025-09-18
Event lost using inbox and kafka when database is offline


dominikjeske:
**Describe the bug**

Message is consumed in [KafkaListener](https://github.com/JasperFx/wolverine/blob/f36e4bb31b15c88c55bdc28300b9216fcde52fba/src/Transports/Kafka/Wolverine.Kafka/Internals/KafkaListener.cs) and offset is commited in kafka before message is stored in "incomming_envelopes" table.
When topic is configured to use kafka with inbox there is following code flow

- Message is consumed in [KafkaListener](https://github.com/JasperFx/wolverine/blob/f36e4bb31b15c88c55bdc28300b9216fcde52fba/src/Transports/Kafka/Wolverine.Kafka/Internals/KafkaListener.cs#L60)
- Next there is [DurableReceiver](https://github.com/JasperFx/wolverine/blob/f36e4bb31b15c88c55bdc28300b9216fcde52fba/src/Wolverine/Runtime/WorkerQueues/DurableReceiver.cs#L226) that tries to execute StoreIncomingAsync when this task is failing because of DB not working it gives back executions to KafkaListener and post async action to [retry block](https://github.com/JasperFx/wolverine/blob/V4.12.1/src/Wolverine/Util/Dataflow/RetryBlock.cs#L137)
- This is problem because message is comited in [KafkaListener](https://github.com/JasperFx/wolverine/blob/f36e4bb31b15c88c55bdc28300b9216fcde52fba/src/Transports/Kafka/Wolverine.Kafka/Internals/KafkaListener.cs#L47) before message is inside inbox. When you are lucky retry block will succeed but in other cases after 3 tries it is lost (you have only information in logs)
- Smaller issue is that [activity](https://github.com/JasperFx/wolverine/blob/f36e4bb31b15c88c55bdc28300b9216fcde52fba/src/Wolverine/Runtime/WorkerQueues/DurableReceiver.cs#L264) is ended before processing eds
- If the process crashes all the messages that are in memory will be lost but kafka offsets are set and will newer be processed by the business handler - so we don't have "at least once" processing behavior. Additionally what will happen when we will have 1M messages - all will enter retry blocks?

 
**To Reproduce**
- Configure endpoint for kafka and durable inbox with SQL server
- Start application (when DB is working)
- Disable DB
- Send some message to kafka topic
- Message will be download from kafka and lost after retries

**Expected behavior**
Message should not be committed before insert to inbox will succeed

**Additional context**
[...]

PR #2114, opened 2026-02-04
Add EnableAtLeastOnceDelivery() option for Kafka transport

vgmello:
This PR introduces an explicit opt-in mechanism for achieving "at-least-once" delivery semantics when consuming messages from Kafka topics. By default, Kafka's consumer stores offsets immediately upon message consumption, which can lead to message loss if the application crashes during processing. 

### Problem

With the default Kafka consumer configuration (EnableAutoOffsetStore=true), the offset is stored as soon as a message is consumed—before the handler processes it. If the application crashes during message processing, the message is lost because Kafka thinks it was already handled.

### Solution

Added a new EnableAtLeastOnceDelivery() fluent API method that:

1. Automatically sets EnableAutoOffsetStore=false on the consumer config
2. Stores the offset only after ReceivedAsync completes, which means:
   - For ProcessInline() mode: after the message handler finishes
   - For UseDurableInbox() mode: after the message is persisted to the database inbox

Important Notes
- Backward compatible: Default is false, so existing behavior is unchanged
- Does NOT guarantee at-least-once with BufferedInMemory(): Since messages are buffered in memory before processing, they can still be lost on crash
- Offset storage is in-memory (default Kafka Client Behaviour): StoreOffset stores offsets in memory on the consumer. The actual commit to Kafka happens at the EnableAutoCommit interval (default 5 seconds) or after processing when auto-commit is disabled. Messages may be redelivered if the app crashes between storing and committing.

This PR could potentially fix https://github.com/JasperFx/wolverine/issues/1708

### DIFF
--- a/PLAN.md
+++ b/PLAN.md
@@ -1,0 +1,205 @@
+# Plan: Sending Failure Policies (Issue #1686)
+
+## Problem
+
+When an outbound message is rejected by a transport/broker (e.g., Kafka message too large, Azure SB size limit), the durable sending agent retries it forever. There's no way to discard, move to DLQ, or apply exception-specific policies to **sending** failures. The existing error handling model only applies to **handler** (incoming message) failures.
+
+Additionally, Azure Service Bus `TryAddMessage` silently fails for oversized messages, causing message loss without any error.
+
+## Design Decisions (from clarification)
+
+- **Error queue**: Default to `wolverine_dead_letters` table; allow transports to override with native DLQ
+- **Policy scope**: Global defaults on `WolverineOptions`, with per-endpoint overrides
+- **Attempt tracking**: New `Envelope.SendAttempts` counter (separate from handler `Attempts`)
+- **Batch failures**: Apply matched continuation to entire failed batch
+- **Lifecycle**: Extend existing `IEnvelopeLifecycle` so existing `IContinuation` implementations can be reused
+- **Latch sender**: Standalone + composable (like `PauseListenerContinuation`)
+- **Raise message**: Base on existing `UserDefinedContinuation`/`CustomAction` pattern
+- **Oversized detection**: Include detection of messages that can never fit in a batch
+
+## Implementation Steps
+
+### Step 1: Add `Envelope.SendAttempts` property
+
+**File**: `src/Wolverine/Envelope.cs`
+
+Add a new `int SendAttempts` property to `Envelope`. This tracks how many times a send has been attempted for this envelope, independently from handler `Attempts`. Increment it in `SendingAgent` before evaluating failure policies.
+
+### Step 2: Create `SendingFailurePolicy` infrastructure
+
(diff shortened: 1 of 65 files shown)


jeremydmiller:
@vgmello Hey, bad timing on my part, but in preparation to this I tried to get the tests in better shape and got a real CI task in, **but** now can I ask you to rebase this on main again?

Thanks!

jeremydmiller:
@vgmello Is there some reason why you didn't tie this into the existing `QualityOfService` enum instead of adding a second flag? There's already a different setting to declare a KafkaTopic as having AtLeastOnce semantics. What's different about your approach here?

vgmello:
Hi @jeremydmiller, the main rationale was not to change any existing behavior. The QualityOfService.AtLeastOnce, is set based on the EnableAutoCommit being true, however, because of the EnableAutoOffsetStore (default: true) behavior, which stores the message cursor as soon as it gets received by the application, allowing it to be committed automatically on the EnableAutoCommit interval even before it's processed, if the application crashes after the message is received and before its processed, that message wont be retried again.

To avoid this scenario, I'm explicitly setting the EnableAutoOffsetStore to false, which prevents storing of the cursor upon receiving the message, enabling the app to control when the message is fully ingested (either saved in the DB or once finished its inline processing), allowing it to be retried on an application crash.

I did consider tying the AtLeastOnce to the EnableAutoCommit = true and EnableAutoOffsetStore = false, however, if anyone already set the EnableAutoOffsetStore false, they are also controlling that offset storing somewhere else, doing it in the CompleteAsync again could cause unexpected behaviors for those users.

Please let me know if that makes sense, or if there is a simpler way to achive this behavior.  Tks ! Great framework btw.

Ref: https://docs.confluent.io/platform/current/clients/confluent-kafka-dotnet/_site/api/Confluent.Kafka.ConsumerConfig.html#Confluent_Kafka_ConsumerConfig_EnableAutoOffsetStore

jeremydmiller:
I don't know man, I need to review this one a lot more carefully. From a usability perspective, it obviously doesn't work to have two completely different flavors of "AtLeastOnce" guarantee configuration with slightly different usage.

vgmello:
I changed to be more QualityOfService centric, having a explict/override config enabling a backwards compatible new at least once behavior. 
I could also simplify the PR and only make a small change to the KafkaListener.CompleteAsync method, so it would store the offset manually if the Consumer.EnableAutoOffsetStore is false, then we don't need the extra EnableAtLeastOnceDelivery method or properties, and it would enable a reliable at least once delivery, however, it could also cause unexpected behaviors for users manually calling the store offset in a different part of the flow.

jeremydmiller:
@vgmello Hey, sorry, but I'm going to go in a different direction and just build in the auto commit behavior as part of what the QualityOfService configuration is while trying to improve this across the board. I'll update the docs to make it more clear what's happenening.

Thanks for your contributions here, it still spurred on more work even if it didn't go in as is.

PR #2116, opened 2026-02-05
Init: Rate limiting

Joelgullander:
Need input here, am I on the right track? @jeremydmiller


jeremydmiller:
@Joelgullander Hey, I think this looks solid. The exception handling policy looks right to me too. 

The only ask I have is could you lift a base "compliance tests" class for the IRateLimitStore and stick that in Wolverine.ComplianceTests? That'll help us move on to other DBs

PR #2121, opened 2026-02-05
Remove duplicated code snippet from "Manually adding Envelope Mapping" section

dmytro-pryvedeniuk:
"Manually adding Envelope Mapping" section (see https://wolverinefx.net/guide/durability/efcore/outbox-and-inbox.html#manually-adding-envelope-mapping) contains 2 almost identical code snippets. The only difference is `map.ToTable("items");` vs `map.ToTable("items", "mt_items");` line. 

Technically they are different sources sharing the same region -  `sample_mapping_envelope_storage_to_dbcontext` but from documentation side they look like strange duplication.

To avoid confusion `SampleMappedDbContext` from PersistentTests project is removed. It's not actually used (as well as `CleanDbContext` there). Another `SampleMappedDbContext ` class from EfCoreTests project remains as a valid code snippet.


jeremydmiller:
@dmytro-pryvedeniuk Thank you!

PR #2129, opened 2026-02-07
update vulnerability  nugets

kaanlab:
aws sqs|sns nuget (4.0.0.5) has security vulnerability with low severity. Project with such nuget can't add and use ef migrations (only if add <NoWarn>$(NoWarn);NU1901</NoWarn> to csproj).


jeremydmiller:
@kaanlab I got this independently in a different PR, but only after you raised this and alerted me. Thank you!

PR #2133, opened 2026-02-08
Added summary starting tag in IMessageBus

HPOD00019:
I was exploring project and noticed, that there was no open tag in summary of InvokeForTenantAsync method in IMessageBus, so the description was not displaying while using this method (I've tried it).


jeremydmiller:
@HPOD00019 Pfft, good catch.

PR #2137, opened 2026-02-08
Fix SNS batch publish missing message attributes and misplaced Request allocation

desty2k:
## Summary
  - Call `ISnsEnvelopeMapper.ToAttributes()` for each entry in `OutgoingSnsBatch`, matching the single-message path in `AmazonSnsTopic.SendMessageAsync()`
  - Move `PublishBatchRequest` construction outside the per-envelope try/catch so a single mapping failure doesn't leave `Request` uninitialized

  ## Tests
  - Added unit tests for `OutgoingSnsBatch`:
    - Batch entries include message attributes from the mapper
    - `Request` is always constructed even when all envelopes fail mapping
    - A single mapping failure does not block other entries from being batched


jeremydmiller:
@desty2k Thank you! The aws CI thing is, um, not ready for prime time. I'll be running this locally before any nuget release *just in case*

Issue #2182, opened 2026-02-14
Unable to resolve Wolverine.Http.FluentValidation.IProblemDetailSource


dmytro-pryvedeniuk:
`UseFluentValidationProblemDetailMiddleware` with `ExtensionDiscovery.ManualOnly` causes codegen failure: 
```
JasperFx was unable to resolve a variable of type
 Wolverine.Http.FluentValidation.IProblemDetailSource<MyEndpoint.Request> as part of
the method...
```

**To Reproduce**
Run `dotnet run -- codegen preview` for code below. 

```
using FluentValidation;
using JasperFx;
using JasperFx.CodeGeneration;
using Wolverine;
using Wolverine.FluentValidation;
using Wolverine.Http;
using Wolverine.Http.FluentValidation;

var builder = WebApplication.CreateBuilder(args);

builder.Services.AddWolverine(ExtensionDiscovery.ManualOnly, opts =>
{
  opts.CodeGeneration.TypeLoadMode = TypeLoadMode.Static;
  opts.UseFluentValidation();
});
builder.Services.AddWolverineHttp();

var app = builder.Build();

app.MapWolverineEndpoints(opts =>
  opts.UseFluentValidationProblemDetailMiddleware()
);

await app.RunJasperFxCommands(args);

public static class MyEndpoint
{
  [WolverinePost("/testme1")]
  public static IResult Handle(Request value)
  {
    return TypedResults.Ok(value);
  }

  public record Request(string Title)
  {
    public class Validator : AbstractValidator<Request>
    {
      public Validator()
      {
        RuleFor(x => x.Title)
          .MaximumLength(200);
      }
    }
  }
}

```

```
<Project Sdk="Microsoft.NET.Sdk.Web">

  <PropertyGroup>
    <TargetFramework>net10.0</TargetFramework>
    <Nullable>enable</Nullable>
    <ImplicitUsings>enable</ImplicitUsings>
  </PropertyGroup>

  <ItemGroup>
    <PackageReference Include="WolverineFx.Http.FluentValidation" Version="5.15.0" />
  </ItemGroup>

</Project>

```

**Expected behavior**
No error

**Screenshots**

<img width="1901" height="872" alt="Image" src="https://github.com/user-attachments/assets/fb586392-ac97-4513-9d5d-04418882b548" />

**Additional context**
* Using default `ExtensionDiscovery ` helps. Though in larger projects codegen suggests disabling automatic scanning. 
[...]

PR #2183, opened 2026-02-14
Attempted to resolve issue #2182:

XL1TTE:
- Registration of **FailureAction**<> and **ProblemDetailSource**<> happens inside the **Wolverine.Http.FluentValidationExtension**. However, this is not handled in ManualOnly mode because **ExtensionLoader** is never called.

- We could make **FluentValidationExtension** public to allow manual registration via the **Wolverine::WolverineOptions::Include** method, but it appears that **ApplyExtensions** is only called in Automatic mode, making this useless.

- Another straightforward solution, presented by this commit, is to create an overload of **UseFluentValidation** that accepts an additional **ExtensionDiscovery** parameter. In ManualOnly mode, we would register the necessary services for the user and then fall back to the original implementation.

In this case we'd need to call **UseFluentValidation** with **ExtensionDiscovery.ManualOnly** to rigister **Wolverine.Http.FluentValidation** dependencies

```csharp
builder.Services.AddWolverine(ExtensionDiscovery.ManualOnly, opts =>
{
     opts.CodeGeneration.TypeLoadMode = TypeLoadMode.Static;
     opts.UseFluentValidation(ExtensionDiscovery.ManualOnly);
});
```
This PR could potentially resolve issue #2182

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -75,6 +75,7 @@
         <PackageVersion Include="Swashbuckle.AspNetCore.SwaggerGen" Version="9.0.1" />
         <PackageVersion Include="Swashbuckle.AspNetCore.SwaggerUI" Version="9.0.1" />
         <PackageVersion Include="System.ComponentModel" Version="4.3.0" />
+        <PackageVersion Include="Testcontainers.ServiceBus" Version="4.10.0" />
         <PackageVersion Include="System.Diagnostics.DiagnosticSource" Version="9.0.5" />
         <PackageVersion Include="System.Net.NameResolution" Version="4.3.0" />
         <PackageVersion Include="System.Threading.Tasks.Dataflow" Version="9.0.5" />


dmytro-pryvedeniuk:
~~Is it possible to avoid the parameter and register the services by default? I don't know all the details but as an API  consumer I already express my intention to "use fluent validation". If the problem is that the services can be register in another place we probably can use TryAddSingleton.~~

I was wrong above.  `opts.UseFluentValidation()` means "register validators for regular, not http-specific, handlers". On the other hand using an extra parameter as a discriminator does not look right. See below
 
```
builder.Services.AddWolverine(ExtensionDiscovery.ManualOnly, opts =>
{
  ...
  opts.UseFluentValidation(); // regular validation, from Wolverine.FluentValidation
  opts.UseFluentValidation(ExtensionDiscovery.ManualOnly); // problem-details specific services, from  Wolverine.Http.FluentValidation
  opts.UseFluentValidation(ExtensionDiscovery.Automatic); // the same as the first one
});
```

I would expect the following instead

```
builder.Services.AddWolverine(ExtensionDiscovery.ManualOnly, opts =>
{
  ...
  opts.UseFluentValidation(); // regular validators, from Wolverine.FluentValidation
  opts.UseFluentValidationProblemDetail(); // problem-details specific services, from Wolverine.Http.FluentValidation
});
```

PR #2201, opened 2026-02-16
Avoid running a query when there's nothing to do

PerfectlyNormal:
Was debugging a different problem we've been experiencing and caught an exception when it attempted to reassign incoming envelopes. It got called with 0 envelopes, built an empty SQL command and that threw an exception.

Haven't noticed any issues or logs related to this, so guessing it gets caught and handled somewhere, but feels unnecessary to open a connection when we know it won't work ahead of time.


jeremydmiller:
Huh, wonder how this hasn't come up before.

PR #2225, opened 2026-02-23
Add handler discovery of MediatR shim interfaces

XL1TTE:
Implements IRequestHandler<T> and IRequestHandler<T, TResponse> shim interfaces allowing MediatR-style handlers without MediatR dependency.

New files:
- src/Wolverine/Shims/MediatR/IRequest.cs
- src/Wolverine/Shims/MediatR/IRequestHandler.cs
- src/Wolverine/Shims/MediatR/WolverineOptionsExtensions.cs
- src/Testing/ShimsTests/ShimsTests.csproj
- src/Testing/ShimsTests/MediatR/* (test files)

Modified:
- build/build.cs - Added ShimsTests target
- wolverine.sln - Added ShimsTests project

Key features:
- No runtime reflection - uses TypeQuery discovery like IWolverineHandler
- No MediatR dependency - pure shim interfaces
- Full code generation support


XL1TTE:
The `CustomizeHandlerDiscovery` feature has been utilized, allowing Wolverine to generate its own handlers, which helps eliminate reflection calls at runtime (Now i get it, thanks to Jeremy). It would be cool if users could migrate to Wolverine without having to explicitly change namespaces from MediatR's to Wolverine's. However, I understand it's likely not an option to reference non-MIT libraries from a commercial point of view.

jeremydmiller:
@XL1TTE I'm going to get this in pretty soon, but move things around just a bit

PR #2230, opened 2026-02-23
Add cancel scheduled message support with SchedulingToken and ScheduleResult

agileramblings:
AI-Supported PR that adds cancellation support for scheduled messages. Claude Opus 4.6 was used for a lot of this.

## Summary

Adds first-class support for **cancelling previously scheduled messages** across Wolverine. Before this change, there was no way to retrieve a handle to a scheduled message in order to cancel it before delivery. This PR introduces the `SchedulingToken` concept on `Envelope`, a `ScheduleWithResultAsync` API on `IMessageBus`, a `CancelScheduledAsync` API on `IDestinationEndpoint`, and an `ISenderWithScheduledCancellation` opt-in interface for transports.

Full cancellation is implemented for **Azure Service Bus** (using its native sequence number) and all **single-tenant database-backed queues** (PostgreSQL, SQL Server, SQLite, MySQL, Oracle) using a `DELETE` against the scheduled table. Multi-tenanted DB queue configurations do not support cancellation in this PR.

---

### `Envelope.SchedulingToken`

A new `object?` property on `Envelope`. When a message is natively scheduled, the transport's sender writes a transport-specific token back to this property after the send completes. The caller captures it from `ScheduleResult` to use later for cancellation.

- **Azure Service Bus**: `long` — the ASB sequence number returned by `ServiceBusSender.ScheduleMessageAsync`
- **DB-backed queues** (single-tenant): `Guid` — the envelope's own `Id`

### `MessageBusExtensions.ScheduleWithResultAsync<T>()`

Two new overloads — one by `DateTimeOffset`, one by `TimeSpan` — that schedule a message and return a `ScheduleResult` containing the routed envelopes with their `SchedulingToken` already populated by the transport.

```csharp
// By absolute time
ScheduleResult result = await bus.ScheduleWithResultAsync(message, DateTimeOffset.UtcNow.AddHours(1));

// By delay
ScheduleResult result = await bus.ScheduleWithResultAsync(message, TimeSpan.FromMinutes(30));
```

> **Note:** This method is incompatible with an active outbox transaction. An `InvalidOperationException` is thrown if called while an outbox is open, because the message is queued in the outbox rather than sent immediately and no transport token can be returned.

### `ScheduleResult`

A new class wrapping the `IReadOnlyList<Envelope>` produced by a `ScheduleWithResultAsync` call.

```csharp
var token = result.Envelopes[0].SchedulingToken;
// Store token alongside the entity that owns the scheduled action
```

### `IDestinationEndpoint.CancelScheduledAsync(object schedulingToken)`

Cancels a previously scheduled message using the token from `ScheduleResult`. Throws `NotSupportedException` by default; transports opt in by implementing `ISenderWithScheduledCancellation`.

```csharp
await bus.EndpointFor("my-queue").CancelScheduledAsync(token);
```

---

### `ISender.SupportsNativeScheduledCancellation`

A new `bool` property added to `ISender`. Returns `false` on all senders except those that implement cancellation (see transport matrix below).

### `ISenderWithScheduledCancellation`

A new opt-in interface extending `ISender`. Transports implement this to provide `CancelScheduledMessageAsync(object schedulingToken, CancellationToken)`.

```csharp
public interface ISenderWithScheduledCancellation : ISender
{
    Task CancelScheduledMessageAsync(object schedulingToken, CancellationToken cancellation = default);
}
```

### `TenantedSender`

Now implements `ISenderWithScheduledCancellation`, delegating to the per-tenant or default inner sender. Also exposes `SenderForTenantId()` so `DestinationEndpoint` can resolve the correct sender in multi-tenant scenarios.

### `DestinationEndpoint.CancelScheduledAsync`

Resolves the cancellation-capable sender through the following priority chain:

1. Sending agent directly (e.g., `StubEndpoint`)
2. `InlineSendingAgent.Sender`
3. Unwrapped from `TenantedSender` using the current tenant ID

---

## Transport Support Matrix

| Transport | Cancellation | Token Type | Mechanism |
|---|---|---|---|
| **Azure Service Bus** | ✅ Yes | `long` | `ServiceBusSender.CancelScheduledMessageAsync(sequenceNumber)` |
| **PostgreSQL queue** (single-tenant) | ✅ Yes | `Guid` | `DELETE FROM <scheduled_table> WHERE id = :id` |
| **SQL Server queue** (single-tenant) | ✅ Yes | `Guid` | `DELETE FROM <scheduled_table> WHERE id = @id` |
| **SQLite queue** | ✅ Yes | `Guid` | `DELETE FROM <scheduled_table> WHERE id = @id` |
| **MySQL queue** (single-tenant) | ✅ Yes | `Guid` | `DELETE FROM <scheduled_table> WHERE id = @id` |
| **Oracle queue** (single-tenant) | ✅ Yes | `Guid` | `DELETE FROM <scheduled_table> WHERE id = :id` |
| **PostgreSQL queue** (multi-tenant) | ❌ No | — | — |
| **SQL Server queue** (multi-tenant) | ❌ No | — | — |
| **MySQL queue** (multi-tenant) | ❌ No | — | — |
| **Oracle queue** (multi-tenant) | ❌ No | — | — |
| AWS SQS / SNS | ❌ No | — | — |
| GCP Pub/Sub | ❌ No | — | — |
| Kafka | ❌ No | — | — |
| Redis Streams | ❌ No | — | — |
| NATS | ❌ No | — | — |
| Pulsar | ❌ No | — | — |
| RabbitMQ | ❌ No | — | — |
| MQTT | ❌ No | — | — |
| SignalR | ❌ No | — | — |
| HTTP | ❌ No | — | — |

---

## Usage Example

```csharp
// Schedule a message and capture the scheduling token
var result = await bus.ScheduleWithResultAsync(
    new SendInvoiceReminderMessage(invoiceId),
    TimeSpan.FromHours(24));

var token = result.Envelopes[0].SchedulingToken;
// Persist `token` in your database alongside the invoice record

// Later, if the invoice is paid before the reminder fires — cancel it
await bus.EndpointFor("invoice-reminders").CancelScheduledAsync(token);
```

---

## Tests

| File | Coverage |
|---|---|
| `SchedulingTokenTests.cs` | `Envelope.SchedulingToken` is null by default; accepts `long` and `Guid`; `ScheduleResult` construction; `NullSender` is not `ISenderWithScheduledCancellation` |
| `SchedulingCapabilityTests.cs` | `TenantedSender` propagates the cancellation flag from its default sender; `SenderForTenantId` resolution; `CancelScheduledMessageAsync` delegates to the inner cancellable sender; throws `NotSupportedException` when inner sender does not support cancellation |
| `InlineAzureServiceBusSenderTests.cs` | Unit tests: `SendAsync` writes the ASB sequence number back to `envelope.SchedulingToken` on scheduled sends; `CancelScheduledMessageAsync` calls the ASB SDK; validates token type |
| `schedule_and_cancel_with_sequence_number.cs` | Integration tests (requires ASB emulator: `docker compose up -d asb-sql asb-emulator`): `ScheduleWithResultAsync` returns a non-null `long` sequence number greater than zero; `CancelScheduledAsync` completes without error |

---

## Notes

- **Outbox incompatibility**: `ScheduleWithResultAsync` requires a direct send to the transport. It throws `InvalidOperationException` when called inside an active outbox transaction, since the message is queued in the outbox rather than sent immediately and no transport token can be returned.
- **Unsupported transports fail loudly**: Calling `CancelScheduledAsync` on an endpoint backed by a transport that does not implement `ISenderWithScheduledCancellation` throws `NotSupportedException` rather than silently doing nothing.
- **Multi-tenanted DB queues**: The `MultiTenantedQueueSender` variants for PostgreSQL, SQL Server, MySQL, and Oracle return `SupportsNativeScheduledCancellation = false` in this PR. Cancellation for multi-tenanted DB queues is a follow-up.
- **ASB integration tests are local-only**: There is no GitHub Actions workflow for ASB tests. They require `docker compose up -d asb-sql asb-emulator` and are run locally only.
- **All other `ISender` implementors**: Every transport sender that does not support cancellation was updated to return `SupportsNativeScheduledCancellation = false`, satisfying the new interface property without changing any existing behavior.

### DIFF
--- a/src/Http/Wolverine.Http/Transport/InlineHttpSender.cs
+++ b/src/Http/Wolverine.Http/Transport/InlineHttpSender.cs
@@ -28,6 +28,7 @@ internal class InlineHttpSender(HttpEndpoint endpoint, IWolverineRuntime runtime
     }
 
     public bool SupportsNativeScheduledSend => endpoint.SupportsNativeScheduledSend;
+    public bool SupportsNativeScheduledCancellation => false;
     public Uri Destination => endpoint.Uri;
     public Task<bool> PingAsync() => Task.FromResult(true);
 
--- a/src/Persistence/MySql/Wolverine.MySql/Transport/IMySqlQueueSender.cs
+++ b/src/Persistence/MySql/Wolverine.MySql/Transport/IMySqlQueueSender.cs
@@ -2,7 +2,7 @@ using Wolverine.Transports.Sending;
 
 namespace Wolverine.MySql.Transport;
 
-internal interface IMySqlQueueSender : ISender
+internal interface IMySqlQueueSender : ISenderWithScheduledCancellation
 {
     Task ScheduleRetryAsync(Envelope envelope, CancellationToken cancellationToken);
 }
--- a/src/Persistence/MySql/Wolverine.MySql/Transport/MultiTenantedQueueSender.cs
+++ b/src/Persistence/MySql/Wolverine.MySql/Transport/MultiTenantedQueueSender.cs
@@ -25,6 +25,7 @@ internal class MultiTenantedQueueSender : IMySqlQueueSender, IAsyncDisposable
     }
 
     public bool SupportsNativeScheduledSend => true;
+    public bool SupportsNativeScheduledCancellation => false;
     public Uri Destination { get; }
 
     public Task<bool> PingAsync()
@@ -86,6 +87,13 @@ internal class MultiTenantedQueueSender : IMySqlQueueSender, IAsyncDisposable
         }
 
         return sender;
+    }
+
+    public Task CancelScheduledMessageAsync(object schedulingToken, CancellationToken cancellation = default)
+    {
+        throw new NotSupportedException(
+            "Cancelling scheduled messages is not supported for multi-tenanted database queue endpoints. " +
+            "Use a single-tenant endpoint or wait for future multi-tenant cancellation support.");
     }
 
     public ValueTask DisposeAsync()
--- a/src/Persistence/MySql/Wolverine.MySql/Transport/MySqlQueueSender.cs
+++ b/src/Persistence/MySql/Wolverine.MySql/Transport/MySqlQueueSender.cs
@@ -59,6 +59,7 @@ ON DUPLICATE KEY UPDATE {DatabaseConstants.Body} = @body, {DatabaseConstants.Mes
     }
 
     public bool SupportsNativeScheduledSend => true;
+    public bool SupportsNativeScheduledCancellation => true;
     public Uri Destination { get; private set; }
 
     public async Task<bool> PingAsync()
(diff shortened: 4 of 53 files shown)


agileramblings:
Initial review found this far too expansive in scope and included a support landmine that I hadn't fully realized. I'll resubmit a far more limited-scope PR, with more attention to limiting the blast radius of changes and support problems.

PR #2240, opened 2026-02-25
Fix solution file virtual folders for EfCoreTests.MultiTenancy

XL1TTE:
- Solution.src.Persistence.EfCoreTests_MultiTenancy -> Solution.Persistence.EfCoreTests_MultiTenancy


XL1TTE:
### Before:


### After:

jeremydmiller:
Meh, thanks. I've missed that a few times

PR #2242, opened 2026-02-25
Guard pause listener continuation and shutdown

Joelgullander:
## Summary
- Add unit coverage for null envelope/listener in PauseListenerContinuation.
- Update RabbitMQ rate-limit E2E to assert the current bug

Suggestion / attempt:
- Guard PauseListenerContinuation and harden listener shutdown against null/disposing races.

## Test plan
- dotnet test src/Testing/CoreTests/CoreTests.csproj --filter PauseListenerContinuationTests
- dotnet test src/Transports/RabbitMQ/Wolverine.RabbitMQ.Tests/Wolverine.RabbitMQ.Tests.csproj --filter rate_limited_messages_do_not_throw_when_rescheduled


jeremydmiller:
@Joelgullander I got this with a local rebase + push. The "fix" is in too. Turned out to be a timing issue.

Issue #2258, opened 2026-03-03
NRE in DurableReceiver when Listener is null during redelivery


Joelgullander:
## Summary
When a DurableReceiver is latched and processes an envelope loaded from the durable inbox, `Envelope.Listener` can be null. The retry/defer block calls `env.Listener!.DeferAsync(...)` and throws a `NullReferenceException`.

## Stack trace
```
System.NullReferenceException: Object reference not set to an instance of an object.
   at Wolverine.Runtime.WorkerQueues.DurableReceiver.<>c.<.ctor>b__16_1(Envelope env, CancellationToken _) in .../DurableReceiver.cs:line 85
   at JasperFx.Blocks.LambdaItemHandler`1.ExecuteAsync(...)
   at JasperFx.Blocks.RetryBlock`1.PostAsync(...)
```

## Reproduction (tests)
- **Unit**: `CoreTests.Runtime.WorkerQueues.durable_receiver_latched_without_listener`
  - Latch receiver, call `ReceivedAsync` with an envelope whose `Listener` is null → NRE
- **Integration (durable inbox)**: `SlowTests.durable_receiver_redelivery_from_inbox`
  - Persist envelope in inbox (Postgres), load it, latch receiver, call `ReceivedAsync` → NRE is logged

## Expected
DurableReceiver should not throw when `Listener` is missing on a redelivered inbox envelope.

## Actual
NRE from `env.Listener!.DeferAsync(env)` in `_deferBlock`.

## Test plan
- dotnet test src/Testing/CoreTests/CoreTests.csproj --filter durable_receiver_latched_without_listener
- dotnet test src/Testing/SlowTests/SlowTests.csproj --filter durable_receiver_redelivery_from_inbox

PR #2259, opened 2026-03-03
Add tests for durable receiver listener null

Joelgullander:
## Summary
- Add a unit test to reproduce the latched durable receiver NRE when `Envelope.Listener` is null.
- Add a slow integration test that loads an envelope from the durable inbox (Postgres) and reproduces the NRE via the redelivery path.
- Wire `Wolverine.Postgresql` into SlowTests for the integration repro.

## Test plan
- dotnet test src/Testing/CoreTests/CoreTests.csproj --filter durable_receiver_latched_without_listener
- dotnet test src/Testing/SlowTests/SlowTests.csproj --filter durable_receiver_redelivery_from_inbox

Related to #2258


jeremydmiller:
I pulled this in directly in another branch

PR #2268, opened 2026-03-05
Switch to slnx

dmytro-pryvedeniuk:
This PR is about using clean slnx instead of sln.

I have to remove build.csproj from the solution. The reason is that support for slnx was added to Nuke in version 10.0.0 (according to https://github.com/nuke-build/nuke/blob/develop/CHANGELOG.md#:~:text=Added%20support%20for%20slnx%20solution%20files). 

This version requires build.csproj to target .net10. If build.csproj targeting .net10 is a part of the solution restore step fails with the default .net9.

I also disabled Central Package Management for build.csproj to make it independent from the main solution (Npgsql and Nuke.Common are used only by build.csproj).

The following warning is fixed implicitly.
image


jeremydmiller:
Hey man, there's too many branches floating around right now to try this. This is the kind of thing you gotta just suggest. Love getting pull requests from the community, but generally not things like this though.


Sorry, I'm not going to take this one.

dmytro-pryvedeniuk:
@jeremydmiller No problem—this PR was more of a suggestion. Initially, I just wanted to fix the warning from the screenshot above, but switching to slnx ended up looking more promising.

If your concern is many active branches atm maybe we can revisit it later.

BTW, in this article (https://devblogs.microsoft.com/visualstudio/new-simpler-solution-file-format/) they state: _the recommended best practice is to coordinate the transition with all clients and move fully to .SLNX once the entire team or organization is ready._

PR #2269, opened 2026-03-06
Fix flaky redis tests

dmytro-pryvedeniuk:
Two flaky tests fixed, simplified and hardened.

1. `message_with_retry_policy_saves_to_redis_and_retries` uses hard-coded delays which sometimes are not enough. Also, instead of strict asserts it relies on if conditions with logging instead of assertion. E.g. it's assumed that the default EndpointMode is Durable while it's not anymore and instead of asserting the test just logs that it can be the problem. So when the test sometimes passes it's actually just due to such relaxed validation.
2. `rate_limited_messages_are_delayed_with_native_scheduling` waits 15 seconds for 2 messages handled and then throws an exception. I have to enable DurableInbox (sometimes one message does not arrive) and make the test ready for the messages arriving in different order (see output below).

For both tests instead of one hard-coded delay more granular delays are used with the log output.

Message B arrives first:
```
3/6/2026 8:05:05 PM Sending message: RedisRateLimitedMessage { Id = A }
3/6/2026 8:05:05 PM Sending message: RedisRateLimitedMessage { Id = B }
3/6/2026 8:05:05 PM Waiting for condition: handled at least 2 messages (total 1000ms)
3/6/2026 8:05:06 PM Waiting for condition: handled at least 2 messages (total 2000ms)
3/6/2026 8:05:07 PM Waiting for condition: handled at least 2 messages (total 3000ms)
3/6/2026 8:05:07 PM Handled message RedisRateLimitedMessage { Id = B }
3/6/2026 8:05:08 PM Waiting for condition: handled at least 2 messages (total 4000ms)
3/6/2026 8:05:09 PM Waiting for condition: handled at least 2 messages (total 5000ms)
3/6/2026 8:05:10 PM Waiting for condition: handled at least 2 messages (total 6000ms)
3/6/2026 8:05:11 PM Waiting for condition: handled at least 2 messages (total 7000ms)
3/6/2026 8:05:12 PM Waiting for condition: handled at least 2 messages (total 8000ms)
3/6/2026 8:05:13 PM Waiting for condition: handled at least 2 messages (total 9000ms)
3/6/2026 8:05:14 PM Waiting for condition: handled at least 2 messages (total 10000ms)
3/6/2026 8:05:15 PM Waiting for condition: handled at least 2 messages (total 11000ms)
3/6/2026 8:05:15 PM Handled message RedisRateLimitedMessage { Id = A }
```

### DIFF
--- a/src/Transports/Redis/Wolverine.Redis.Tests/rate_limiting_end_to_end.cs
+++ b/src/Transports/Redis/Wolverine.Redis.Tests/rate_limiting_end_to_end.cs
@@ -1,5 +1,4 @@
 using JasperFx.Core;
-using JasperFx.Events.Daemon;
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.Hosting;
 using Shouldly;


jeremydmiller:
@dmytro-pryvedeniuk Thank you for doing that!

PR #2287, opened 2026-03-11
Fix stale messages during rollout and improve graceful shutdown

### LINKED ISSUE #2279: Wolverine stale messages during rollout (Teardown Lifecycle Race Condition)
### **Describe the bug** 
During a container rollout (e.g., Kubernetes, Docker, Cloud), scheduled or incoming messages can become permanently stuck in the Inbox/Outbox with an owner_id pointing to a dead node.

Because the dead node's record is successfully deleted from the wolverine_nodes table during the teardown, the active Leader's fast recovery loop (RecoverDormantNodes) is completely blind to these orphaned messages. They remain stuck indefinitely unless the heavy, globally disabled `InboxStaleTime` / `OutboxStaleTime` fallback is manually configured.

**The Root Cause**: An Asynchronous Teardown Race Condition By inspecting the source code, we confirmed that `NodeAgentController#_persistence.DeleteAsync()` unregisters the node and unlocks its messages as a single atomic database batch. Therefore, the ghost messages are not caused by a mid-query crash, but by an asynchronous lifecycle ordering flaw during graceful shutdown. Agents are returning from StopAsync before their background tasks have fully drained.

[Mermaid chart link](https://mermaid.ai/play#pako:eNptk9uO2jAQhl9lykUFagu3FaqQAoEtWpaqQMXNSpWxh2DhjFMfdhtVfffaSSBhaS4sJZl__m8O_tPjWmBv3LP4yyNxTCXLDMufCcJTMOMklwUjB4-fLTALj_6AhtChhRHMlPYCNp7uw78ZfkLrDHPaRN062CQZkptpckYrheZeVAXE6NQbdpBKurL-NIKVtA4Jjb1XpdNKwhw7MIvPDUzg_TSZfOhyjGG7fNjNN0_QfzCM49Er2J68E_qVBrWqPruimKSCGIN1ukiUqt5sYkvi_RvZWjsE_YKmLuQj3Jrvvs5hsUr246ZOzpSyMNN5oUI_-wOQFGpkAvTxy8GMJnsm3UKbJkBqaiyHsHRg0HlDFmSeo5DMoSrf1RRV8kDdQP80nkhSNlWan4etWzc4RL9BZbaNFfGtW2do_u18ZwoZ-aL--58GptMxpBhzNSVAP3E6lxz2ssBBq6s6qPDoQg8gquLagNGvICq5gPdwpjAuyNFaloUt9BQLQ1EnYSTCqlCGJhwwZfycGe3Dx70259bn0qNoMf-N3Afb4MDK4LD9voJwF0z5BsvI7HTlqicYncOtgPV8fwECp8Gd8MpbFbBM762b8XQQY5vhKCmsRQm8ab5tlQIZd_IlzLpOUv9BEpeVb_cvIi6SXbKCzXz7Y7Ubw1ODJy2E_kGBJmcUkqiy2rW6hxGeXdEpoA_rzB3r7mh7f_8Bc5VsYw)

### **The Exact Sequence of Failure:**
The orchestrator sends a `SIGTERM` and the pod begins graceful shutdown.

`NodeAgentController` awaits on `stopAllAgentsAsync()`, which calls `StopAsync()` on its agents (like DurabilityAgent).

**The Early Return Race Condition:** DurabilityAgent calls _runningBlock.Complete(). Looking at `JasperFx.Blocks.Block, Complete()` is synchronous—it flips a latch and returns instantly. It does not await the currently executing background tasks.

`NodeAgentController` assumes the agents are fully stopped and executes the atomic `_persistence.DeleteAsync()`. The database is now perfectly clean.

**The Locked Message after database cleanup**: A lingering background thread from the `DurabilityAgent` (or a transport listener) is still executing (stemming from either of the Race Conditions above). Milliseconds after the atomic wipe, it executes its SQL query, locking a message by setting `owner_id = [DeletedNodeId]`.

The orchestrator's grace period expires, sending `SIGKILL`. The newly locked message is left behind as a permanent ghost.

**Why a run-once recovery on startup is insufficient:** It might seem logical to patch this by having newly started nodes run a one-time "Orphaned Lock Sweep" on startup (e.g., unlocking messages where owner_id is not in wolverine_nodes). However, this fundamentally fails due to how standard orchestrators handle Rolling Updates (e.g., Kubernetes maxSurge deployments):

To ensure zero downtime, the orchestrator **spins up the new pods first**.

The new pods boot, run their one-time startup sweeps, and find nothing to recover (because the old pods are still alive and actively processing).

Once the new pods report as healthy, the orchestrator begins sequentially terminating the old pods.

The old pods fall victim to the teardown race condition and leave ghost messages behind.

Because all new pods have already booted and completed their startup sweeps, there is no node left to wake up and sweep the messages orphaned by the dying nodes.

Because cloud-native rollouts create new nodes before stopping old ones, the framework cannot rely on startup sweeps to clean up the mess. It must enforce strict "happens-before" teardown semantics to prevent ghost messages from being created in the first place.

### **To Reproduce**

Start a Wolverine instance using a relational database (Postgres/SQL Server) for durability.

Place a thread breakpoint in a background task (for example, `DurabilityAgent#RunScheduledMessagesOperation`) right before it locks a message in the database. This simulates a hanging/delayed background task.

Send a `SIGTERM` to the process.

Notice that the node's row in `wolverine_nodes` is wiped clean before the background task paused at the breakpoint finishes its execution.

Release the breakpoint, allowing the background task to lock the message.

Send a `SIGKILL` to the process (or let it die naturally).

Error: The message is now locked by a ghost node.

### **Expected behavior** 

A node orchestrating a graceful shutdown should not picking up messages and leave it as stale message.

**OS / Hosting**:  Linux / GCP Cloud Run

**.NET Version:** .NET 10

**Wolverine Version**: 5.16.1

**Database Engine**: PostgreSQL

### **Additional context** 

We will not be using `InboxStaleTime` and `OutboxStaleTime` as a fallback, it could be too long or causing unwanted replay of messages.

### LINKED ISSUE #2282: Wolverine instance graceful shutdown

When using container orchestration (like k8s, cloud-run, aspire, etc) during version roll-out deployment, new instances are spun-out while the old ones are gracefully shut down.

When handling HTTP requests, traffic is routed to the new instances while inflight requests have grace period to complete until they are forcefully interrupted.

We would like to know how wolverine is supposed to handle shutdown request. The only piece of [doc](https://wolverinefx.net/introduction/best-practices.html#graceful-shutdown-of-nodes) we found is lacking the intended process followed by a sunsetting wolverine instances, especially regarding in flight and long running messages. Also, are `CancellationToken`s passed to handlers leveraged or not ? We have not found any occurrences of `IHostApplicationLifetime.ApplicationStopping` in the wolverine code base.

Documentation of what is supposed to happen, guarantees wolverine provides, and/or the steps an implementer should take to achieve a graceful shutdown.

For the process itself, we expect the graceful shutdown process to be something as the following:

* On receiving SIGTERM, stop pulling messages from the transport inbox
* Let the inflight handlers complete
* After a configurable timeout, abort all remaining handlers
* Release wolverine node messages assignation (set it to zero)
* Flush the outbox to the transport
* Dies

This should guarantees smooth instance sunsetting, and no messages lost nor stale.

None, N/A.

**Additional context**

The ASP.Net documentation on [graceful shutdown](https://github.com/dotnet/dotnet-docker/blob/main/samples/kubernetes/graceful-shutdown/graceful-shutdown.md) has lot of good information and examples.

jeremydmiller:
### Orphaned message recovery (#2279)
- Adds periodic recovery of messages stranded by dead nodes during rolling deployments
- `ReleaseOrphanedMessagesOperation` (main databases): uses `NOT IN (SELECT node_number FROM wolverine_nodes)` subquery
- `ReleaseOrphanedMessagesForAncillaryOperation` (ancillary/tenant databases): fetches active node numbers from the main store, builds explicit `NOT IN (1, 2, 3)` list
- Both operations run every recovery cycle in `DurabilityAgent.buildOperationBatch()`

### Graceful shutdown improvements (#2282)
- Reorder shutdown sequence: drain endpoints first (complete in-flight handlers), then release ownership and teardown agents
- Add bounded `WaitForCompletionAsync` in `DurableReceiver.DrainAsync()` and `BufferedReceiver.DrainAsync()` so in-flight message handlers finish before ownership is released
- Add configurable `DrainTimeout` setting (default 30 seconds) to `DurabilitySettings`
- New `GracefulShutdown` and `RollingRestart` chaos test scripts exercising shutdown scenarios

## Test plan
- [x] CoreTests: 1160 passed, 0 failed
- [x] PostgreSQL tests: 331 passed, 4 failed (pre-existing compliance test flakiness)
- [x] SQL Server tests: 294 passed, 1 failed (pre-existing tracking/timing issue)
- [x] Added GracefulShutdown and RollingRestart chaos test scripts with 6 new test methods

Closes #2279
Closes #2282

### DIFF
--- a/src/Persistence/Wolverine.RDBMS/Durability/ReleaseOrphanedMessagesForAncillaryOperation.cs
+++ b/src/Persistence/Wolverine.RDBMS/Durability/ReleaseOrphanedMessagesForAncillaryOperation.cs
@@ -1,0 +1,52 @@
+using System.Data.Common;
+using JasperFx.Core;
+using Weasel.Core;
+using Wolverine.RDBMS.Polling;
+using Wolverine.Runtime.Agents;
+using DbCommandBuilder = Weasel.Core.DbCommandBuilder;
+
+namespace Wolverine.RDBMS.Durability;
+
+/// <summary>
+/// Releases inbox/outbox messages owned by nodes that no longer exist.
+/// Used for ancillary/tenant databases that lack the wolverine_nodes table,
+/// so the active node list must be provided externally.
+/// </summary>
+internal class ReleaseOrphanedMessagesForAncillaryOperation : IDatabaseOperation, IDoNotReturnData
+{
+    private readonly IMessageDatabase _database;
+    private readonly IReadOnlyList<int> _activeNodeNumbers;
+
+    public ReleaseOrphanedMessagesForAncillaryOperation(IMessageDatabase database, IReadOnlyList<int> activeNodeNumbers)
+    {
+        _database = database;
+        _activeNodeNumbers = activeNodeNumbers;
+    }
+
+    public string Description => "Release inbox/outbox messages owned by nodes that no longer exist (ancillary database)";
+
+    public void ConfigureCommand(DbCommandBuilder builder)
+    {
+        if (_activeNodeNumbers.Count == 0) return;
+
+        var schemaName = _database.SchemaName;
+        var incomingTable = new DbObjectName(schemaName, DatabaseConstants.IncomingTable);
+        var outgoingTable = new DbObjectName(schemaName, DatabaseConstants.OutgoingTable);
+        var nodeList = string.Join(", ", _activeNodeNumbers);
+
+        builder.Append(
+            $"update {incomingTable} set {DatabaseConstants.OwnerId} = 0 where {DatabaseConstants.OwnerId} != 0 and {DatabaseConstants.OwnerId} not in ({nodeList});");
+        builder.Append(
+            $"update {outgoingTable} set {DatabaseConstants.OwnerId} = 0 where {DatabaseConstants.OwnerId} != 0 and {DatabaseConstants.OwnerId} not in ({nodeList});");
+    }
+
+    public Task ReadResultsAsync(DbDataReader reader, IList<Exception> exceptions, CancellationToken token)
+    {
+        return Task.CompletedTask;
+    }
+
+    public IEnumerable<IAgentCommand> PostProcessingCommands()
+    {
+        yield break;
+    }
+}
--- a/src/Persistence/Wolverine.RDBMS/Durability/ReleaseOrphanedMessagesOperation.cs
+++ b/src/Persistence/Wolverine.RDBMS/Durability/ReleaseOrphanedMessagesOperation.cs
@@ -1,0 +1,46 @@
+using System.Data.Common;
+using Weasel.Core;
+using Wolverine.RDBMS.Polling;
+using Wolverine.Runtime.Agents;
+using DbCommandBuilder = Weasel.Core.DbCommandBuilder;
+
+namespace Wolverine.RDBMS.Durability;
+
+/// <summary>
+/// Releases inbox/outbox messages owned by nodes that no longer exist in the wolverine_nodes table.
+/// Used for main databases where the nodes table is co-located.
+/// </summary>
+internal class ReleaseOrphanedMessagesOperation : IDatabaseOperation, IDoNotReturnData
+{
+    private readonly IMessageDatabase _database;
+
+    public ReleaseOrphanedMessagesOperation(IMessageDatabase database)
+    {
+        _database = database;
+    }
+
+    public string Description => "Release inbox/outbox messages owned by nodes that no longer exist";
+
+    public void ConfigureCommand(DbCommandBuilder builder)
+    {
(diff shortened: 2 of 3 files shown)


linxuhao:
Weird, I don't see changes related to Graceful shutdown improvements (https://github.com/JasperFx/wolverine/issues/2282) about WaitForCompletionAsync

PR #2306, opened 2026-03-16
Fix null message_type causing NOT NULL violation when dead-lettering …

lahma:
…CloudEvents with unregistered types

When a Kafka CloudEvents message arrives with an unregistered `type`, CloudEventsMapper throws UnknownMessageTypeNameException before setting envelope.MessageType. The subsequent dead-letter INSERT hits a NOT NULL violation on message_type, causing an infinite retry loop.

Layered fix:
- CloudEventsMapper: preserve raw CloudEvent type on envelope before attempting resolution, so the original type identity survives for dead-letter persistence
- MoveToErrorQueue: add fallback guard that assigns a sentinel message type (unknown/{ExceptionName}) when Message is null and MessageType was never set
- DatabasePersistence: null-coalesce MessageType to "unknown" in the dead-letter INSERT as a belt-and-suspenders defense

Triggered by #2308 .

### DIFF
--- a/src/Persistence/Wolverine.RDBMS/DatabasePersistence.cs
+++ b/src/Persistence/Wolverine.RDBMS/DatabasePersistence.cs
@@ -162,7 +162,7 @@ public static class DatabasePersistence
             builder.AddParameter(envelope.Id),
             builder.AddParameter(envelope.ScheduledTime),
             builder.AddParameter(data),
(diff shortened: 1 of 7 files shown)


lahma:
Main seems to have the failing tests.

jeremydmiller:
Yuck, yeah, I'll get this in now. Thank you!

PR #2218, opened 2026-02-20
Fix flaky Wolverine.Http.Tests

dmytro-pryvedeniuk:
This PR fixes an issue with Wolverine.Http tests. Note that the issue is not reproducible when the tests run on GitHub.

On my machine (Windows 11) Wolverine.Http tests **always** fail, both when running via cmd or from VS. There are different tests and failures: NullReferenceException, ObjectDisposedException, unexpected NotFound errors in AlbaHost.Scenario. See the screenshots below.

To reproduce it's enough to run 2 tests in VS, e.g. `multi_tenanted_session_factory_without_wolverine.can_do_the_tenancy_detection` and `soft_deleted_attribute_usage.try_it_out`.
If they run separately they succeed as expected even when running multiple times, but when they run together `try_it_out` fails with `NullReferenceException`. 

Reasons:
1. When a test from "integration" collection (e.g. `soft_deleted_attribute_usage.try`) runs together with a test that creates its own host (e.g. `multi_tenanted_session_factory_without_wolverine.can_do_the_tenancy_detection`) the latter manages to set `WolverineOptions.RememberedApplicationAssembly` BEFORE Host creation in `AppFixture.InitializeAsync`. As the result the endpoints from `WolverineWebApi` can not be discovered in "integration" collection tests.
2. `IntegrationContext` class in `Scenario` method tries to reset host when 404 status code is received. It's not nice as 404 can be a valid response. Also, it does not help and causes `NullReferenceException `for `soft_deleted_attribute_usage.try_it_out`.

As a possible solution I set `WolverineOptions.RememberedApplicationAssembly` in `IntegrationContext `and remove dangerous ResetHost. 
Also, `can_not_compile_with_manual_discovery_by_default` test is "fixed" to make the outcome the same when running on GitHub (the scenario returns the expected error as 500 response) and locally (the scenario throws the expected exception directly). Have not managed to find the reason of such different behavior. 

Failure1

Failure2

Failure3

Failure4

Failure5

PR #2321, opened 2026-03-18
Fix windows event log disposed in tests

dmytro-pryvedeniuk:
This PR updates to the latest Alba version to apply https://github.com/JasperFx/alba/pull/211 for Wolverine tests.

This is also continuation of https://github.com/JasperFx/wolverine/pull/2218 where as a workaround Wolverine runtime was stopped manually. While it helped for SOME tests (ones based on IntegrationContext) other tests remained flaky.

With the fix from Alba:
- workaround is not needed for IntegrationContext 
- Windows Event Log is disabled thus it can't cause exceptions anymore (includes `Microsoft.Identity.Web.Logger` singleton)
- StopAsync is called for Wolverine runtime as expected regardless of how `IHost` is constructed.

### DIFF
--- a/src/Http/Wolverine.Http.Tests/Bugs/Bug_using_host_stop.cs
+++ b/src/Http/Wolverine.Http.Tests/Bugs/Bug_using_host_stop.cs
@@ -1,7 +1,11 @@
 ﻿using Alba;
+using JasperFx.CommandLine;
+using JasperFx.Core.Reflection;
 using Microsoft.AspNetCore.Builder;
+using Microsoft.Extensions.Hosting;
 using Shouldly;
 using System.Reflection;
+using Microsoft.Extensions.DependencyInjection;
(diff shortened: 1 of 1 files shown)


dmytro-pryvedeniuk:
There are failing tests, but I am shamelessly blaming the main branch.

(Checked `efcore` test, but could not find the right combination to fail `Bug_1846_duplicate_execution_of_scheduled_jobs.should_not_double_execute`. It's flaky but looks like just in some combination with others)

@jeremydmiller I do believe something should be done with the failing/flaky/cancelled tests in main. New features are cool, but if we don't have stable main it's just an open door for problems.

dmytro-pryvedeniuk:
@jeremydmiller My proposals for "something should be done" - https://github.com/JasperFx/wolverine/issues/2329

PR #2322, opened 2026-03-18
feat: Improve graceful shutdown for inline processor, rmq, and sqs

benjamin-alexander-simplisafe:
Hi there. We've recently started using Wolverine for a queue processor that, for various reasons, we need to use with inline mode. This fixes a couple issues we've noticed with it. Admittedly some of the understanding of how this works was Claude assisted so i hope nothing's too off base.

Adds latching and draining for the inline receiver, waiting for already started messages to complete processing and deferring any that haven't. Also added tests for this behavior. This avoids unnecessary re-processing and delays (especially when used with aws sqs).

The following I don't seen any tests covering the existing behavior, so I held off on adding any for now. Let me know if you think it makes sense to add some.

- Updates listening agent to latch `InlineReceiver`s. Also fixes a bug (i think) where a ReceiverWithRules won't latch the wrapped receiver. If this is unnecessary for some reason I can revert it out.
- Updates RmqListener to only stop consuming, and not dispose of the consumer, so the channel stays around for acking/nacking/deferring already processing messages while the receiver is draining.
- Updates SqsListener to cancel the polling loop on StopAsync instead of disposing the entire listener and abandoning in-flight messages. Currently they'll be kicked back to the queue until the visibility timeout expires, which means delayed messages on every deploy.

### DIFF
--- a/src/Testing/CoreTests/Runtime/WorkerQueues/inline_receiver_drain_and_latch.cs
+++ b/src/Testing/CoreTests/Runtime/WorkerQueues/inline_receiver_drain_and_latch.cs
@@ -51,7 +51,7 @@ public class inline_receiver_drain_waits_for_in_flight
     }
 
     [Fact]
-    public async Task drain_waits_for_in_flight_message_to_complete()
+    public async Task drain_waits_for_in_flight_message_to_complete_when_latched()
     {
         var messageBlocking = new TaskCompletionSource();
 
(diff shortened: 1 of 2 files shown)


jeremydmiller:
@benjamin-alexander-simplisafe Hey, this got brought in through a different branch, but it'll be in 5.22 shortly. Thank you for contributing this!

PR #2338, opened 2026-03-23
Fix GroupIdToPartitionKeyRule to prefer outgoing GroupId over originator's consumer group

lyall-sc:
This fix aims to solve two cases where the outgoing Kafka partition key was observed to be either:

- Kafka consumer group name (e.g. application name)
- Wolverine message id

The first case is caused by the Kafka listener setting the envelope.GroupId = config.GroupId on the incoming message. If the handler from the Kafka topic cascades an outgoing message, the GroupIdToPartitionKeyRule runs and copies this GroupId value to be the outgoing partition key.

The second case happens when the handler is processing a message from a local queue, or when publishing fresh from application code. At routing time, DetermineGroupId correctly resolves outgoing.GroupId = stream id (via ByPropertyNamed). But GroupIdToPartitionKeyRule only looks at originator.Envelope?.GroupId, which is null or unrelated, so it sets nothing, and outgoing.PartitionKey stays null. Then CreateMessage in the Kafka producer reaches its fallback: envelope.Id.ToString().

Fix the propagation priority in ApplyCorrelation:
1. Outgoing message's own GroupId (set by ByPropertyNamed) - the business partition key
2. Originator's PartitionKey (the actual Kafka message key of the incoming message)
3. Originator's GroupId as a last resort

Also implement Modify so that messages published outside a handler context (e.g. background services) have their GroupId promoted to PartitionKey when no explicit key is set.

Stop stamping envelope.GroupId = config.GroupId in KafkaListener and KafkaTopicGroupListener - the consumer group name is not meaningful metadata on the incoming envelope.

### DIFF
--- a/src/Transports/Kafka/Wolverine.Kafka/Internals/KafkaListener.cs
+++ b/src/Transports/Kafka/Wolverine.Kafka/Internals/KafkaListener.cs
@@ -47,6 +47,7 @@ public class KafkaListener : IListener, IDisposable, ISupportDeadLetterQueue
                         var envelope = mapper!.CreateEnvelope(result.Topic, message);
                         envelope.Offset = result.Offset.Value;
                         envelope.MessageType ??= _messageTypeName;
+                        if (topic.StampConsumerGroupIdOnEnvelope) envelope.GroupId = config.GroupId;
 
                         await receiver.ReceivedAsync(this, envelope);
                     }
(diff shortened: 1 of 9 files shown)


src/Transports/Kafka/Wolverine.Kafka/Internals/KafkaListener.cs
```diff
@@ -47,7 +47,6 @@ public KafkaListener(KafkaTopic topic, ConsumerConfig config,
                         var envelope = mapper!.CreateEnvelope(result.Topic, message);
                         envelope.Offset = result.Offset.Value;
                         envelope.MessageType ??= _messageTypeName;
-                        envelope.GroupId = config.GroupId;
```
jeremydmiller: This change will have to be "opt in" some how

PR #2357, opened 2026-03-26
Make the polling interval for SQL transports configurable

xwipeoutx:
This PR adds new optional configuration to the SQL listeners to set the polling interval.

This applies to MySql, Oracle, Postgres, Sql Server and Sql Lite.

## Changes
* New extension method `.PollingInterval(TimeSpan)` to configuration 
* New field `TimeSpan? PollingInterval` to the queue
* Updated Queue Listener to use that polling interval in place of `DurabilitySettings.ScheduledJobPollingTime` if it is set
* Updated tests to ensure it wires up
* Added to docs and doc samples

## Testing and validation
* Have run the unit and integration tests in the repository
* Have manually tested the configuration


jeremydmiller:
Looks good, but the timing isn't perfect. I'm calling a pause on Wolverine & Marten releases until at least Monday.

Issue #2387, opened 2026-03-31
Codegen fails with WriteAggregate on non Endpoint handler with attribute [WriteAggregate(OnMissing = Wolverine.Persistence.OnMissing.ThrowException)]


oT2:
**Describe the bug**
Codegen fails to compile non endpoint handler with `[WriteAggregate(Required = true, OnMissing = Wolverine.Persistence.OnMissing.ThrowException)]` attribute.
Works without the `OnMissing` parameter.
Works with `[ReadAggregate(Required = true, OnMissing = Wolverine.Persistence.OnMissing.ThrowException)]`.
Works with both ReadAggregate and WriteAggregate on endpoint handler. 

Minimal reproduction project can be found here : https://github.com/oT2/wolverine-writeaggregate-bug-reproduction. Using the 'test' endpoint in the swagger will fail the compilation of the handler invoked by `bus.InvokeAsync()`.

stream_entity is used before it's declaration at `if (stream_entity.Aggregate == null)`. Below are the logs when the codegen fails:

    fail: Wolverine.Runtime.WolverineRuntime[0]
      Failed to create a message handler for BugReproduction.DeleteCommand
      System.InvalidOperationException: Compilation failures!

    CS0841: Cannot use local variable 'stream_entity' before it is declared
    CS0841: Cannot use local variable 'stream_entity' before it is declared
    CS0841: Cannot use local variable 'stream_entity' before it is declared

    Code:

    // <auto-generated/>
      #pragma warning disable
      using Wolverine.Marten.Publishing;

      namespace Internal.Generated.WolverineHandlers
      {
          // START: DeleteCommandHandler254052316
          [global::System.CodeDom.Compiler.GeneratedCode("JasperFx", "1.0.0")]
          public sealed class DeleteCommandHandler254052316 : Wolverine.Runtime.Handlers.MessageHandler
          {
              private readonly Wolverine.Marten.Publishing.OutboxedSessionFactory _outboxedSessionFactory;

              public DeleteCommandHandler254052316(Wolverine.Marten.Publishing.OutboxedSessionFactory outboxedSessionFactory)
              {
                  _outboxedSessionFactory = outboxedSessionFactory;
              }

              public override async System.Threading.Tasks.Task HandleAsync(Wolverine.Runtime.MessageContext context, System.Threading.CancellationToken cancellation)
              {
                  // Building the Marten session
                  await using var documentSession = _outboxedSessionFactory.OpenSession(context);
                  // The actual message body
[...]

PR #2388, opened 2026-03-31
Fixed WriteAggregate failing to compile in non endpoint handlers

oT2:
Fixes https://github.com/JasperFx/wolverine/issues/2387

As I am not familiar with Wolverine's code and I used LLM to help me out, this is likely to have side effects.


jeremydmiller:
@oT2 I got this a little different way. Thank you for the reproduction, it was the combination of Validate() plus batch querying that got us here.

PR #2389, opened 2026-03-31
GitHub Actions improvements

dmytro-pryvedeniuk:
This PR includes some improvements

* ~Run builds for any push (not only main branch) to discover issues earlier.~ Reverted in #2398. 
* Skip publishing steps for unrelated setup to avoid false email notifications for the forked repositories.
* Allow running builds manually (helpful for troubleshooting).
* Update build actions to fix warnings related to Node.js 20. See below.
* Avoid logging attempts to connect to DB as errors. See below.

image

### DIFF
--- a/.github/workflows/dotnet.yml
+++ b/.github/workflows/dotnet.yml
@@ -53,7 +53,7 @@ jobs:
 
 
       - name: Publish CI Packages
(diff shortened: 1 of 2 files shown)


dmytro-pryvedeniuk:
It can be interesting to trigger builds only if there are actual changes in the related paths. On the other hand it requires extra efforts to keep the paths up-to-date.

jeremydmiller:
Awesome! Thank you!

PR #2391, opened 2026-03-31
Latch receiver after stopping listener on shutdown

benjamin-alexander-simplisafe:
Currently, wolverine latches all receivers immediately when shutdown is requested. This means that any messages consumed by listeners after that happens are deferred, which (for the same reasons as in https://github.com/JasperFx/wolverine/pull/2347) can have poor performance characteristics.

This change moves latching each receiver to after its respective listener is stopped, so all consumed messages get a fair shot at being processed.

This also moves to shutting down the listeners in parallel so that they don't wait to latch before the previous ones finish shutting down. This was also done to the ParallelListener, with similar exception aggregation to the CompoundListener. I think this is a general improvement to the shutdown flow, but also avoids extra bad behavior from moving latching.

### DIFF
--- a/src/Wolverine/Configuration/EndpointCollection.cs
+++ b/src/Wolverine/Configuration/EndpointCollection.cs
@@ -384,8 +384,9 @@ public class EndpointCollection : IEndpointCollection
 
     /// <summary>
     /// Immediately latch all receivers to stop picking up new messages from their internal queues.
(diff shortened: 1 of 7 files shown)


jeremydmiller:
@benjamin-alexander-simplisafe Hey, I'm going to try to get this in for the next release, but as I'm sure you completely understand, I need to review this one carefully!

PR #2445, opened 2026-04-07
Fix OpenAPI OperationId and add Summary/Description support

codeswithfists:
## Problem
When setting `OperationId` on a Wolverine HTTP route attribute (e.g. `[WolverineDelete("/{id}", OperationId = "DeleteStaffMember")]`), the value is ignored in the generated OpenAPI document. Instead, the `operationId`, `summary`, and `description` fields all default to a generated filename like `DELETE_id`.

This happens because `HttpChain` implements `IEndpointNameMetadata`, `IEndpointSummaryMetadata`, and `IEndpointDescriptionMetadata`, but is never added to the endpoint's metadata collection — so the .NET OpenAPI document generator never sees these values.

## Changes
- Add `EndpointNameMetadata(OperationId)` to endpoint metadata in `BuildEndpoint()` so the OpenAPI generator picks up the user-provided `OperationId`
- Add `Summary` and `Description` properties to `WolverineHttpMethodAttribute` to allow setting these in OpenAPI documentation
- Fix `IEndpointNameMetadata.EndpointName` and `IEndpointSummaryMetadata.Summary` implementations on `HttpChain` to use user-provided values instead of `ToString()`

### DIFF
--- a/src/Http/Wolverine.Http/HttpChain.EndpointBuilder.cs
+++ b/src/Http/Wolverine.Http/HttpChain.EndpointBuilder.cs
@@ -120,6 +120,18 @@ public partial class HttpChain : IEndpointConventionBuilder
             builder.Metadata.Add(new RouteNameMetadata(RouteName));
         }
 
+        builder.Metadata.Add(new EndpointNameMetadata(OperationId));
+
+        if (EndpointSummary.IsNotEmpty())
+        {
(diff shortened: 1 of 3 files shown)


jeremydmiller:
Got it, but it took some follow up work., Thank you!

PR #2511, opened 2026-04-14
Fix silent message loss when envelope mapper throws (RabbitMQ + MQTT)

### LINKED ISSUE #2510: RabbitMQ listener silently acks and drops messages when the envelope mapper throws
## Summary

When `_mapper.MapIncomingToEnvelope` throws, `WorkerQueueMessageConsumer.HandleBasicDeliverAsync` calls `BasicAckAsync` and returns. The message is permanently lost: not persisted to the durable inbox, not routed to the DLQ, not requeued. Only a log line remains.

## Environment

- Wolverine 5.30.0
- Wolverine.RabbitMQ + durable inbox (MSSQL)

## Affected code

`src/Transports/RabbitMQ/Wolverine.RabbitMQ/Internal/WorkerQueueMessageConsumer.cs:47-58`

```csharp
try
{
    envelope.Data = body.ToArray();
    _mapper.MapIncomingToEnvelope(envelope, properties);
}
catch (Exception e)
{
    _logger.LogError(e, "Error trying to map an incoming RabbitMQ message {MessageId} to an Envelope", properties.MessageId);
    await Channel.BasicAckAsync(envelope.DeliveryTag, false, _cancellation); // <-- silent loss
    return;
}
```

## Trigger conditions

Noticed during a MassTransit → Wolverine cutover, where the same RabbitMQ queue was fed by a legacy MassTransit producer. The mapper itself tolerates MassTransit's top-level properties (`MessageId`, `CorrelationId`, `Content-Type`, `Type`), but throws in `tryReadIncomingHeader` (`RabbitMqEnvelopeMapper.cs:71`) when a header value is `null` or a non-string/non-`byte[]` AMQP type (e.g. nested `IDictionary<string,object>` such as `MT-Host-Info` on some client versions). The `raw!.ToString()` call then NREs or produces a value that a downstream typed conversion rejects.

Independent of MassTransit, any of the following will trigger the same loss:

- a custom mapper that throws;
- `Encoding.Default` on a host with a non-UTF8 system codepage decoding a UTF-8 header (`RabbitMqEnvelopeMapper.cs:71, 85`);
- a foreign producer that omits required headers a user-defined mapper assumes present.

## Expected

An unmappable message must not be silently acknowledged. It should be routed to the broker DLX or persisted to the Wolverine dead-letter store.

## Actual

`BasicAck` discards the message.

## Inconsistency with other transports

| Transport                                                           | On mapper exception                 | Outcome         |
| ------------------------------------------------------------------- | ----------------------------------- | --------------- |
| RabbitMQ `WorkerQueueMessageConsumer.cs:55`                         | `BasicAckAsync`                     | lost            |
| Azure Service Bus (batched) `BatchedAzureServiceBusListener.cs:140` | `tryMoveToDeadLetterQueue(message)` | preserved       |
| GCP Pub/Sub `PubsubListener.cs:208-211`                             | `return false` (no ack)             | redelivered     |
| MQTT `MqttListener.cs:103`                                          | `_complete.PostAsync`               | lost (same bug) |

## Proposed fix

Replace `BasicAck` with `BasicNack(multiple: false, requeue: false)` so the broker's DLX routes the message. Wolverine provisions a DLX per queue by default, so this works out of the box for users who rely on Wolverine's defaults.

```csharp
catch (Exception e)
{
    _logger.LogError(e, "Error trying to map an incoming RabbitMQ message {MessageId} to an Envelope", properties.MessageId);
    try
    {
        await Channel.BasicNackAsync(envelope.DeliveryTag, multiple: false, requeue: false, _cancellation);
    }
    catch (Exception nackEx)
    {
        _logger.LogError(nackEx, "Failed to Nack un-mappable RabbitMQ message {MessageId}", properties.MessageId);
    }
    return;
}
```

Optional alternative: on `ISupportDeadLetterQueue` receivers, additionally persist to `wolverine_dead_letters` via `MoveToErrorsAsync` before nacking, so the message is observable in the Critter Stack dashboard even without broker DLX configuration.

## Workaround

Subclass `RabbitMqEnvelopeMapper` and wrap `MapIncomingToEnvelope` / `tryReadIncomingHeader` overrides in try/catch that sets `envelope.MessageType = "unknown/mapping-error"` instead of throwing — the handler pipeline will then dead-letter the message normally.

## Severity

High when sharing a queue with non-Wolverine producers or running custom mappers. The failure mode is worst-case: silent data loss, no exception, no DLQ trace.

## Additional context

I have already implemented a fix in my fork and will submit a PR shortly.

BlackChepo:
Closes #2510

  ## Summary

  When the incoming envelope mapper threw, both RabbitMQ's `WorkerQueueMessageConsumer` and MQTT's `MqttListener` acked the broker delivery and dropped the message. No broker
  DLQ, no entry in `wolverine_dead_letters`, no redelivery — the message was simply lost. Triggered in the wild by foreign producers (e.g. MassTransit) whose headers the
  mapper can't decode, custom mappers that throw, or header decoding on hosts with a non-UTF8 system codepage.

  Both listeners now:

  1. Synthesize `envelope.Id` (white lie — the mapper failed before populating it, but `MoveToErrorsAsync` keys by Id).
  2. Attempt `ISupportDeadLetterQueue.MoveToErrorsAsync` on the receiver so the message lands in `wolverine_dead_letters` when the durable inbox is configured.
  3. Transport-specific fallback:
     - **RabbitMQ:** `BasicNack(requeue: false)` so the broker's DLX captures the message even without durable inbox.
     - **MQTT:** always PUBACK after the dead-letter attempt — MQTT has no broker DLQ, so leaving the message unacked would cause a poison-redelivery loop.
  4. Wrap every outbound call in try/catch so a failure inside the fix path cannot propagate out of the consumer callback.

  `MqttEnvelope` now sets `Destination = topic.Uri` in its constructor, matching how `Data` and `MessageType` are already populated. Previously `Destination` was set by
  `DurableReceiver.MarkReceived`, which is bypassed when the mapper throws — causing `MessageDatabase.MoveToDeadLetterStorageAsync` to NRE on
  `envelope.Destination!.ToString()`.

  ## Transport behavior comparison (before → after)

  | Transport | Before | After |
  |---|---|---|
  | RabbitMQ | `BasicAck` → silent loss | `MoveToErrorsAsync` → broker DLX + DB when configured |
  | MQTT | `_complete.PostAsync` → silent loss | `MoveToErrorsAsync` → DB (when durable inbox) + PUBACK |
  | Azure SB / GCP Pub/Sub / Kafka / SQS / Redis / NATS / Pulsar | already correct | unchanged |

  ## Test plan

  - [x] New integration test `Bug_mapper_exception_routes_to_dlq` under `Wolverine.RabbitMQ.Tests.Bugs` — publishes a raw AMQP message to a queue with a throwing
  `IRabbitMqEnvelopeMapper` and asserts the broker DLQ receives it.
  - [x] New integration test `Bug_mapper_exception_routes_to_dlq` under `Wolverine.MQTT.Tests.Bugs` — durable inbox on SQL Server + throwing `IMqttEnvelopeMapper`; asserts an
  entry appears in `wolverine_dead_letters` with `ExceptionType` containing `InvalidOperationException`.
  - [x] Both tests fail on `main` with a "(silent loss)" exception and pass on this branch.
  - [x] Close regression sweep over RabbitMQ DLQ / interop / unknown-message suites: no new failures.
  - [x] Full MQTT test suite: 89/89 pass (1 pre-existing skip).

  ## Files changed

  - `src/Transports/RabbitMQ/Wolverine.RabbitMQ/Internal/WorkerQueueMessageConsumer.cs`
  - `src/Transports/MQTT/Wolverine.MQTT/Internals/MqttListener.cs`
  - `src/Transports/MQTT/Wolverine.MQTT/MqttEnvelope.cs` (1 line — set `Destination` in constructor)
  - `src/Transports/MQTT/Wolverine.MQTT.Tests/Wolverine.MQTT.Tests.csproj` (added `Wolverine.SqlServer` project reference)
  - New: `src/Transports/RabbitMQ/Wolverine.RabbitMQ.Tests/Bugs/Bug_mapper_exception_routes_to_dlq.cs`
  - New: `src/Transports/MQTT/Wolverine.MQTT.Tests/Bugs/Bug_mapper_exception_routes_to_dlq.cs`

### DIFF
--- a/src/Transports/MQTT/Wolverine.MQTT.Tests/Bugs/Bug_mapper_exception_routes_to_dlq.cs
+++ b/src/Transports/MQTT/Wolverine.MQTT.Tests/Bugs/Bug_mapper_exception_routes_to_dlq.cs
@@ -1,0 +1,98 @@
+using IntegrationTests;
+using JasperFx.Core;
+using Microsoft.Extensions.Hosting;
+using MQTTnet;
+using MQTTnet.Client;
+using Shouldly;
+using Wolverine.Persistence.Durability.DeadLetterManagement;
+using Wolverine.Runtime;
+using Wolverine.SqlServer;
+using Wolverine.Tracking;
+
+namespace Wolverine.MQTT.Tests.Bugs;
+
+[Collection("mosquitto")]
+public class Bug_mapper_exception_routes_to_dlq : IAsyncLifetime
+{
+    private readonly string _topic = "mapper-explosion/" + Guid.NewGuid().ToString("N");
+    private IHost _host = null!;
+
+    public async Task InitializeAsync()
+    {
+        _host = await Host.CreateDefaultBuilder()
+            .UseWolverine(opts =>
+            {
+                opts.UseMqtt(builder =>
+                {
+                    builder.WithClientOptions(client =>
+                        client.WithTcpServer(MosquittoContainerFixture.Host, MosquittoContainerFixture.Port));
+                });
+
+                opts.PersistMessagesWithSqlServer(Servers.SqlServerConnectionString, "wolverine");
+
+                opts.ListenToMqttTopic(_topic)
+                    .UseInterop(new AlwaysThrowingMqttMapper())
+                    .UseDurableInbox();
+            }).StartAsync();
+
+        // Clear prior dead-letter rows so the test is deterministic.
+        await _host.RebuildAllEnvelopeStorageAsync();
+    }
+
+    public async Task DisposeAsync()
+    {
+        if (_host is not null)
+        {
+            await _host.StopAsync();
+            _host.Dispose();
+        }
+    }
+
+    [Fact]
+    public async Task unmappable_message_is_persisted_to_wolverine_dead_letters()
+    {
+        var factory = new MqttFactory();
+        using var client = factory.CreateMqttClient();
+
+        var options = new MqttClientOptionsBuilder()
+            .WithTcpServer(MosquittoContainerFixture.Host, MosquittoContainerFixture.Port)
+            .Build();
+
+        await client.ConnectAsync(options);
+        await client.PublishStringAsync(_topic, "hello");
+
+        var storage = _host.GetRuntime().Storage;
+
+        var attempts = 0;
+        while (attempts < 40)
+        {
+            var results = await storage.DeadLetters.QueryAsync(
+                new DeadLetterEnvelopeQuery(TimeRange.AllTime()),
+                CancellationToken.None);
+
+            if (results.Envelopes.Count > 0)
+            {
+                results.Envelopes[0].ExceptionType.ShouldContain(nameof(InvalidOperationException));
+                return;
+            }
+
+            attempts++;
+            await Task.Delay(250.Milliseconds());
+        }
+
+        throw new Exception(
+            "Expected an entry in wolverine_dead_letters after a mapper exception, but none was persisted (silent loss).");
+    }
+}
+
+internal class AlwaysThrowingMqttMapper : IMqttEnvelopeMapper
+{
+    public void MapEnvelopeToOutgoing(Envelope envelope, MqttApplicationMessage outgoing)
+    {
+    }
+
+    public void MapIncomingToEnvelope(Envelope envelope, MqttApplicationMessage incoming)
+    {
+        throw new InvalidOperationException("simulated mapper failure");
+    }
+}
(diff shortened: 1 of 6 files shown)


jeremydmiller:
@BlackChepo You've been on fire! But also wish you didn't have quite so much to go fix lately:/

I'll deal w/ the CI problems

BlackChepo:
Thanks Jeremy! The good news is we're finding them before users do. 😄 Happy to keep contributing fixes as I run into things 
Wolverine is a joy to work with.

PR #2554, opened 2026-04-21
Fix RavenDb scheduled messages not being delivered when using a durable transport

Bishbulb:
## Summary

I hit this chasing down why `bus.ScheduleAsync(...)` against a durable external sender was not working in our app - we were seeing saga timeouts persisted to the `IncomingMessage` collection in RavenDB, but they were never being delivered. We observed that `RavenDbDurabilityAgent.runScheduledJobs` was never releasing these messages. After some digging, I traced it to `RavenDbMessageStore.Locking.cs` and found the issue noted below.

## Root cause

`TryAttainScheduledJobLockAsync` decides between its two branches with:
```
if (_leaderLock == null)
```
but nothing in this method (or anywhere else that matters to the scheduled lock) reads or writes `_leaderLock`.

## Fix

- Changing to use the `_scheduledLock` variable
- Ensure we set `_scheduledLock = null` at the end of the method (to match what happens in `ReleaseLeadershipLockAsync`)

### DIFF
--- a/src/Persistence/RavenDbTests/scheduled_job_locking.cs
+++ b/src/Persistence/RavenDbTests/scheduled_job_locking.cs
@@ -1,0 +1,91 @@
+using JasperFx.Core.Reflection;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Hosting;
+using Raven.Client.Documents;
+using Shouldly;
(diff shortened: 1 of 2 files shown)


jeremydmiller (approved):
Thanks for the clean fix and regression test.

PR #2555, opened 2026-04-21
Fix NRE in RavenDb durability recovery

Bishbulb:
## Summary

I ran into this running multiple Wolverine services against a single shared RavenDb envelope database. Every durability recovery pass was throwing a NullRef exception and spamming the logs every few seconds. I traced it to `RavenDbDurabilityAgent.tryRecoverIncomingMessages`.

## Root cause

The method scans all distinct `ReceivedAt` URIs in the `IncomingMessage` collection and calls `FindListenerCircuit` for each. When a URI belongs to a queue that *this* node doesn't service (because a different service owns it), `FindListenerCircuit` returns null which causes an NRE trying to access `circuit.Status`:
```
  var circuit = _runtime.Endpoints.FindListenerCircuit(receivedAt)!;
  if (circuit.Status != ListeningStatus.Accepting)
```

Note that I believe the same issue would occur in a single-service deployment if a listener URI was removed from config (or renamed), at that point any inbox documents that were still referencing the old URI would result in the same null reference exception.

## Fix
Null-check the `circuit` variable and skip it

### DIFF
--- a/src/Persistence/RavenDbTests/durability_recovery_orphaned_listener.cs
+++ b/src/Persistence/RavenDbTests/durability_recovery_orphaned_listener.cs
@@ -1,0 +1,113 @@
+using System.Reflection;
+using JasperFx.Core.Reflection;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Hosting;
+using Microsoft.Extensions.Logging;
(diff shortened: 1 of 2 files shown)


jeremydmiller (approved):
Thanks for the clean fix and regression test.

PR #2572, opened 2026-04-23
Preserve context fields on scheduled sends to non-native transports

### LINKED ISSUE #2571: Scheduled sends to non-native-scheduling transports lose context fields (TenantId, SagaId, CorrelationId, UserName)
## Summary

When an envelope is scheduled to a transport that does not support native scheduled-send (RabbitMQ, Kafka, SharedMemory, ...), `MessageRoute` wraps it via `ForScheduledSend`. The wrap eagerly serializes the inner envelope into `outer.Data` **before** `MessageBus.TrackEnvelopeCorrelation` stamps context-derived fields. When the scheduled time fires and the inner is forwarded over the transport, the receive side reconstructs an envelope with `TenantId = null`, `SagaId = null`, etc.

## Impact

Most visible on multi-tenanted sagas that use cascading `TimeoutMessage`:

- Saga runs under tenant `red` and cascades a timeout.
- The timeout is routed to RabbitMQ (no native scheduled-send) → wrapped.
- The inner envelope in `outer.Data` has no TenantId.
- On fire, the listener receives the timeout with no tenant.
- Under Marten conjoined multi-tenancy the saga lookup lands on the default tenant, misses, and `NotFound` runs silently — the saga is never completed.

The same loss affects `CorrelationId`, `UserName`, and `SagaId`. `TenantId` and `SagaId` cause correctness issues; `CorrelationId`/`UserName` cause observability/audit gaps.

## Reproduction

End-to-end in the RabbitMQ test suite: `scheduled_saga_timeout_preserves_tenant` — saga starts under `"red"`, cascaded `TimeoutMessage` routes through Rabbit, pre-fix the handler never runs (silent drop).

In-process in CoreTests: `inner_envelope_is_stamped_before_serialization` — publishes a scheduled message via the `SharedMemoryTopic` wrap path; pre-fix the delivered envelope has `TenantId = null`.

## Affected transports

Any sender with `SupportsNativeScheduledSend == false`. Confirmed: RabbitMQ, Kafka, SharedMemory. Native schedulers (Azure Service Bus, SQS with delay, etc.) are unaffected.

## Proposed fix

- Drop the eager `Data = EnvelopeSerializer.Serialize(envelope)` from `ForScheduledSend` so serialization is deferred to `EnvelopeReaderWriter.Write` — by which time stamping has happened.
- Extract `MessageBus.StampEnvelope` (virtual) containing the idempotent, context-derived stamps: `Source`, `CorrelationId`, `TenantId`, `UserName`. `MessageContext` overrides to add `SagaId`.
- In `TrackEnvelopeCorrelation`, call `StampEnvelope(outbound)`; when the envelope is a scheduled wrap, also call `StampEnvelope(inner)`.

PR: *#TBD*

ArieGato:
Fixes #2571.

## Summary

- When an envelope is scheduled to a transport without native scheduled-send (RabbitMQ, Kafka, SharedMemory, ...), `MessageRoute.ForScheduledSend` wraps it and previously eagerly serialized the inner into `outer.Data` before `TrackEnvelopeCorrelation` could stamp context fields. The inner therefore arrived on the receive side with `TenantId = null`, `SagaId = null`, `CorrelationId`/`UserName` unset — breaking multi-tenanted saga timeouts and observability chains.
- Defers serialization to `EnvelopeReaderWriter.Write`, which now unwraps `model.Message` to serialize the inner.
- Extracts a virtual `MessageBus.StampEnvelope` with the idempotent context-derived stamps (`Source`, `CorrelationId`, `TenantId`, `UserName`); `MessageContext` overrides it to add `SagaId`. `TrackEnvelopeCorrelation` calls it on the outer and, when the envelope is a scheduled wrap, also on the inner.

## Test plan

- [x] `CoreTests.Runtime.Scheduled.inner_envelope_is_stamped_before_serialization` — in-process via SharedMemory wrap path; asserts `TenantId`, `CorrelationId`, `UserName` on the delivered inner.
- [x] `Wolverine.RabbitMQ.Tests.scheduled_saga_timeout_preserves_tenant` — end-to-end multi-tenant saga with Marten + RabbitMQ + durable outbox; verifies the handler runs under the original tenant and the saga completes.
- [x] Full CoreTests suite (1347/1347) passes on net9.0.
- [x] RabbitMQ suite: new test passes; unrelated DLQ/tenancy failures are pre-existing flakes (reproduce on `main`, pass in isolation).

### DIFF
--- a/src/Testing/CoreTests/Runtime/Scheduled/inner_envelope_is_stamped_before_serialization.cs
+++ b/src/Testing/CoreTests/Runtime/Scheduled/inner_envelope_is_stamped_before_serialization.cs
@@ -1,0 +1,83 @@
+using JasperFx.Core;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Hosting;
+using Wolverine.ComplianceTests.Compliance;
+using Wolverine.Runtime;
+using Wolverine.Tracking;
+using Wolverine.Transports.SharedMemory;
+using Xunit;
+
+namespace CoreTests.Runtime.Scheduled;
+
+public class inner_envelope_is_stamped_before_serialization : IAsyncLifetime
+{
+    private IHost _host = null!;
+
+    public async Task InitializeAsync()
+    {
+        ScheduledEnvelopeCapture.Reset();
+
+        _host = await Host.CreateDefaultBuilder()
+            .UseWolverine(opts =>
+            {
+                opts.EnableRelayOfUserName = true;
+                opts.PublishAllMessages().ToSharedMemoryTopic("scheduled_tenant_topic");
+                opts.ListenToSharedMemorySubscription("scheduled_tenant_topic", "sub").ProcessInline();
+            }).StartAsync();
+    }
+
+    public async Task DisposeAsync()
+    {
+        await _host.StopAsync();
+        _host.Dispose();
+    }
+
+    [Fact]
+    public async Task scheduled_send_to_non_native_transport_preserves_context_fields()
+    {
+        var bus = (MessageBus)_host.Services.GetRequiredService<IMessageBus>();
+        bus.TenantId = "red";
+        bus.CorrelationId = "corr-123";
+        bus.UserName = "alice";
+
+        var tracked = await _host.TrackActivity()
+            .Timeout(10.Seconds())
+            .ExecuteAndWaitAsync(_ =>
+                bus.PublishAsync(new Message1(), new DeliveryOptions { ScheduleDelay = 1.Hours() }).AsTask());
+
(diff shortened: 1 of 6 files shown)


jeremydmiller:
Hey, I'm going to want a little more review on this one, so it's not going to make the release this morning

ArieGato:
Should I put any effort in this to resolve the merge conflict?

jeremydmiller:
I took your test, but went a little different way to keep overhead out of the hot path: https://github.com/JasperFx/wolverine/pull/2605

ArieGato:
Great you fixed it, thnx!

PR #2575, opened 2026-04-23
Add missing project to wolverine.slnx

saithis:


saithis:
Without it wolverine doesn't compile, since it can't find the project.

jeremydmiller:
Got it w/ a different PR

PR #2589, opened 2026-04-26
Honor [Consumes] / IAcceptsMetadata on form endpoints

rmasciarella:
## Summary

`HttpChain.fillRequestType` only honors `[Consumes]` / `IAcceptsMetadata` when the endpoint has a body request type and is **not** a form endpoint (`HasRequestType && !IsFormData`). Form endpoints fall through and `apiDescription.SupportedRequestFormats` stays empty.

ASP.NET Core 10's `OpenApiDocumentService.GetFormRequestBody` then defaults to `application/x-www-form-urlencoded`:

```csharp
if (supportedRequestFormats.Count == 0)
{
    supportedRequestFormats = [new ApiRequestFormat { MediaType = "application/x-www-form-urlencoded" }];
}
```

This silently drops `[Consumes("multipart/form-data")]` on file-upload endpoints, so generated clients (Orval, NSwag, Kiota) emit `URLSearchParams` bodies instead of `FormData` — file uploads break at runtime.

## Fix

Extend `CreateApiDescription` so that for endpoints with form-source parameters and no formats already populated, `IAcceptsMetadata` flows into `SupportedRequestFormats`:

```csharp
if (apiDescription.SupportedRequestFormats.Count == 0 &&
    apiDescription.ParameterDescriptions.Any(p =>
        p.Source == BindingSource.Form || p.Source == BindingSource.FormFile))
{
    foreach (var metadata in Endpoint!.Metadata.OfType<IAcceptsMetadata>())
    {
        foreach (var contentType in metadata.ContentTypes)
        {
            apiDescription.SupportedRequestFormats.Add(new ApiRequestFormat
            {
                MediaType = contentType
            });
        }
    }
}
```

## Semantics

- **No `[Consumes]`** → no change. Wolverine still emits zero formats; MS OpenAPI still defaults to `application/x-www-form-urlencoded`.
- **`[Consumes("multipart/form-data")]`** → multipart is added to `SupportedRequestFormats`; MS OpenAPI no longer adds the urlencoded default (its `count == 0` guard); spec correctly emits multipart.
- **Non-form body endpoints** → unaffected (the form-parameter guard short-circuits).

## Regression test

New test endpoint:
```csharp
[WolverinePost("/form/multipart-consumes")]
[Consumes("multipart/form-data")]
public static string MultipartConsumes([FromForm] string value) => value ?? "";
```

New test:
```csharp
[Fact]
public void form_endpoints_honor_consumes_metadata_for_supported_request_formats()
{
    var chain = HttpChains.ChainFor("POST", "/form/multipart-consumes");
    var apiDescription = chain!.CreateApiDescription("POST");

    apiDescription.SupportedRequestFormats
        .Select(f => f.MediaType)
        .ShouldContain("multipart/form-data");
}
```

Fails on `main`:
```
but was actually
Failed!  - Failed: 1, Passed: 0
```

Passes after the fix:
```
Passed!  - Failed: 0, Passed: 1
```

## Test plan

- [x] New test fails on `main`
- [x] New test passes after fix
- [x] Form + OpenAPI / Swashbuckle filtered run: 64 passed, 0 failed, 8 pre-existing skips

## Related

Independent of #2587 (which fixes a separate `ParameterDescriptor` null-deref in the same file). Either can land first.

### DIFF
--- a/src/Http/Wolverine.Http.Tests/using_form_parameters.cs
+++ b/src/Http/Wolverine.Http.Tests/using_form_parameters.cs
@@ -541,12 +541,13 @@ public class using_form_parameters : IntegrationContext
     public void form_endpoints_honor_consumes_metadata_for_supported_request_formats()
     {
         // fillRequestType only seeds SupportedRequestFormats from
-        // IAcceptsMetadata when the endpoint has a body request type and is
-        // not a form endpoint. Form endpoints used to fall through and rely
-        // on ASP.NET Core OpenAPI's "application/x-www-form-urlencoded"
-        // default, which silently dropped [Consumes("multipart/form-data")]
-        // and caused generated clients (Orval, NSwag) to emit URLSearchParams
-        // bodies instead of multipart for file-upload endpoints.
+        // IAcceptsMetadata when the endpoint has a body request type, is
+        // not a form endpoint, and is not a GET. Form endpoints used to
+        // fall through and rely on ASP.NET Core OpenAPI's
+        // "application/x-www-form-urlencoded" default, which silently
+        // dropped [Consumes("multipart/form-data")] and caused generated
+        // clients (Orval, NSwag) to emit URLSearchParams bodies instead of
+        // multipart for file-upload endpoints.
         var chain = HttpChains.ChainFor("POST", "/form/multipart-consumes");
         var apiDescription = chain!.CreateApiDescription("POST");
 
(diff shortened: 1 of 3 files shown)


jeremydmiller:
@rmasciarella Thank you for taking this on. OpenAPI is about my least favorite part of working on Wolverine.

jeremydmiller (approved):
Approving — the form-endpoint OpenAPI fix is correct and the regression test demonstrates the bug clearly. Will follow up with the Copilot review's dedup recommendation in a separate PR.

rmasciarella:
@jeremydmiller no worries it just came up for me while vibe coding, i did run it through a few different models from different angles to make sure before i put it up, glad that it was helpful.

Issue #2619, opened 2026-04-29
CustomizeOutgoingMessagesOfType<T> cannot override Envelope.Source — overwritten by TrackEnvelopeCorrelation before serialization


devployment:
**Describe the bug**

`MessageBus.TrackEnvelopeCorrelation` unconditionally overwrites `Envelope.Source` with `WolverineOptions.ServiceName`, silently discarding any per-message Source set earlier (for example by a `CustomizeOutgoingMessagesOfType<T>` rule or a `DeliveryOptions` override). 

This makes it impossible to publish CloudEvents with a spec-valid per-message source URI when using .InteropWithCloudEvents() on Azure Service Bus (and other transports), because the CloudEventsMapper reads Envelope.Source after TrackEnvelopeCorrelation has already clobbered it.

The same method already takes care to other preserve pre-populated values.

Source location: [`src/Wolverine/Runtime/MessageBus.cs#L351-L371`](https://github.com/JasperFx/wolverine/blob/main/src/Wolverine/Runtime/MessageBus.cs#L351-L371)

**To Reproduce**

1. Configure a Wolverine app with the Azure Service Bus transport and enable CloudEvents interop on the outbound endpoint:

```c#
opts.ServiceName = "MY.API";
opts.PublishMessage<MyMessage>()
    .ToAzureServiceBusTopic("my-topic")
    .InteropWithCloudEvents()
    .UseDurableOutbox();
```

2. Register a customizer that sets a per-message CloudEvents source URI:

```c#
opts.Policies.CustomizeOutgoingMessagesOfType<MyMessage>(
    (env, msg) => env.Source = $"https://api.example.com/my/apy/{msg.UserId}");
```

3. Publish a MyMessage message: `await bus.PublishAsync(new MyMessage(userId));`

4. Inspect the outgoing CloudEvent on the Service Bus topic

**Expexted behavior**

The serialized CloudEvent's source attribute should be the per-message URI assigned by the customizer (https://api.example.com/my/api/<id>), exactly like CorrelationId, TenantId, and UserName are preserved when set ahead of time.

**Actual behavior**

The CloudEvent's source is "MY.API" (the value of WolverineOptions.ServiceName). The customizer's assignment is silently discarded because TrackEnvelopeCorrelation runs after the routing/customization stage and unconditionally overwrites Envelope.Source.

Wolverine: 5.32.1
Wolverine.AzureServiceBus: 5.32.1
.NET: 9.0 / 10.0
[...]

Issue #2545, opened 2026-04-20
Automatic correlation ID propagation through event-sourced handler chains


RikDriever:
## Summary

When an event triggers a handler chain that emits a new event, there is no first-class way to propagate the correlation ID through the chain. The only current workaround is injecting `IDocumentSession` into the `Handle` method and setting `CorrelationId` manually, which makes handlers impure.

## Motivation

A common pattern in event-sourced systems is a **todo-list**: an event creates a pending task in a projection, a handler performs the work and emits a completion event, and the completion event marks that task as done — identified by the originating correlation ID.

```
Event A emitted                  → todo-list projection adds task (keyed by correlation ID)
→ side effect raised
→ message dispatched to handler
→ handler emits Event B          → todo-list projection marks task complete (by correlation ID)
```

For Event B to close the task opened by Event A, it must carry the same correlation ID. Currently the only way to achieve this in an `[AggregateHandler]` is to inject `IDocumentSession` and set `CorrelationId` manually:

```csharp
public static Events Handle(MyCommand command, MyAggregate aggregate, IDocumentSession session)
{
    session.CorrelationId = /* somehow retrieve originating correlation id */;
    // ...
}
```

This leaks infrastructure concerns into domain handler code. A `Before(IDocumentSession)` method works as a workaround, but still requires explicit per-handler wiring.

## Proposed Approach

The chain has two handoff points where the correlation ID needs to flow automatically:

1. **Event → outgoing message** — when a Marten event subscription raises a side effect or publishes a message, the outgoing envelope should automatically be stamped with the `CorrelationId` from the originating event's metadata.

2. **Incoming envelope → IDocumentSession** — when Wolverine's transactional middleware sets up the `IDocumentSession` for an `[AggregateHandler]`, it should automatically copy the envelope's `CorrelationId` onto the session, so any events appended during that invocation inherit it.

With both in place, the todo-list pattern works with no handler code changes. For cases where explicit control is needed, `DeliveryOptions` is the natural extension point:

```csharp
await bus.PublishAsync(message, new DeliveryOptions { CorrelationId = "..." });
```

## Acceptance Criteria

- [ ] Events emitted by a handler triggered by another event automatically inherit the originating correlation ID
- [ ] No `IDocumentSession` injection required in handler code purely for correlation purposes
- [ ] Explicit override via `DeliveryOptions` when a different correlation ID is desired

🤖 Generated with [Claude Code](https://claude.com/claude-code)

[...]

PR #2620, opened 2026-04-29
fix(MessageBus): preserve per-message Envelope.Source instead of clob…

devployment:
TrackEnvelopeCorrelation unconditionally overwrote Envelope.Source with WolverineOptions.ServiceName, which silently discarded any per-message Source set by a CustomizeOutgoingMessagesOfType<T> rule or other DeliveryOptions override. This made it impossible to publish CloudEvents with a spec-valid per-message `source` URI via .InteropWithCloudEvents().

The same method already takes care to preserve pre-populated CorrelationId (#2545), TenantId, and UserName values — Source is now treated the same way: ServiceName is applied only as a fallback when Source is empty.

Updated the existing track_envelope_correlation test to clear Source first (it was implicitly asserting the buggy clobbering behavior since ObjectMother.Envelope() pre-sets Source) and added two new tests covering the override-preservation and fallback paths.

#2619

### DIFF
--- a/src/Testing/CoreTests/Runtime/MessageContextTests.cs
+++ b/src/Testing/CoreTests/Runtime/MessageContextTests.cs
@@ -128,6 +128,11 @@ public class MessageContextTests
     [Fact]
     public void track_envelope_correlation()
     {
+        // Source is preserved when already set on the envelope (see
+        // track_envelope_correlation_does_not_override_existing_source);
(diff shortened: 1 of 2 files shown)


jeremydmiller:
@devployment Hey, I'm just going to ask you to use a custom header instead. Source has a specific meaning in Wolverine world.

Issue #2401, opened 2026-04-02
API Versioning support for Wolverine.HTTP


jeremydmiller:
## Summary

Wolverine.HTTP currently has no built-in API versioning support. Users must manually encode versions in route templates (e.g. `/api/v1/orders`), which works but doesn't support header-based or query string versioning, and doesn't provide deprecation metadata in OpenAPI specs.

## Proposal

Add convention-based API versioning to Wolverine.HTTP, potentially including:

1. **Version attribute on endpoints** — e.g. `[WolverineGet("/orders", Version = "2.0")]` or a dedicated `[ApiVersion("2.0")]` attribute
2. **Built-in `IHttpPolicy`** that automatically prepends/appends the version number to routes (e.g. a policy that turns `/orders` into `/api/v1/orders` or `/api/v2/orders` based on the declared version)
3. **OpenAPI metadata** — versioned endpoints should emit proper deprecation and version metadata in generated OpenAPI documents
4. **Deprecation support** — ability to mark older versions as deprecated

## Context

This is a commonly requested feature. All comparable frameworks support versioning:
- **Minimal APIs / MVC**: via `Asp.Versioning.Http` / `Asp.Versioning.Mvc` (URL, header, query string strategies)
- **FastEndpoints**: via `Asp.Versioning.Http` integration

The preferred direction is a **convention-based approach native to Wolverine** (rather than just wrapping `Asp.Versioning`), leveraging `IHttpPolicy` to keep the implementation consistent with Wolverine's existing extensibility model.

## Open Questions

- Should we support multiple versioning strategies (URL prefix, header, query string) or focus on URL-based first?
- Should the `IHttpPolicy` be opt-in or enabled by default when version attributes are detected?
- Integration with `Asp.Versioning` as an alternative/complementary approach?

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2633, opened 2026-04-30
Native API versioning for Wolverine.Http (closes #2627)

### LINKED ISSUE #2627: Improved support for Wolverine.Http versionning
Hi Jeremy — picking up on #2401  because we're hitting it on a large system overhaul. Before opening a PR I'd like to validate the direction with you.

## Proposal: Wolverine-native versioning, no Asp.Versioning dependency

After looking at how Asp.Versioning v10 + `Microsoft.AspNetCore.OpenApi` plug together (one document per version via `WithDocumentPerVersion()`, driven by `IApiVersionDescriptionProvider` and `ApiDescription.GroupName`), I think trying to bridge Wolverine into that pipeline is the wrong call. It would force us back into the MVC/ApiExplorer machinery that 1.13 deliberately stepped away from.

Instead, I'd propose a fully native design that mirrors the Asp.Versioning *experience* without the dependency.

## v1 scope

Just enough to be useful, nothing speculative:

- `[ApiVersion("1.0", Deprecated = true, Sunset = "2026-12-31")]` on endpoint methods or classes, plus a `Version =` shortcut on the existing `[WolverineGet]` / `[WolverinePost]` / etc. attributes.
- `HttpChain.Version` as a first-class property (not opaque metadata), set by `Configure(HttpChain)` > method attribute > class attribute precedence.
- An internal `ApiVersioningPolicy : IHttpPolicy` that:
  - Resolves the version per chain
  - Rewrites `RoutePattern` for URL-segment strategy (`v{version}` prefix, configurable)
  - Adds postprocessors for `Deprecation` / `Sunset` / `api-supported-versions` headers
  - Detects `(verb, route, version)` collisions at bootstrap
- Configuration via `opts.UseApiVersioning(v => { ... })` with `UnversionedPolicy` (`PassThrough` / `RequireExplicit` / `AssignDefault`) so existing apps don't break.
- **URL-segment strategy only in v1.** Header / query string / combined readers can come later — they're a smaller % of real-world use, and they're where the versioning-via-conventions story gets messy.

## The non-negotiable: native SwaggerUI / Scalar support

This is where I want to make sure we're aligned, because it drives the OpenAPI design more than the routing does.

The whole point of versioned APIs in 2026 is that people expect to drop in SwaggerUI or Scalar and get a version dropdown. With Asp.Versioning + Microsoft.AspNetCore.OpenApi this works because:

1. Each version produces a separate document at `/openapi/v1.json`, `/openapi/v2.json`, etc.
2. `app.DescribeApiVersions()` returns the list of versions, which the UI middleware iterates over to populate the dropdown.

I'd like Wolverine to provide the equivalent natively, by extending the OpenAPI provider added in 1.13:

- The provider emits one document per discovered version (filtering chains by `chain.Version`), plus optionally an aggregated document.
- A `MapWolverineOpenApi()` (or a switch on the existing call) registers `/openapi/{documentName}.json` routes for every discovered version automatically.
- A public `app.DescribeWolverineApiVersions()` returns the list of versions so users can wire up SwaggerUI / Scalar like this:

```csharp
app.UseSwaggerUI(options =>
{
    foreach (var v in app.DescribeWolverineApiVersions().Reverse())
        options.SwaggerEndpoint($"/openapi/{v.DocumentName}.json", v.DisplayName);
});

// or for Scalar:
app.MapScalarApiReference(options =>
{
    foreach (var v in app.DescribeWolverineApiVersions())
        options.AddDocument(v.DocumentName, v.DisplayName);
});
```

Same ergonomics as the Microsoft stack, zero Asp.Versioning dependency. If we don't ship this, users will go back to wrapping Asp.Versioning anyway and we've solved nothing.

## Open questions before I start coding

1. **Policy execution order**: mutating `RoutePattern` needs to happen before code generation and before `RouteEndpoint` emission. Is the current `IHttpPolicy` phase early enough, or do we need a dedicated pre-policy phase?
2. **Provider extension surface**: where in the 1.13 OpenAPI provider is the right hook to filter chains and emit N documents? Happy to dig in but a pointer would save time.
3. **Naming**: `HttpChain.Version` vs `HttpChain.ApiVersion`? Latter is clearer, former avoids type-name collision with a future `ApiVersion` record.
4. **Asp.Versioning interop**: I'd keep this *out* of core. A small `WolverineFx.Http.AspVersioning.Interop` package can ship later for users who already have the infra, copying `chain.Version` into `ApiVersionMetadata`. Agreed?

If the direction looks right to you I'll start with v1 and put a draft PR up. Happy to iterate.

outofrange-consulting:
Closes #2627. Sibling reference: #2401.

## Summary

Adds native API versioning to `WolverineFx.Http` with first-class SwaggerUI / Scalar compatibility. Single dependency on `Asp.Versioning.Abstractions` (no `Asp.Versioning.Http`, no `Microsoft.AspNetCore.OpenApi`).

```csharp
[ApiVersion("1.0")]
public static class OrdersV1Endpoint
{
    [WolverineGet("/orders")] public static OrdersV1Response Get() => new(...);
}

app.MapWolverineEndpoints(opts =>
{
    opts.UseApiVersioning(v =>
    {
        v.Sunset("3.0").On(DateTimeOffset.Parse("2027-01-01")).WithLink(new Uri("https://example.com/migrate"));
        v.Deprecate("1.0").On(DateTimeOffset.Parse("2026-12-31"));
    });
});
```

Produces `/v1/orders`, `/v2/orders`, `/v3/orders` automatically, plus per-version OpenAPI documents and the right RFC headers on every response.

## What's included

- **URL-segment versioning**. `[ApiVersion("1.0")]` → `/v1/orders` by default. `UrlSegmentPrefix` and `UrlSegmentVersionFormatter` are both overridable.
- **Method-wins attribute resolution**. Method-level `[ApiVersion]` takes precedence over class-level. Multiple `[ApiVersion]` attributes on a single method are explicitly rejected at startup.
- **`UnversionedPolicy`** with three modes: `PassThrough` (default), `RequireExplicit` (throws at startup if any chain lacks `[ApiVersion]`), `AssignDefault` (assigns a configured `DefaultVersion`).
- **Sunset / deprecation policies** with fluent `Sunset("1.0").On(date).WithLink(uri, title?, type?)` builders. Attribute-driven `[ApiVersion(..., Deprecated = true)]` takes precedence over options-driven deprecation per chain.
- **Response headers**: RFC 9745 `Deprecation` (with `true` token when no date), RFC 8594 `Sunset`, RFC 8288 `Link`, plus `api-supported-versions`. Two independent toggles (`EmitDeprecationHeaders`, `EmitApiSupportedVersionsHeader`).
- **Duplicate detection**. Two chains with the same `(verb, original-route, version)` triple throw at bootstrap with both display names in the message.
- **OpenAPI document partitioning** via `IEndpointGroupNameMetadata` (already plumbed through to `ApiDescription.GroupName`). `WolverineApiVersioningOpenApiOptions.DocumentNameStrategy` lets users customise the document name (default `v{major}`).
- **`app.DescribeWolverineApiVersions()`** returns one description per discovered version for wiring SwaggerUI / Scalar dropdowns.
- **Sample Swashbuckle `IOperationFilter`** (`WolverineApiVersioningSwaggerOperationFilter`) lives in `WolverineWebApi` as copy-paste reference code — sets `OpenApiOperation.Deprecated` and emits an `x-api-versioning` extension carrying sunset date and links.

## Dependency choice

`Asp.Versioning.Abstractions` 10.0.0 only — **not** `Asp.Versioning.Http`. Routing is driven from `IHttpPolicy`, so we need the types (`ApiVersion`, `ApiVersionAttribute`, `SunsetPolicy`, `DeprecationPolicy`, `ApiVersionMetadata`, `LinkHeaderValue`) but not the readers, version sets, or DI policy manager. Smaller surface, no MVC-centric assumptions.

`Asp.Versioning.Http` is verified absent from the transitive closure (`dotnet list package --include-transitive`).

## Default URL-segment format

`[ApiVersion("1.0")]` produces `/v1/orders` (major-only) by default. Users wanting `/v1.0/orders` set `UrlSegmentVersionFormatter = v => v.ToString()`. Documentation explicitly notes both options.

## Implementation notes

- Wolverine.Http does not have an internal OpenAPI document builder; it surfaces metadata via `IApiDescription` and external tools (Swashbuckle, MS.AspNetCore.OpenApi) read it. The policy attaches `IEndpointGroupNameMetadata` so document partitioning works automatically with Swashbuckle's `DocInclusionPredicate` or the equivalent ASP.NET Core mechanism.
- `ApiVersionEndpointHeaderState` and `ApiVersionHeaderWriter` are public. The state record is part of the observable contract for user OpenAPI filters; the writer is referenced by Wolverine's dynamic codegen (empirically verified — `internal` produces CS0122 in generated source).

## Tests

| Suite | Count | Status |
|---|---|---|
| Full `Wolverine.Http.Tests` (net9.0) | 707 | ✅ |
| ApiVersioning suite | 73 | ✅ |
| Swashbuckle integration | 7 | ✅ |
| Alba integration scenarios (`api_versioning_integration_tests.cs`) | 10 | ✅ |

Smoke test on the WolverineWebApi sample confirms `/v1/orders`, `/v2/orders`, `/v3/orders` all 200 with the expected `Deprecation` / `Sunset` / `Link` / `api-supported-versions` headers, and `/swagger/{default,v1,v2,v3}/swagger.json` all serve correctly.

## Test plan

- [ ] `dotnet build` succeeds on net8.0 / net9.0 / net10.0
- [ ] `dotnet test src/Http/Wolverine.Http.Tests` is green
- [ ] Run `WolverineWebApi` and visit `/swagger` — confirm dropdown shows `default`, `v1`, `v2`, `v3` and the v1 endpoint shows the deprecated marker
- [ ] Visit `/scalar/v1` — confirm the Scalar UI renders versioned operations
- [ ] `curl -i /v1/orders` shows `Deprecation`, `Link`, `api-supported-versions` headers
- [ ] `curl -i /v3/orders` shows `Sunset`, `Link`, `api-supported-versions` headers
- [ ] `dotnet list src/Http/Wolverine.Http package --include-transitive` confirms `Asp.Versioning.Http` is **not** present

## Documentation

`docs/guide/http/versioning.md` covers attribute rules, URL-segment strategy, the three `UnversionedPolicy` modes, sunset/deprecation policies, OpenAPI integration (Swashbuckle + Scalar + MS.AspNetCore.OpenApi). Cross-linked from `metadata.md`. Sidebar entry registered in `docs/.vitepress/config.mts`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,11 @@
+# Changelog
+
+## Unreleased
+
+### WolverineFx.Http
+
+- Added native API versioning support via `Asp.Versioning.Abstractions` 10.x. Supports URL-segment versioning
+  (`/v1/...`, `/v2/...`), sunset/deprecation policies with RFC 9745/8594/8288 response headers, and automatic
+  OpenAPI document partitioning with Swashbuckle/Scalar/Microsoft.AspNetCore.OpenApi. No dependency on
+  `Asp.Versioning.Http` — versioning is driven entirely via `IHttpPolicy`.
+  See [versioning guide](docs/guide/http/versioning.md).
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -4,6 +4,7 @@
   </PropertyGroup>
   <ItemGroup>
     <PackageVersion Include="Alba" Version="8.5.2" />
+    <PackageVersion Include="Asp.Versioning.Abstractions" Version="10.0.0" />
     <PackageVersion Include="AWSSDK.S3" Version="4.0.22.1" />
     <PackageVersion Include="AWSSDK.SimpleNotificationService" Version="4.0.2.14" />
     <PackageVersion Include="AWSSDK.SQS" Version="4.0.2.14" />
@@ -83,6 +84,7 @@
     <PackageVersion Include="RavenDB.TestDriver" Version="7.0.2" />
     <PackageVersion Include="Refit" Version="8.0.0" />
     <PackageVersion Include="Refit.HttpClientFactory" Version="8.0.0" />
+    <PackageVersion Include="Scalar.AspNetCore" Version="2.14.8" />
     <PackageVersion Include="Shouldly" Version="4.3.0" />
     <PackageVersion Include="Spectre.Console" Version="0.55.0" />
     <PackageVersion Include="StackExchange.Redis" Version="2.9.17" />
--- a/docs/.vitepress/config.mts
+++ b/docs/.vitepress/config.mts
@@ -264,6 +264,7 @@ const config: UserConfig<DefaultTheme.Config> = {
                         {text: 'Exception Handling', link: '/guide/http/exception-handling'},
                         {text: 'Policies', link: '/guide/http/policies.md'},
                         {text: 'OpenAPI Metadata', link: '/guide/http/metadata'},
+                        {text: 'API Versioning', link: '/guide/http/versioning'},
                         {text: 'Using as Mediator', link: '/guide/http/mediator'},
                         {text: 'Multi-Tenancy and ASP.Net Core', link: '/guide/http/multi-tenancy'},
                         {text: 'Publishing Messages', link: '/guide/http/messaging'},
--- a/docs/guide/http/index.md
+++ b/docs/guide/http/index.md
@@ -273,4 +273,10 @@ leads to code that is hard to reason about and hence, potentially buggy in real 
 need this functionality in the real world, so here you go. 
 :::
 
+## API Versioning <Badge type="tip" text="5.36" />
+
+Wolverine.Http has native support for versioning your HTTP APIs over time using URL-segment strategies (e.g.
+`/v1/orders`, `/v2/orders`), per-version sunset and deprecation policies with RFC 9745/8594 response headers,
+and automatic OpenAPI document partitioning. See the [HTTP API Versioning guide](./versioning.md) for full details.
+
 
--- a/docs/guide/http/metadata.md
+++ b/docs/guide/http/metadata.md
@@ -58,6 +58,12 @@ public static void Configure(HttpChain chain)
 <sup><a href='https://github.com/JasperFx/wolverine/blob/main/src/Http/Wolverine.Http/Runtime/PublishingEndpoint.cs#L15-L33' title='Snippet source file'>snippet source</a> | <a href='#snippet-sample_programmatic_one_off_openapi_metadata' title='Start of snippet'>anchor</a></sup>
 <!-- endSnippet -->
 
+::: tip
+For HTTP API versioning that partitions the OpenAPI output into one document per version, see the
+[Versioning guide](./versioning.md). It covers the multi-document `SwaggerDoc` setup, `DocInclusionPredicate`,
+`DescribeWolverineApiVersions()`, and Scalar integration.
+:::
+
 ## Swashbuckle and Wolverine
 
 [Swashbuckle](https://github.com/domaindrivendev/Swashbuckle.AspNetCore) is de facto the OpenAPI tooling for ASP.Net Core
--- a/docs/guide/http/versioning.md
+++ b/docs/guide/http/versioning.md
@@ -1,0 +1,319 @@
+# HTTP API Versioning <Badge type="tip" text="5.36" />
+
+Wolverine.Http provides native API versioning support that lets you evolve your HTTP services over time without
+breaking existing clients. Versioned endpoints coexist at separate URL prefixes (e.g. `/v1/orders` and `/v2/orders`),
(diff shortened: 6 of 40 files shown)


jeremydmiller:
@outofrange-consulting There's enough for another release tomorrow. You okay with this as is? I'll review too, but from your perspective?

outofrange-consulting:
I am, I stongly advice you to review the API introduced and the dependency, and I’ll be happy the take it further if you want header/querystring support later ;)

jeremydmiller:
@outofrange-consulting Let's make somebody ask for the header/querystring usage later. I'm starting a review.

jeremydmiller:
@outofrange-consulting We're going for it. It's a little convenient maybe that this will get in just before Wolverine 6, so we can make any API adjustments based on feedback for 6.0.

Thanks for taking this on!

PR #2642, opened 2026-04-30
Add badges to readme

dmytro-pryvedeniuk:
This is how readme can look with the badges.

image


jeremydmiller:
@dmytro-pryvedeniuk I'm sorry, but I just don't want to do this one

PR #2661, opened 2026-05-02
fix(http): emit api-versioning headers on non-2xx responses

outofrange-consulting:
Follow-up to #2633.

## Summary
`ApiVersionHeaderWriter` now registers headers via `Response.OnStarting` from the first frame of every relevant chain — versioning headers (`Deprecation`, `Sunset`, `Link`, `api-supported-versions`) are emitted on all framework-produced responses regardless of status code (2xx, IResult short-circuits, validation `ProblemDetails`, middleware-IResult exits).

Previously emitted as a `Postprocessor`, which Wolverine codegen skips when `MaybeEndWithResultFrame` / `MaybeEndWithProblemDetailsFrame` returns out of the generated handler — leaving 4xx error paths without headers.

## Implementation
- New `ApiVersionHeaderFinalizationPolicy` registered in `MapWolverineEndpoints` after `configure()`. Runs last, positions the writer call at `chain.Middleware[0]` (outranks `FluentValidation` / `DataAnnotations` / `RequestId` / `TenantId` frames that also insert at index 0). Idempotent: skips when writer is already at index 0; `Debug.Assert` invariant under DEBUG.
- `ApiVersionHeaderWriter.WriteAsync` is the chain frame; uses `static` lambda + `state` argument to avoid per-request closure allocation. Re-fetches `ApiVersionEndpointHeaderState` from the matched endpoint inside the `OnStarting` callback because the endpoint can be re-routed by middleware between the chain frame and header-flush time. Resolves itself via `RequestServices.GetRequiredService<ApiVersionHeaderWriter>()` — fail-fast if DI is misconfigured.
- `ApiVersionHeaderWriter.WriteVersioningHeadersTo(HttpContext, ApiVersionEndpointHeaderState)` is exposed as `public` (`[EditorBrowsable(Advanced)]`) so the documented 5xx-middleware snippet can call the same logic without copy-pasting RFC-9745 emission.
- `ChainsRequiringHeaderEmission` (`IReadOnlySet<HttpChain>`) gives the finalization policy O(1) lookup.

## Out of scope
Responses produced by the global ASP.NET Core exception handler bypass the chain pipeline entirely. Users wanting headers on 5xx wire them via separate middleware. Documented in `versioning.md` with a snippet that calls `WriteVersioningHeadersTo`.

## Test plan
- [x] `api_versioning_error_path_header_tests` — 5 exit-path tests:
  - 404 IResult short-circuit
  - 400 validation `ProblemDetails`
  - 401 middleware short-circuit
  - 200 success
  - **5xx via global `UseExceptionHandler` — headers ABSENT (regression test pinning the documented out-of-scope, asserts response body marker `"global-exception-handler"`)**
- [x] `ApiVersionHeaderWriterTests` — added missing-writer-in-DI fail-fast test
- [x] `ApiVersioningPolicyHeaderWiringTests` — finalization idempotency + DEBUG-only `finalization_assert_fires_when_writer_was_displaced`
- [x] All ApiVersioning-suite tests green: 80/80
- [x] Full `Wolverine.Http.Tests` suite green: 714/714

## Sample app
Four split fixtures in `WolverineWebApi/ApiVersioning/`:
- `OrdersV1NotFoundEndpoint.cs`
- `OrdersV1CreateEndpoint.cs`
- `OrdersV1RestrictedEndpoint.cs`
- `OrdersV1ThrowsEndpoint.cs` (intentional throw, scoped under `UseWhen(.../v1/orders/throws, UseExceptionHandler)`)


jeremydmiller:
@outofrange-consulting I'm not good with this one doing the service location. Is there any way you could try to make this a bit more idiomatic Wolverine and eliminate the service location hooks in the hot path? 

Sorry man, but I don't want to have to support this one.

outofrange-consulting:
@jeremydmiller addressed — dropped the service-locator hop in `ApiVersionHeaderWriter.WriteAsync`. The OnStarting callback now captures `this` directly (writer is already DI-resolved by the generated handler via `MethodCall.For`). Trades the static-lambda micro-opt for one closure alloc per request, same cost ASP.NET middleware pays for `OnStarting` everywhere. 5 lines of code, all 79 ApiVersioning tests green.

jeremydmiller:
@outofrange-consulting Hey, can you just combine the two PRs in one rebased PR please? No hurry on my part though

outofrange-consulting:
Superseded by #2683 (combined with #2662 and rebased as requested).

PR #2676, opened 2026-05-05
Fix PublishingRelay misrouting default-tenant events under conjoined tenancy

### LINKED ISSUE #2675: PublishingRelay drops TenantId for default-tenant events, misrouting them in conjoined tenancy
## Summary

In `Wolverine.Marten`, the `PublishingRelay` subscription forwards Marten events
to Wolverine via `IMessageBus.PublishAsync`. For events whose `TenantId` is
**not** the default tenant, it correctly attaches
`DeliveryOptions.TenantId = e.TenantId`. For events whose `TenantId` **is**
`StorageConstants.DefaultTenantId`, it falls through with no `DeliveryOptions`,
relying on the bus context's `TenantId`. `WolverineSubscriptionRunner` has set
that to `operations.Database.Identifier`, which conflates database identity
with tenant identity.

In conjoined-tenancy stores this misroutes default-tenant events: they are
dispatched under the database identifier rather than `*DEFAULT*`, the handler
chain runs in the wrong tenant scope (or as a silent no-op), but Marten still
advances the subscription sequence because publishing succeeded.

## Affected versions

Reproduced against `WolverineFx.Marten` 5.12.1 and current `main`
(HEAD `cfbfbdd4`). The relevant code has not changed.

## Reproduction conditions

1. A Marten document store configured with
   `opts.Events.TenancyStyle = TenancyStyle.Conjoined`.
2. Events appended under the default tenant (no explicit tenant id passed when
   starting/appending the stream).
3. `PublishEventsToWolverine(...)` registered as a Marten subscription.
4. A Wolverine handler that depends on the inbound envelope's `TenantId` (or
   on tenant-scoped state).

The handler observes a tenant id equal to the Marten database identifier
(e.g. `"Main"`) rather than `*DEFAULT*`. Named-tenant events route correctly.

## Root cause (traced through three files)

**1. `WolverineSubscriptionRunner` sets the bus tenant to the database identifier:**

```csharp
// src/Persistence/Wolverine.Marten/Subscriptions/WolverineSubscriptionRunner.cs
public override async Task<IChangeListener> ProcessEventsAsync(
    EventRange page, ISubscriptionController controller,
    IDocumentOperations operations, CancellationToken cancellationToken)
{
    var context = new MessageContext(_runtime);
    context.TenantId = operations.Database.Identifier;   // <-- (1)
    ...
}
```

(Note: the Polecat sibling at
`src/Persistence/Wolverine.Polecat/Subscriptions/WolverineSubscriptionRunner.cs`
already does the right thing — it reads `operations.TenantId` and only sets the
bus context when it's a real, non-default tenant.)

**2. `PublishingRelay` skips `DeliveryOptions` for default-tenant events:**

```csharp
// src/Persistence/Wolverine.Marten/Subscriptions/PublishingRelay.cs
foreach (var e in page.Events)
{
    if (_publishers.TryFind(e.EventType, out var publisher))
    {
        await publisher.PublishAsync(e, bus);
    }
    else
    {
        if (e.TenantId != StorageConstants.DefaultTenantId)
        {
            await bus.PublishAsync(e, new DeliveryOptions { TenantId = e.TenantId });
        }
        else
        {
            await bus.PublishAsync(e);                   // <-- (2) no DeliveryOptions
        }
    }
}
```

**3. `MessageBus` outbound flow falls back to the bus context's `TenantId`:**

```csharp
// src/Wolverine/Runtime/MessageBus.cs
outbound.TenantId ??= TenantId;                          // <-- (3) falls back to (1)
```

Result: outbound envelope `TenantId == operations.Database.Identifier`, which
is not `*DEFAULT*` in conjoined-tenancy setups with a non-default database
identifier. The receiver dispatches under the wrong tenant.

## Expected behaviour

Default-tenant Marten events relayed via `PublishEventsToWolverine` should be
dispatched to handlers under the default tenant (`*DEFAULT*`), regardless of
the underlying database identifier or tenancy style.

## Compatibility / impact analysis

The fix only changes behaviour for cases that hit the default-tenant
fallthrough. The recommended fix below is gated on
`Events.TenancyStyle == Conjoined`, so it touches only the broken case:

| Setup | `IEvent.TenantId` | Hits fallthrough? | Pre-fix `Envelope.TenantId` | Post-fix (recommended) `Envelope.TenantId` |
|-------|-------------------|-------------------|------------------------------|----------------------------------------------|
| No tenancy (default Marten store) | `*DEFAULT*` | yes | `Database.Identifier` (e.g. `"Main"`) | `Database.Identifier` (unchanged) |
| Conjoined, default-tenant event | `*DEFAULT*` | yes | `Database.Identifier` (the bug) | `*DEFAULT*` (correct) |
| Conjoined, named-tenant event | `"one"` | no — `!=` branch | `"one"` | `"one"` (unchanged) |
| Multi-tenanted databases (`m.MultiTenantedDatabases(...AddSingleTenantDatabase("tenant1"...))`) | `"tenant1"` (Marten stamps the named tenant on the IEvent) | no — `!=` branch | `"tenant1"` | `"tenant1"` (unchanged) |
| Per-tenant ancillary stores with custom `Database.Identifier` used *as* the tenant (`TenancyStyle.Single`) | `*DEFAULT*` | yes | the custom identifier | the custom identifier (unchanged) |

The only behavioural delta is the targeted bug case (row 2). Every other row,
including the per-tenant-ancillary-store pattern that uses
`Database.Identifier` as the tenant, keeps its pre-fix behaviour exactly
because the override is only activated when the events store is in conjoined
mode.

## Suggested fix

Realign the relay with Marten's tenant model in conjoined stores. Marten
already provides the authoritative tenant on every event via
`IEvent<T>.TenantId` — for default-tenant events under conjoined tenancy that
value is `StorageConstants.DefaultTenantId`, exactly as it should be when the
tenant lives in the row, not the database. The relay should propagate that
value verbatim instead of letting it fall back to the bus context's
`TenantId`, which `WolverineSubscriptionRunner` has set to the database
identifier (a database identity, not a tenant identity).

To keep the change scoped strictly to the broken case and avoid any risk to
the per-tenant-ancillary-store pattern in the table above, gate the override
on the events store's tenancy style:

```csharp
else
{
    var tenancyStyle = ((IDocumentSession)operations).DocumentStore
        .Options.Events.TenancyStyle;

    if (e.TenantId != StorageConstants.DefaultTenantId
        || tenancyStyle == TenancyStyle.Conjoined)
    {
        await bus.PublishAsync(e, new DeliveryOptions { TenantId = e.TenantId });
    }
    else
    {
        await bus.PublishAsync(e);
    }
}
```

For non-conjoined stores this preserves the existing behaviour exactly. For
conjoined stores it ensures the event's authoritative tenant id reaches the
outbound envelope.

A more comprehensive fix would also align `WolverineSubscriptionRunner` with
its Polecat sibling: read `operations.TenantId` (the session's tenant) and
only set the bus context when it is a real, non-default tenant. That removes
the conflation at the source rather than papering over it in the relay, but
it carries broader compatibility implications that would need a separate
discussion.

## Tests

I have two regression tests ready to contribute alongside a PR:

1. `carry_default_tenant_id_through_under_conjoined_tenancy` — sets up a
   conjoined store, publishes a default-tenant event, and asserts
   `Envelope.TenantId == StorageConstants.DefaultTenantId`. Reproduces the bug
   on `main` (asserts `"*DEFAULT*"`, gets the database identifier `"Main"`).
   Passes after the fix.
2. `non_conjoined_store_preserves_legacy_default_tenant_fallthrough` —
   companion test that uses a plain non-conjoined Marten store and asserts the
   relay still falls through to the bus context's tenant id (the database
   identifier), guaranteeing the per-tenant-ancillary-store pattern remains
   intact. Passes both before and after the fix.

The existing `carry_the_tenant_id_through_on_the_subscription` test already
covers the named-tenant code path and continues to pass.

## Workaround

Consumers can register `PublishEvent<T>` with an explicit publisher that
always sets `DeliveryOptions.TenantId = @event.TenantId`, including for
`*DEFAULT*`. After applying that, both default-tenant and named-tenant relays
dispatch into the correct handler context.

svenclaesson:
Closes #2675

## Summary

Under conjoined tenancy, `PublishingRelay`'s default-tenant fallthrough let `WolverineSubscriptionRunner`'s `bus.TenantId` (set to `operations.Database.Identifier`) leak onto the outbound envelope. Default-tenant events were dispatched under the database identifier (e.g. `"Main"`) instead of `*DEFAULT*`, the handler chain ran under the wrong tenant scope (or silently no-op'd), but Marten still advanced the subscription sequence because publishing succeeded.

This change gates the override on `Events.TenancyStyle == Conjoined`, so only the broken case is touched. Non-conjoined stores keep the existing fallthrough exactly, preserving setups that legitimately use `Database.Identifier` as the message tenant (e.g. per-tenant ancillary stores with `TenancyStyle.Single`).

## Compatibility

| Setup | Pre-fix `Envelope.TenantId` | Post-fix `Envelope.TenantId` |
|-------|------------------------------|--------------------------------|
| No tenancy (default Marten store) | `Database.Identifier` | `Database.Identifier` (unchanged) |
| Conjoined, default-tenant event | `Database.Identifier` (the bug) | `*DEFAULT*` (correct) |
| Conjoined, named-tenant event | `"one"` | `"one"` (unchanged) |
| `MultiTenantedDatabases(...AddSingleTenantDatabase("tenant1"...))` | `"tenant1"` | `"tenant1"` (unchanged) |
| Per-tenant ancillary stores with custom `Database.Identifier` (`TenancyStyle.Single`) | the custom identifier | the custom identifier (unchanged) |

The only behavioural delta is the targeted bug case. Every other row keeps its pre-fix behaviour because the override only activates in conjoined mode.

## Tests

Two regression tests added in `MartenSubscriptionTests`:

1. **`carry_default_tenant_id_through_under_conjoined_tenancy`** — sets up a conjoined store, publishes a default-tenant event, asserts `Envelope.TenantId == StorageConstants.DefaultTenantId`. Reproduces the bug on `main` (asserts `"*DEFAULT*"`, gets `"Main"`); passes after the fix.
2. **`non_conjoined_store_preserves_legacy_default_tenant_fallthrough`** — companion test using a plain non-conjoined Marten store; asserts the relay still falls through to the bus context's tenant id, guarding the per-tenant-ancillary-store pattern against accidental over-application. Passes both before and after.

`MartenSubscriptionTests` runs green on net8/net9/net10 (12/12).

## Test plan

- [x] Reproduce the bug with a failing test on `main`
- [x] Apply fix; regression test passes
- [x] Companion test confirms non-conjoined fallthrough preserved
- [x] Full `MartenSubscriptionTests` suite green on net8.0, net9.0, net10.0
- [ ] Full `MartenTests` suite (rerun in progress; previously: 433/439 with 6 unrelated infrastructure flakes — postgres connection pool exhaustion + agent-distribution timeouts; none touch `PublishingRelay`)

See #2675 for the full root-cause trace through `WolverineSubscriptionRunner` → `PublishingRelay` → `MessageBus`, and a side-by-side with the Polecat sibling that already handles tenancy correctly.

### DIFF
--- a/src/Persistence/MartenSubscriptionTests/subscriptions_end_to_end.cs
+++ b/src/Persistence/MartenSubscriptionTests/subscriptions_end_to_end.cs
@@ -333,6 +333,140 @@ public class subscriptions_end_to_end
     }
 
     [Fact]
+    public async Task non_conjoined_store_preserves_legacy_default_tenant_fallthrough()
+    {
+        // Companion to carry_default_tenant_id_through_under_conjoined_tenancy:
+        // for non-conjoined stores the relay must keep falling through to the bus
+        // context's TenantId, so any setup that relied on the database identifier as
+        // the message tenant (e.g. per-tenant ancillary stores with a custom
+        // Database.Identifier) keeps working.
+        await dropSchema();
+
+        using var host = await Host.CreateDefaultBuilder()
+            .UseWolverine(opts =>
+            {
+                opts.Durability.Mode = DurabilityMode.Solo;
+
+                opts.Policies.UseDurableLocalQueues();
+
+                opts.Services.AddMarten(m =>
+                    {
+                        m.DisableNpgsqlLogging = true;
+                        m.Connection(Servers.PostgresConnectionString);
+                        m.DatabaseSchemaName = "subscriptions";
+                        // Default (single) tenancy on the events store
+                    }).IntegrateWithWolverine()
+                    .UseLightweightSessions()
+                    .PublishEventsToWolverine("Publish", x =>
+                    {
+                        x.PublishEvent<AEvent>();
+                    });
+            }).StartAsync();
+
+        var store = host.DocumentStore();
+
+        var daemon = await store.BuildProjectionDaemonAsync();
+
+        await daemon.StartAllAsync();
+
+        Func<IMessageContext, Task> writeEvents = async _ =>
+        {
+            await using var session = store.LightweightSession();
+            session.Events.StartStream(Guid.NewGuid(), new AEvent(), new AEvent());
+            await session.SaveChangesAsync();
+
+            await daemon.WaitForNonStaleData(30.Seconds());
+        };
+
+        var tracked = await host
+            .TrackActivity()
+            .ExecuteAndWaitAsync(writeEvents);
+
+        var aEnvelopes = tracked.MessageSucceeded.Envelopes()
+            .Where(x => x.Message is IEvent<AEvent>)
+            .ToList();
+
+        aEnvelopes.ShouldNotBeEmpty();
+
+        // Pre-fix and post-fix behaviour for non-conjoined stores: the relay falls through
+        // and envelope.TenantId is the bus context value that
+        // WolverineSubscriptionRunner set from operations.Database.Identifier
+        // (e.g. "Main" for a single-database Marten store). Crucially it must NOT be the
+        // default-tenant marker — that would mean the fix had over-applied and dropped the
+        // database-identifier-as-tenant convention used by some ancillary-store setups.
+        foreach (var envelope in aEnvelopes)
+        {
+            envelope.TenantId.ShouldNotBeNull();
+            envelope.TenantId.ShouldNotBe(JasperFx.StorageConstants.DefaultTenantId);
+        }
+    }
+
+    [Fact]
+    public async Task carry_default_tenant_id_through_under_conjoined_tenancy()
+    {
+        // Regression: PublishingRelay used to drop DeliveryOptions for default-tenant
+        // events, letting WolverineSubscriptionRunner's bus.TenantId
+        // (= operations.Database.Identifier) leak onto the outbound envelope. Under
+        // conjoined tenancy where the database identifier is not the tenant, that
+        // misrouted default-tenant events into the wrong tenant context.
+        await dropSchema();
+
+        using var host = await Host.CreateDefaultBuilder()
+            .UseWolverine(opts =>
+            {
+                opts.Durability.Mode = DurabilityMode.Solo;
+
+                opts.Policies.UseDurableLocalQueues();
+
+                opts.Services.AddMarten(m =>
+                    {
+                        m.DisableNpgsqlLogging = true;
+                        m.Connection(Servers.PostgresConnectionString);
+                        m.DatabaseSchemaName = "subscriptions";
+                        m.Events.TenancyStyle = Marten.Storage.TenancyStyle.Conjoined;
+                    }).IntegrateWithWolverine()
+                    .UseLightweightSessions()
+                    .PublishEventsToWolverine("Publish", x =>
+                    {
+                        x.PublishEvent<AEvent>();
+                    });
+            }).StartAsync();
+
+        var store = host.DocumentStore();
+
+        var daemon = await store.BuildProjectionDaemonAsync();
+
+        await daemon.StartAllAsync();
+
+        Func<IMessageContext, Task> writeEvents = async _ =>
+        {
+            // No tenant arg -> events get IEvent.TenantId == StorageConstants.DefaultTenantId
+            await using var session = store.LightweightSession();
+            session.Events.StartStream(Guid.NewGuid(), new AEvent(), new AEvent());
+            await session.SaveChangesAsync();
+
+            await daemon.WaitForNonStaleData(30.Seconds());
+        };
+
+        var tracked = await host
+            .TrackActivity()
+            .ExecuteAndWaitAsync(writeEvents);
+
+        var aEnvelopes = tracked.MessageSucceeded.Envelopes()
+            .Where(x => x.Message is IEvent<AEvent>)
+            .ToList();
+
+        aEnvelopes.ShouldNotBeEmpty();
+
+        foreach (var envelope in aEnvelopes)
+        {
+            // Marten's IEvent for a default-tenant event carries TenantId == "*DEFAULT*";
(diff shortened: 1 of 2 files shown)


svenclaesson:
Broader MartenTests rerun completed: **434/439 pass** on net9.0 with this fix applied.

The 5 failures are pre-existing infrastructure flakes unrelated to `PublishingRelay`:

| Test | Failure mode |
|------|---------------|
| `Persistence.marten_durability_end_to_end.sending_recovered_messages_when_sender_starts_up` | message recovery timing |
| `Distribution.basic_agent_mechanics_versioned_composition.start_with_multiple_databases_on_one_single_node` | agent-distribution timeout |
| `Distribution.basic_agent_mechanics_versioned_composition.spread_databases_out_via_host` | agent-distribution timeout |
| `Saga.When_starting_an_order.should_not_be_completed` | `Failed to setup resource Main of type WeaselDatabase` |
| `Saga.not_found_usage.try_to_call_handle_on_already_expired_invitation` | `Npgsql.PostgresException 53300: sorry, too many clients already` |

Verified by inspection that none reference `PublishingRelay`, `PublishEventsToWolverine`, or `SubscribeToEvents`. A baseline run on `main` (without this fix) showed 433/439 with the same failure modes — i.e. the fix didn't introduce any regression and slightly *improved* the count (one fewer flake hit).

jeremydmiller:
@svenclaesson I'm going to change this one just a wee bit to try to remove the runtime logic and the downcasting tomorrow, but the test will go in as is.

jeremydmiller:
@svenclaesson I did this just a little bit differently, but you're listed as the co-author of the new PR

PR #2677, opened 2026-05-05
Outbox behavior alignment with EF Core transaction model

Ferchke7:
#### EF Core allows two valid approaches:

- implicit transactions via `SaveChanges()`
- explicit transactions via `BeginTransaction()` (with isolation levels, etc.)

#### The new implementation removes this enforcement:

- If the calling code starts an explicit transaction, it will be used.
- If not, EF Core’s implicit transaction behavior is respected.

#### In other words, transaction ownership is now fully delegated to the application layer, making Wolverine more predictable and easier to integrate into existing unit-of-work or repository-based architectures.

```csharp

app.MapGet("/msg", async (IDbContextOutbox outbox, AppDbContext db) =>
{
    await db.Orders.AddAsync(new Order());
    outbox.Enroll(db);
    await outbox.PublishAsync(new Msg("New order!"));
    await db.SaveChangesAsync();
    await outbox.FlushOutgoingMessagesAsync();
});

```

now it acts like a publish and doesn't enforce starting the transaction

### DIFF
--- a/src/Persistence/EfCoreTests/eager_idempotency_with_non_wolverine_mapped_db_context.cs
+++ b/src/Persistence/EfCoreTests/eager_idempotency_with_non_wolverine_mapped_db_context.cs
@@ -106,6 +106,7 @@ public class eager_idempotency_with_non_wolverine_mapped_db_context : IClassFixt
         };
 
         var transaction = new EfCoreEnvelopeTransaction(dbContext, context);
+        await dbContext.Database.BeginTransactionAsync();
         await transaction.PersistOutgoingAsync([envelope1, envelope2]);
         await dbContext.Database.CurrentTransaction!.CommitAsync();
 
--- a/src/Persistence/EfCoreTests/eager_idempotency_with_non_wolverine_mapped_db_context.cs
+++ b/src/Persistence/EfCoreTests/eager_idempotency_with_non_wolverine_mapped_db_context.cs
@@ -106,7 +106,6 @@ public class eager_idempotency_with_non_wolverine_mapped_db_context : IClassFixt
         };
 
         var transaction = new EfCoreEnvelopeTransaction(dbContext, context);
-        await dbContext.Database.BeginTransactionAsync();
         await transaction.PersistOutgoingAsync([envelope1, envelope2]);
         await dbContext.Database.CurrentTransaction!.CommitAsync();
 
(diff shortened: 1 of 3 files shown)


jeremydmiller:
@Ferchke7 I'm very deeply unenthusiastic about anyone from outside the core touching those internals. And this:

```
app.MapGet("/msg", async (IDbContextOutbox outbox, AppDbContext db) =>
{
    await db.Orders.AddAsync(new Order());
    outbox.Enroll(db);
    await outbox.PublishAsync(new Msg("New order!"));
    await db.SaveChangesAsync();
    await outbox.FlushOutgoingMessagesAsync();
});
```

is not necessarily idiomatic Wolverine usage. 

I'll look much closer in the next release cycle to try to get this in, but there's so much potential for this being problematic and I see at least one problem in the PR, that this will take time on my part.

Ferchke7:
> @Ferchke7 I'm very deeply unenthusiastic about anyone from outside the core touching those internals. And this:
> 
> ```
> app.MapGet("/msg", async (IDbContextOutbox outbox, AppDbContext db) =>
> {
>     await db.Orders.AddAsync(new Order());
>     outbox.Enroll(db);
>     await outbox.PublishAsync(new Msg("New order!"));
>     await db.SaveChangesAsync();
>     await outbox.FlushOutgoingMessagesAsync();
> });
> ```
> 
> is not necessarily idiomatic Wolverine usage.
> 
> I'll look much closer in the next release cycle to try to get this in, but there's so much potential for this being problematic and I see at least one problem in the PR, that this will take time on my part.

I understand your concern, but there are real use cases for this. Some teams don’t use Wolverine as the core framework they want to integrate it alongside an existing architecture, similar to how MassTransit is often used.
In that case, transaction control and flexibility become important.

jeremydmiller:
Why would you be using it *with* MassTransit at the same time? That's definitely not something we're planning for.

And you can use `IDbContextOutbox<AppDbContext>` up above to reduce the boilerplate

Oh, nevermind, I see what you're saying. I think I'll go a little differently and completely separate out the EF Core transaction management for inside of Wolverine from outside.

But still going to tell you that you'll have much cleaner, simpler code using idiomatic Wolverine.

jeremydmiller:
@Ferchke7 I'm sorry, this isn't going to make it into 5.38. Look for 5.39 early next week. Maybe Monday. 

It's just a bandwidth problem on my part.

Ferchke7:
> @Ferchke7 I'm sorry, this isn't going to make it into 5.38. Look for 5.39 early next week. Maybe Monday.
> 
> It's just a bandwidth problem on my part.

Thanks!

jeremydmiller:
@Ferchke7 And I change my mind again, I think this goes in as is. Will be in Wolverine 5.39 today or tomorrow. By Monday at worst.

Issue #2685, opened 2026-05-06
Ability to Dynamically Add Mqtt Brokers as Listeners at Runtime


jeremydmiller:
Just what it sounds like. We need a new API that allows users to add a new MQTT broker at runtime and activate message listening across the entire running cluster.

I assume we'll build this similar to the managed tenancy model. 

Only an API for now

uniquelau:
The real world situation we are working with:

- Not possible to use a top level wildcard due to security enforcement by 3rd party
- This means we will dynamically be registering IoT devices
- Each device must be subscribed (or removed) individually 
- At the moment, we will be handling between 300 to 10,000 devices

If this results in an API we can hit, we can then co-ordinate execution.

Rough thoughts:

- What happens if we removed an item? Does the in-memory Q still process and then terminate
- How are failures handled?

jeremydmiller:
Done for now

PR #2699, opened 2026-05-07
[1/N] dynamic listeners — foundation only (GH-2685)

jeremydmiller:
First slice of #2685. Foundation only — establishes the contract surface every `IMessageStore` implementation will satisfy in subsequent PRs, but ships **zero behaviour change** for existing apps. The opt-in flag defaults to `false` and every existing `IMessageStore` returns `NullListenerStore.Instance`.

## What lands in this PR

- **`IListenerStore`** abstraction in `Wolverine.Persistence.Durability`. Three URI-based operations: `RegisterListenerAsync`, `RemoveListenerAsync`, `AllListenersAsync`. Idempotent register/remove. Transport-agnostic — every entry is a single `Uri` so the same registry shape works for MQTT, RabbitMQ, Kafka, Service Bus, etc.
- **`IMessageStore.Listeners`** property added to the contract; every existing implementation returns `NullListenerStore.Instance` for now. The multi-tenant store delegates to its main store (listener URIs aren't tenant-scoped).
- **`WolverineOptions.Durability.EnableDynamicListeners`** flag, **default `false`**. Database-backed stores must NOT create the listener table unless the flag is set — users upgrading Wolverine see no schema migration churn.
- **Compliance contract** added to `MessageStoreCompliance` (7 new tests): empty-on-clean, register persists, register is idempotent, remove removes, remove is idempotent on unknown URI, register-then-remove-then-re-register, all-listeners-includes-every-registered. Each test pass-throughs when `thePersistence.Listeners is NullListenerStore` so providers adopt the contract incrementally — once a provider opts in, the same suite validates real persistence end-to-end.
- **`DynamicListenersDefaultsTests`** locks down that the opt-in flag defaults to `false` on a fresh `DurabilitySettings` AND on a fresh `WolverineOptions`, plus the no-op contract on `NullListenerStore`. **5/5 passing.**

## What's deferred to subsequent PRs (still tracked by #2685)

- Real RDBMS-backed `IListenerStore` in `Wolverine.RDBMS` (covers Postgres / SqlServer / MySql / Oracle / Sqlite via the shared base) + opt-in schema gate in each provider's `AllObjects()`.
- Marten-backed `IListenerStore`.
- `DynamicListenerAgentFamily` for cluster coordination, with Solo-mode local fan-out, registered conditionally during `UseWolverine` (not in the `WolverineOptions` ctor — to avoid race/order-of-operations issues).
- **MQTT-specific multiplexed listener** — one shared `MqttListener` + one `BufferedReceiver`, N broker subscriptions, topic-name routing inside `ReceiveAsync`. Targets the 10K-IoT-device scaling case from @uniquelau's comment on #2685; static `ListenToMqttTopic(...)` callers are unaffected.
- `IWolverineRuntime` extension methods for MQTT topic-name → URI convenience.
- New section in the MQTT guide + new section in the Durability guide.
- Follow-up issue for the remaining persistence impls (RavenDb, CosmosDb, Polecat) and additional transport coverage (RabbitMQ, AWS SQS, Azure Service Bus, Kafka, Pulsar).

## Naming decisions baked in

- `IMessageStore.Listeners : IListenerStore` per @jeremydmiller's direction
- `RegisterListenerAsync()` / `RemoveListenerAsync()` / `AllListenersAsync()`
- `Durability.EnableDynamicListeners` (boolean property style, matches `Durability.NodeAssignmentHealthCheckTracingEnabled` precedent)
- Compliance lives in the existing `MessageStoreCompliance` (not a new file)

## Test plan
- [x] `dotnet test src/Testing/CoreTests --filter DynamicListenersDefaultsTests` — **5/5 passed**
- [x] All 10 persistence projects build clean (`Wolverine.RDBMS`, `Wolverine.Postgresql`, `Wolverine.SqlServer`, `Wolverine.MySql`, `Wolverine.Oracle`, `Wolverine.Sqlite`, `Wolverine.Marten`, `Wolverine.RavenDb`, `Wolverine.CosmosDb`, `Wolverine.Polecat`)
- [x] No existing test broken — the compliance contract is pass-through wherever `Listeners is NullListenerStore` (i.e. everywhere, until provider impls land)

## Why this is shipped as a draft

This is staged work; the headline feature for #2685 (Lau's IoT broker case) needs the deferred pieces to actually function. The foundation is shipping first to:
1. Lock the API shape in CI before more code piles on it
2. Get reviewer eyes on the contract while the surface is small
3. Exercise the compliance contract on every existing persistence provider as a no-op (regression guard)

Marking `[1/N]` so the staged plan is visible. Promote to ready-for-review once the next PR lands the real RDBMS + Marten + agent + MQTT pieces.

### DIFF
--- a/src/Wolverine/Persistence/Durability/MultiTenantedMessageStore.cs
+++ b/src/Wolverine/Persistence/Durability/MultiTenantedMessageStore.cs
@@ -358,6 +358,13 @@ public partial class MultiTenantedMessageStore : IMessageStore, IMessageInbox, I
     public IDeadLetters DeadLetters => this;
     public IScheduledMessages ScheduledMessages => Main.ScheduledMessages;
     public INodeAgentPersistence Nodes => this;
+
+    // Multi-tenant store delegates dynamic-listener registration to the main
+    // store. Listener URIs aren't tenant-scoped — registering the same URI
+    // across tenants would create duplicate listeners — so the master is
+    // authoritative for the registry.
+    public IListenerStore Listeners => Main.Listeners;
+
     public IMessageStoreAdmin Admin => this;
 
     public DatabaseDescriptor Describe()
--- a/src/Persistence/Oracle/Wolverine.Oracle/OracleMessageStore.cs
+++ b/src/Persistence/Oracle/Wolverine.Oracle/OracleMessageStore.cs
@@ -113,6 +113,12 @@ internal partial class OracleMessageStore : IMessageDatabase, IMessageInbox, IMe
     public IMessageInbox Inbox => this;
     public IMessageOutbox Outbox => this;
     public INodeAgentPersistence Nodes => _nodes!;
+
+    // Real Oracle-backed listener store wired in once the schema bits land.
+    // Default no-op keeps the Wolverine boot path clean while
+    // EnableDynamicListeners is false (no schema, no behavior).
+    public IListenerStore Listeners { get; protected set; } = NullListenerStore.Instance;
+
     public IMessageStoreAdmin Admin => this;
     public IDeadLetters DeadLetters => this;
     public string Name { get; set; }
--- a/src/Persistence/Wolverine.CosmosDb/Internals/CosmosDbMessageStore.cs
+++ b/src/Persistence/Wolverine.CosmosDb/Internals/CosmosDbMessageStore.cs
@@ -66,6 +66,12 @@ public partial class CosmosDbMessageStore : IMessageStoreWithAgentSupport
     public IMessageInbox Inbox => this;
     public IMessageOutbox Outbox => this;
     public INodeAgentPersistence Nodes => this;
+
+    // Default no-op listener store. CosmosDB-backed listener registry is a
+    // follow-up implementation; stays a no-op while EnableDynamicListeners
+    // is false (default).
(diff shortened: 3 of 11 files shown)


src/Wolverine/Persistence/Durability/MultiTenantedMessageStore.cs
```diff
@@ -358,6 +358,13 @@ public void Initialize(IWolverineRuntime runtime)
     public IDeadLetters DeadLetters => this;
     public IScheduledMessages ScheduledMessages => Main.ScheduledMessages;
     public INodeAgentPersistence Nodes => this;
+
+    // Multi-tenant store delegates dynamic-listener registration to the main
+    // store. Listener URIs aren't tenant-scoped — registering the same URI
+    // across tenants would create duplicate listeners — so the master is
+    // authoritative for the registry.
+    public IListenerStore Listeners => Main.Listeners;
```
uniquelau, line 366: Understood, that is OK for our situation.

We would then associate the listener data with a tenant as the device will belong to a tenant, but that can happen outside of this process.

uniquelau (approved):

jeremydmiller:
@uniquelau It's going to take a couple follow ups, this isn't the whole show:)

PR #2707, opened 2026-05-08
docs(logging): correct OpenTelemetry AddMeter sample

stigrune:
The OTel sample on the logging page suggests registering `AddMeter("Wolverine")`, but the actual exported meter is named `Wolverine:{ApplicationName}`.

Since `AddMeter` requires an exact match, the bare name silently captures zero metrics. Switching the sample to `AddMeter("Wolverine*")`.


jeremydmiller:
@stigrune Ugh, good catch!

PR #2708, opened 2026-05-08
fix(http): honor [RoutePrefix] when no global prefix is configured

### LINKED ISSUE #2705: [RoutePrefix] is completely ignored when API Versioning is enabled
**Describe the bug**
When using the newly introduced API versioning feature alongside the `[RoutePrefix]` attribute on an endpoint class, the route prefix is completely ignored in the generated route. The resulting route only contains the API version and the method-level path, skipping the class-level prefix entirely.

**To Reproduce**
1. Configure Wolverine to use API Versioning in your app setup:
```cs
app.MapWolverineEndpoints(opts => {
    opts.UseApiVersioning(v => {
        v.DefaultVersion = new ApiVersion(1);
        v.UnversionedPolicy = UnversionedPolicy.AssignDefault;
    });
});
```
2. Create an endpoint class with a `[RoutePrefix]` and an endpoint method:
```cs
[RoutePrefix("items")]
public static class TodoCreationEndpoint {
    [WolverinePost("/create")]
    public static void Post()
    {
        // Code
    }
}
```
3. Run the application and inspect the registered endpoints.
4. See error: The generated route is `v1/create` instead of the expected `v1/items/create`.

**Expected behavior**
The generated route should respect and combine both the API version and the `[RoutePrefix]`. In the provided example, the route should be `v1/items/create`.

**Screenshots**
N/A

**Desktop**
 - OS: Windows 11 Pro, 25H2, 26200.8328
 - Browser: N/A (Backend routing issue)
 - Version: WolverineFx 5.36.2

**Additional context**
This was recently discussed in the Discord server. @outofrange-consulting requested that I open this issue with a concrete example to help track and implement a fix. As Jeremy noted in the chat, the API versioning and route prefixing features were introduced independently and currently don't interact properly when combined.

outofrange-consulting:
Fixes #2705.

`RoutePrefixPolicy.Apply` had an early-exit that skipped iterating chains whenever there was no `GlobalRoutePrefix` and no `NamespacePrefixes`. That short-circuit also bypassed the per-handler `[RoutePrefix]` attribute, so a class-level prefix declared on its own was silently dropped.

You normally wouldn't notice — the route just ended up missing one segment. API Versioning made it obvious because the version-segment policy still ran and produced things like `/v1/create` instead of `/v1/items/create` for:

```csharp
[RoutePrefix("items")]
public static class TodoCreationEndpoint
{
    [WolverinePost("/create")]
    public static void Post() { }
}
```

The fix is just to drop the early-exit. `DeterminePrefix` already returns `null` when nothing applies, so the loop is a no-op for chains without any prefix.

## Tests

- `apply_honors_attribute_prefix_when_no_global_or_namespace_prefix_configured` — unit test for the `Apply` path that wasn't covered before (existing tests only exercised `DeterminePrefix` in isolation, which never tripped the bug).
- `gh_2705_attribute_prefix_combines_with_api_versioning` — exact reproduction from the issue: `[RoutePrefix("items")]` + `[WolverinePost("/create")]` + versioning with `DefaultVersion = ApiVersion(1)` and `UnversionedPolicy.AssignDefault`. Asserts the final route is `/v1/items/create`.

Verified red → green by stashing the fix (test fails with `/v1/create`) and restoring it.

### DIFF
--- a/src/Http/Wolverine.Http.Tests/RoutePrefixTests.cs
+++ b/src/Http/Wolverine.Http.Tests/RoutePrefixTests.cs
@@ -162,6 +162,56 @@ public class RoutePrefixTests
         // Should match the most specific (longest) namespace
         prefix.ShouldBe("api/http/tests");
     }
+
+    [Fact]
+    public void apply_honors_attribute_prefix_when_no_global_or_namespace_prefix_configured()
+    {
+        // Regression for GH-2705: when no global/namespace prefix is configured, the
+        // policy used to short-circuit before iterating chains, silently dropping every
+        // [RoutePrefix] attribute. Surfaced when API Versioning was enabled because the
+        // version-segment policy still ran, producing routes like "v1/create" instead of
+        // "v1/items/create" for [RoutePrefix("items")].
+        var options = new WolverineHttpOptions();
+
+        var chain = HttpChain.ChainFor<PrefixedEndpoints>(x => x.GetOrders());
+        chain.RoutePattern!.RawText.ShouldBe("/orders");
+
+        var policy = new RoutePrefixPolicy(options);
+        policy.Apply(new[] { chain }, new JasperFx.CodeGeneration.GenerationRules(), null!);
+
+        chain.RoutePattern!.RawText.ShouldBe("/v2/orders/orders");
+    }
+
+    [Fact]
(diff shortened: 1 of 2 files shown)


jeremydmiller:
@outofrange-consulting Thank you!

PR #2709, opened 2026-05-08
Preserve correlation/trace context across Global Partitioning interceptor

lyall-sc:
The GlobalPartitionedInterceptor re-routes inbound external messages through Wolverine's routing so they land on the correct internal shard. Previously it built a fresh MessageBus(_runtime) with no inbound envelope, so PropagateHeadersRule.ApplyCorrelation short-circuited (originator.Envelope was null) and the re-published envelope dropped correlation, tenant id, and distributed-trace continuity. Trace context restarted at the interceptor hop and any PropagateIncomingHeadersToOutgoing allowlist was silently ignored for globally-partitioned messages.

The interceptor now constructs a per-receive RouteBus seeded with the inbound envelope, copies CorrelationId/TenantId/GroupId via DeliveryOptions, and brackets the publish in an Activity rooted at envelope.ParentId so the re-routed envelope continues the original trace. Custom headers are not copied wholesale — they flow only via PropagateIncomingHeadersToOutgoing, matching cascade semantics. The interceptor ctor is simplified to (IReceiver, IWolverineRuntime); topologies and logger are derived from runtime.


jeremydmiller:
Using the slightly different #2710, but with attribution

PR #2716, opened 2026-05-11
Add RabbitMq DeclarePassive option to config interface used by DeclareExchange()

saithis:
Sorry, I forgot this last time :/


jeremydmiller:
@saithis What's the point of this? It's not used by anything

saithis:
I can use it when I declare the exchange in our project. We know the exchange is there, but it is not owned by our project and we don't have management permissions on it. So we want to declare it passively.

Since my last PR about 2 weeks ago wolverine can do that, but I forgot the property on the interface, so I have to cast the config object to the real class in the DeclareExchange() config method at the moment to set it. 

With this PR I don't need the cast anymore.

```
.DeclareExchange("some.exchange", e => {
    e.DeclarePassive = true; // this works with this PR now without needing to cast "e" to the real class
})
```

Edit: This is the other PR I mentioned: https://github.com/JasperFx/wolverine/pull/2574

saithis:
If you prefer it, I could also try to change the PR to make it work like this:

```
.DeclareExchangePassive("some.exchange")
```
Since all other settings of the exchange are ignored anyways, if it is declared passive.

PR #2822, opened 2026-05-15
Prevent `DuplicateIncomingEnvelopeException` on second scheduled retry with RabbitMQ and ProcessInline config 

kakins:
# Problem
Fix `RescheduleExistingEnvelopeForRetryAsync` INSERT/UPDATE semantics in `MessageDatabase<T>`

`RescheduleExistingEnvelopeForRetryAsync` unconditionally called `StoreIncomingAsync` (INSERT). This caused `DuplicateIncomingEnvelopeException` on the second scheduled retry when using RabbitMQ with ProcessInline() listeners and SQL Server persistence.

# Root cause
- On ProcessInline retry #1, no inbox row exists, so INSERT succeeds. 
- The durability agent then redelivers the scheduled envelope. 
- On retry #2, the prior row still exists and the INSERT hits a unique constraint violation.

# Fix
- Attempt UPDATE first. If 0 rows affected (no pre-existing row), fall back to INSERT. 
- This correctly handles all call sites: ProcessInline first retry (no row yet → INSERT), 
- ProcessInline subsequent retries (row exists → UPDATE), and UseDurableInbox (row pre-persisted on arrival → UPDATE).

This aligns `MessageDatabase<T>` with the Oracle implementation, which already uses UPDATE semantics in `RescheduleExistingEnvelopeForRetryAsync`.

Related: #2462 (same root cause, different failure mode — UseDurableInbox fails on retry #1; ProcessInline fails on retry #2).

### DIFF
--- a/src/Persistence/Wolverine.RDBMS/MessageDatabase.Scheduled.cs
+++ b/src/Persistence/Wolverine.RDBMS/MessageDatabase.Scheduled.cs
@@ -20,12 +20,26 @@ public abstract partial class MessageDatabase<T>
             .ExecuteNonQueryAsync(_cancellation);
     }
 
-    public Task RescheduleExistingEnvelopeForRetryAsync(Envelope envelope)
+    public async Task RescheduleExistingEnvelopeForRetryAsync(Envelope envelope)
     {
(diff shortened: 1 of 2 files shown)


kakins:
Closing this for now as @jeremydmiller said he'd probably take this one up.

PR #2837, opened 2026-05-18
fix(rdbms): upsert in RescheduleExistingEnvelopeForRetryAsync (#2823)

### LINKED ISSUE #2823: RabbitMQ `ProcessInline` and `ScheduleRetry` throws `DuplicateIncomingEnvelopeException`
## Bug Description

When using RabbitMQ with `ProcessInline()` listeners, RDBMS persistence (`PersistMessagesWithSqlServer` or equivalent), and a `ScheduleRetry` error policy with RetryCount > 1, the second scheduled retry always fails with `DuplicateIncomingEnvelopeException`.

## Environment

- Wolverine 5.26.0 (confirmed on current main)
- RabbitMQ transport (WolverineFx.RabbitMQ)
- SQL Server persistence (WolverineFx.SqlServer)
- Listeners configured with `ProcessInline()`
- Error policy using `ScheduleRetry(...)` with RetryCount > 1

## Observed Behavior

| RetryCount | Result |
|---|---|
| 1 | Passes |
| 2+ | `DuplicateIncomingEnvelopeException` on second retry |

The exception originates at `Wolverine.RDBMS.MessageDatabase.StoreIncomingAsync`, called from `Wolverine.Runtime.MessageContext.ReScheduleAsync`, called from `Wolverine.Runtime.HandlerPipeline.InvokeAsync`.

## Root Cause

`MessageDatabase<T>.RescheduleExistingEnvelopeForRetryAsync` in `src/Persistence/Wolverine.RDBMS/MessageDatabase.Scheduled.cs` unconditionally calls `StoreIncomingAsync`, which is an INSERT operation.

With `ProcessInline()`, the envelope is never pre-persisted to the inbox before handling. The failure sequence is:

1. Message arrives inline — no inbox row exists
2. Handler throws, `ScheduleRetry` fires, `RescheduleExistingEnvelopeForRetryAsync` is called, INSERT succeeds — row now exists with status=Scheduled
3. Durability agent picks up the scheduled row and redelivers it
4. Handler throws again, `RescheduleExistingEnvelopeForRetryAsync` is called again, INSERT against the same (id, received_at) primary key — duplicate key exception

This is why RetryCount=1 passes (only one INSERT ever occurs) and RetryCount>1 fails (the second INSERT collides with the row left by the first).

**Why this is ProcessInline-specific:** 
- With `UseDurableInbox`, the `IChannelCallback` in `MessageContext` is `DurableReceiver`, which implements `ISupportNativeScheduling`. 
- `MessageContext.tryGetRescheduler` finds it and routes through `DurableReceiver.MoveToScheduledUntilAsync` to `ScheduleExecutionAsync`, which is a pure UPDATE. 
- That path works correctly and never touches `RescheduleExistingEnvelopeForRetryAsync`. 
- With `ProcessInline` the callback is `RabbitMqChannelCallback`, which does not implement `ISupportNativeScheduling`, so all reschedule calls fall through to the INSERT path.

## Proposed Fix

Change `RescheduleExistingEnvelopeForRetryAsync` in `src/Persistence/Wolverine.RDBMS/MessageDatabase.Scheduled.cs` to attempt an UPDATE first
- Only fall back to INSERT if 0 rows were affected. The UPDATE SQL is identical to what `ScheduleExecutionAsync` already uses. 
- If the UPDATE affects 0 rows (no pre-existing row, i.e. ProcessInline retry #1), fall through to `StoreIncomingAsync` as today.
- This aligns `MessageDatabase<T>` with the Oracle implementation in `src/Persistence/Oracle/Wolverine.Oracle/OracleMessageStore.Incoming.cs`, which already uses UPDATE semantics for this method. 

The `Destination`/`received_at` roundtrip has been confirmed safe — `BuildIncomingStorageCommand` writes `envelope.Destination.ToString()` into the `received_at` column, and `ReadIncomingAsync` reads it back as `envelope.Destination = new Uri(rawUri)`, so the UPDATE WHERE clause will always resolve correctly when a row exists.

## Related

#2462 describes the same root cause (`RescheduleExistingEnvelopeForRetryAsync` calling INSERT instead of UPDATE) but in the `UseDurableInbox` configuration where the envelope is pre-persisted on arrival, causing the collision on retry #1 rather than retry #2. This issue covers the remaining `ProcessInline` gap.

jeremydmiller:
## Summary
Closes #2823. Same root cause as #2462 (this fix resolves both).

`MessageDatabase<T>.RescheduleExistingEnvelopeForRetryAsync` in `src/Persistence/Wolverine.RDBMS/MessageDatabase.Scheduled.cs` unconditionally went through `StoreIncomingAsync` (INSERT). The failure mode reported by @kakins:

| Config | Symptom |
|---|---|
| ProcessInline + `ScheduleRetry(RetryCount > 1)` | Retry #1 inserts a Scheduled row; durability agent redelivers; retry #2 collides on `(id, received_at)` PK with `DuplicateIncomingEnvelopeException`. (#2823) |
| UseDurableInbox + `ScheduleRetry` | Inbox row exists from arrival; retry #1 already collides. (#2462) |

Why this is ProcessInline-specific in the #2823 form: with `UseDurableInbox` the `IChannelCallback` is `DurableReceiver`, which implements `ISupportNativeScheduling`, so `MessageContext.tryGetRescheduler` routes through `DurableReceiver.MoveToScheduledUntilAsync` → `ScheduleExecutionAsync` (UPDATE-only). With `ProcessInline` the callback is `RabbitMqChannelCallback` which doesn't implement `ISupportNativeScheduling`, so reschedules fall through to `RescheduleExistingEnvelopeForRetryAsync`'s INSERT path.

The existing comment at `src/Wolverine/Transports/Local/BufferedLocalQueue.cs:80-85` already described this method as an upsert ("misnamed-but-correct API ... it sets Status=Scheduled + OwnerId=AnyNode, then upserts via StoreIncomingAsync") — the upsert behavior was aspirational; this PR makes it real.

## Fix
Try the same UPDATE SQL that `ScheduleExecutionAsync` already uses; fall back to `StoreIncomingAsync` only when 0 rows are affected. That covers:

| Caller / scenario | Pre-fix | Post-fix |
|---|---|---|
| `BufferedLocalQueue.ScheduleAsync` — no existing row | INSERT ✓ | UPDATE 0 → INSERT ✓ |
| ProcessInline retry #1 — no existing row | INSERT ✓ | UPDATE 0 → INSERT ✓ |
| ProcessInline retry #2+ — row exists from #1 | INSERT 💥 | UPDATE ✓ |
| UseDurableInbox retry #1 — row exists from arrival | INSERT 💥 | UPDATE ✓ |

Body bytes are not rewritten on the UPDATE path — `DatabasePersistence.ReadIncomingAsync` (line 96-109) already overrides `Status`, `OwnerId`, `Attempts`, `ScheduledTime` from the column values after deserializing the body, so the UPDATE state is what the runtime sees on the next dequeue.

## Test plan
- [x] New regression `Bug_2823_reschedule_existing_envelope_for_retry.two_consecutive_reschedules_do_not_throw_duplicate` (`src/Persistence/SqlServerTests/Persistence/`) — calls `RescheduleExistingEnvelopeForRetryAsync` twice on the same envelope, asserts no duplicate exception, exactly one row, with `Attempts` and `ScheduledTime` from the second call.
- [x] Verified pre-fix: the new test fails on baseline `main` with `DuplicateIncomingEnvelopeException` at `MessageDatabase.Incoming.cs:167` — the path #2823 reports.
- [x] `dotnet test src/Persistence/SqlServerTests -f net9.0 --filter "FullyQualifiedName~Bug_2823|SqlServerBackedMessageStoreTests"` — 5/5 pass (covers the new test + the existing reschedule round-trip in `SqlServerBackedMessageStoreTests`).
- [x] `dotnet test src/Persistence/PostgresqlTests -f net9.0 --filter "FullyQualifiedName~PostgresqlMessageStoreTests"` — 48/48 pass; fix lives in shared `MessageDatabase<T>` so the Postgres path benefits identically.
- [ ] CI

### DIFF
--- a/src/Persistence/Oracle/OracleTests/Bug_2823_reschedule_existing_envelope_for_retry_oracle.cs
+++ b/src/Persistence/Oracle/OracleTests/Bug_2823_reschedule_existing_envelope_for_retry_oracle.cs
@@ -1,0 +1,64 @@
+using IntegrationTests;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Hosting;
+using Microsoft.Extensions.Logging.Abstractions;
+using Shouldly;
+using Wolverine;
+using Wolverine.ComplianceTests;
+using Wolverine.ComplianceTests.Compliance;
+using Wolverine.Oracle;
+using Wolverine.Persistence.Durability;
+using Wolverine.RDBMS;
+using Wolverine.Transports;
+
+namespace OracleTests;
+
+// Oracle mirror of the regression for https://github.com/JasperFx/wolverine/issues/2823.
+// The Oracle inbox previously did UPDATE-only on retry, which silently no-op'd when
+// no row existed (ProcessInline retry #1), and never matched the same column shape
+// as ScheduleExecutionAsync. RescheduleExistingEnvelopeForRetryAsync is now an upsert
+// aligned with MessageDatabase<T>.
+[Collection("oracle")]
+public class Bug_2823_reschedule_existing_envelope_for_retry_oracle
+{
+    [Fact]
+    public async Task two_consecutive_reschedules_do_not_throw_duplicate()
+    {
+        var dataSource = new OracleDataSource(Servers.OracleConnectionString);
+        var settings = new DatabaseSettings
+        {
+            SchemaName = "WOLVERINE",
+            CommandQueuesEnabled = true,
+            Role = MessageStoreRole.Main
+        };
+        var store = new OracleMessageStore(settings, new DurabilitySettings(), dataSource,
+            NullLogger<OracleMessageStore>.Instance);
+
+        await store.Admin.MigrateAsync();
+        await store.Admin.ClearAllAsync();
+
+        var envelope = ObjectMother.Envelope();
+        envelope.Message = new Message1();
+        envelope.ScheduledTime = DateTimeOffset.UtcNow.AddMinutes(1);
+        envelope.Attempts = 1;
+
+        // Retry #1: no pre-existing row — UPDATE affects 0 rows, INSERT fallback runs.
+        await store.Inbox.RescheduleExistingEnvelopeForRetryAsync(envelope);
+
+        // Retry #2: row from #1 already exists — UPDATE succeeds.
+        envelope.ScheduledTime = DateTimeOffset.UtcNow.AddMinutes(5);
+        envelope.Attempts = 2;
+        await store.Inbox.RescheduleExistingEnvelopeForRetryAsync(envelope);
+
+        var rows = (await store.Admin.AllIncomingAsync())
+            .Where(x => x.Id == envelope.Id)
+            .ToList();
+
+        rows.Count.ShouldBe(1);
+
+        var persisted = rows[0];
+        persisted.Attempts.ShouldBe(2);
+        persisted.ScheduledTime!.Value.ShouldBe(envelope.ScheduledTime.Value, TimeSpan.FromSeconds(2));
+        persisted.OwnerId.ShouldBe(TransportConstants.AnyNode);
+    }
+}
--- a/src/Persistence/Oracle/Wolverine.Oracle/OracleMessageStore.Incoming.cs
+++ b/src/Persistence/Oracle/Wolverine.Oracle/OracleMessageStore.Incoming.cs
@@ -125,16 +125,34 @@ internal partial class OracleMessageStore
 
     public async Task RescheduleExistingEnvelopeForRetryAsync(Envelope envelope)
     {
-        await using var conn = await _dataSource.OpenConnectionAsync(_cancellation);
-        var cmd = conn.CreateCommand(
-            $"UPDATE {SchemaName}.{DatabaseConstants.IncomingTable} SET " +
-            $"{DatabaseConstants.ExecutionTime} = :time, {DatabaseConstants.Attempts} = :attempts " +
-            "WHERE id = :id");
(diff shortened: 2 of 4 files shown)


jeremydmiller:
Follow-up: aligned the Oracle provider (`OracleMessageStore.RescheduleExistingEnvelopeForRetryAsync`) to the same upsert semantics in 1966f8c01.

**Oracle was UPDATE-only and only updated `execution_time` + `attempts`** — silently no-op'd for `ProcessInline` retry #1 (no row exists, UPDATE updates nothing, durability agent has nothing to redeliver) and never matched `ScheduleExecutionAsync`'s column shape (`status`, `owner_id`).

The new behavior mirrors the MessageDatabase<T> fix:

- UPDATE with the full `ScheduleExecutionAsync` shape (`status='Scheduled'`, `owner_id=AnyNode`, `execution_time`, `attempts`) keyed on `id AND received_at`.
- Fall back to `StoreIncomingAsync` when 0 rows are affected.
- Mutate `envelope.Status` / `OwnerId` up front so the body persisted on the INSERT fallback matches the column state.

Added Oracle mirror of the regression test (`Bug_2823_reschedule_existing_envelope_for_retry_oracle`). Verified the test fails on baseline (the second-call assertion `rows.Count.ShouldBe(1)` fails because the prior UPDATE-only path never inserted anything) and passes after the fix. Sibling Oracle test (`can_persist_and_delete_outgoing_envelope`) still green.

PR #2842, opened 2026-05-19
fix(http): default tenant-id assertion to true when detection is configured

### LINKED ISSUE #2841: Missing tenant id silently coerced to the master ("default") tenant across HTTP, bus, and storage
# Missing tenant id silently coerced to the master ("default") tenant across HTTP, bus, and storage

**Severity:** Critical

## Summary

In a multi-tenant Wolverine HTTP application, a request that arrives without a tenant identifier is silently routed to the master/default tenant store instead of being rejected. `MessageBus.TenantId`'s setter coerces a null tenant to `"*DEFAULT*"`, `AssertTenantExists` defaults to `false`, and the persistence layer treats `"*DEFAULT*"`/`"default"`/`"Main"` as the master database. The net effect is that "headerless" multi-tenant traffic — and any in-process bus call where `TenantId` was never set — lands in the master tenant's outbox, saga, and dead-letter storage by default.

## Affected versions

Wolverine 6.0 (`main` as of 2026-05-18).

## Reproduction

1. Bootstrap a multi-tenant Wolverine HTTP app with `opts.TenantId.IsRequestHeaderValue("X-Tenant-Id")` and **without** `opts.TenantId.AssertExists()`.
2. Define an HTTP endpoint that starts a saga whose id is derived from a request field (e.g. `request.CustomerId`).
3. From two different tenant contexts, issue the same `POST /start-saga {"CustomerId":"acme"}` without sending the `X-Tenant-Id` header.
4. Observe that both saga writes are persisted to `Main.<saga_table>` keyed by `"acme"` — the second caller overwrites the first caller's saga row. Cascaded messages from the handler are emitted with `TenantId = "*DEFAULT*"` and may be consumed by handlers that interpret the default tenant as a master/admin scope.

Expected behaviour: once any tenant-detection strategy is registered on a multi-tenanted runtime, a request that arrives without a tenant marker should fail loudly — either rejected by the HTTP layer or thrown by the framework — rather than being silently accepted.

## Affected code

- `src/Wolverine/Runtime/MessageBus.cs:83-87` — the `TenantId` setter rewrites a null assignment to `"*DEFAULT*"`.
- `src/Http/Wolverine.Http/Runtime/MultiTenancy/TenantIdDetection.cs:63-72` — `AssertTenantExists` defaults to `false`.
- `src/Http/Wolverine.Http/Runtime/MultiTenancy/DetectTenantIdFrame.cs:42-48` — assertion is guarded by `ShouldAssertTenantIdExists`, so it never fires under the default.
- `src/Http/Wolverine.Http/CodeGen/MessageBusSource.cs:43-47` — codegen unconditionally assigns `ctx.TenantId = tenantId;` even when `tenantId` is null.
- `src/Wolverine/Persistence/Durability/MultiTenantedMessageStore.cs:634-665` — `GetDatabaseAsync` short-circuits empty/`"*DEFAULT*"`/`"default"`/`"Main"` tenant ids to the `Main` database.
- `src/Wolverine/Persistence/Durability/MultiTenantedMessageStore.cs:594-613` — `EnrollAndFetchSagaStorage` routes default-tenant context to `Main`'s saga storage.

## Why this matters

In a multi-tenant deployment that configures tenant detection (request header, query string, subdomain, claim, or route argument) but does not call `opts.TenantId.AssertExists()` — and leaves at least one endpoint with `TenancyMode` null or `Maybe` — a request that omits the tenant marker reaches the framework with `tenantId == null`. The bus rewrites that to `"*DEFAULT*"`, persistence treats it as the master tenant, and every side effect (durable outbox row, scheduled-message row, saga row, dead-letter row, cascaded message) lands in `Main`. Two unauthenticated callers omitting the tenant header who address the same saga identity (e.g. a saga keyed off a request field) collide on a single row in `Main` and can read or overwrite each other's saga state. The operator is required to opt in to safe-by-default behaviour rather than opt out of unsafe behaviour — the per-tenant database isolation the framework otherwise enforces is silently defeated for any caller that forgets a header.

## Suggested fix

- In `TenantIdDetection`, default `AssertTenantExists` to `true` whenever any tenant-detection strategy is registered and the runtime has multi-tenanted message stores configured.
- Alternatively (or additionally), have `DetectTenantIdFrame` / `MessageBusSource` codegen route through the assertion path even when `AssertTenantExists` is false, so that a null tenant under a multi-tenanted runtime fails loudly rather than coercing.
- Update the tenant-detection documentation to flag `AssertExists()` as required, not optional, for any multi-tenant HTTP surface.

The work is scoped to the tenant-detection configuration object and the HTTP codegen frame; it does not need to alter the `MessageBus.TenantId` setter's coercion behaviour to land. Backward compatibility: applications that today rely on falling through to the master tenant for "headerless" requests will need to either explicitly mark those routes as `TenancyMode.None`/`Maybe` or add a default header.

BlackChepo:
# fix(http): default tenant-id assertion to true when detection is configured

Closes #2841.

## Summary

Configuring tenant detection without calling `AssertExists()` previously let header-omitted requests fall through to the master ("default") tenant store, silently writing per-tenant outbox/saga/dead-letter rows into `Main`. The default in 6.0 is now to return `400 ProblemDetails` on a missing tenant id whenever any detection strategy is registered.

Opt-outs:

- **`DefaultIs(...)`** — register a fallback tenant for header-omitted requests (preferred when you actually want a fallback).
- **`DoNotAssertExists()`** — explicitly restore the prior fall-through behaviour. Documented as a footgun; reserved for non-multi-tenant scenarios.

The fix is scoped to `TenantIdDetection` and the HTTP options surface. `MessageBus.TenantId`'s null→`"*DEFAULT*"` coercion and the persistence-layer short-circuits on `"*DEFAULT*"`/`"default"`/`"Main"` are unchanged — the loud failure now happens earlier, before a null tenant reaches them.

## Behaviour

`TenantIdDetection.EffectiveAssertTenantExists` resolves like this:

1. If `AssertExists()` or `DoNotAssertExists()` was called explicitly, honour that (last-write-wins).
2. Otherwise, if no detection strategies are registered → `false` (single-tenant app, nothing to assert).
3. Otherwise, if a `FallbackDefault` strategy is registered (via `DefaultIs(...)`) → `false` (the fallback handles missing ids).
4. Otherwise → `true` (multi-tenant detection is configured, fail loud on missing id).

## Breaking change (6.0)

The public `AssertTenantExists` setter on `TenantIdDetection` is removed. Callers must use `AssertExists()` / `DoNotAssertExists()` instead. The 5.x to 6.0 migration note in `docs/guide/http/multi-tenancy.md` covers this.

## Files changed

- `src/Http/Wolverine.Http/Runtime/MultiTenancy/TenantIdDetection.cs` — replace public `AssertTenantExists` property with nullable `_assertSetting` + `EffectiveAssertTenantExists`; add `DoNotAssertExists()`.
- `src/Http/Wolverine.Http/WolverineHttpOptions.cs` — add `DoNotAssertExists()` to `ITenantDetectionPolicies` with a footgun-warning xmldoc.
- `docs/guide/http/multi-tenancy.md` — document the new default, `DefaultIs(...)` opt-out, `DoNotAssertExists()` opt-out, and 5.x → 6.0 migration note.
- `src/Testing/Wolverine.Http.Tests/Samples/MultiTenancy.cs` — sample updates.
- `src/Testing/Wolverine.Http.Tests/multi_tenancy_detection_and_integration.cs` — new tests covering the four resolution branches above.

### DIFF
--- a/docs/guide/http/multi-tenancy.md
+++ b/docs/guide/http/multi-tenancy.md
@@ -246,24 +246,35 @@ See [Referencing the TenantId](/guide/handlers/multi-tenancy.html#referencing-th
 
 ## Requiring Tenant Id -- or Not!
 
-You can direct Wolverine.HTTP to verify that there is a non-null, non-empty tenant id on all requests with this syntax:
+Starting in Wolverine 6.0, the tenant id is **required by default** on any HTTP endpoint that participates in multi-tenancy whenever at least one detection strategy is registered (`IsRequestHeaderValue`, `IsQueryStringValue`, `IsClaimTypeNamed`, `IsRouteArgumentNamed`, `IsSubDomainName`, or `DetectWith(...)`). A request that omits the tenant id returns a 400 with a [ProblemDetails](https://learn.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.mvc.problemdetails?view=aspnetcore-7.0) body stating that the tenant id was missing.
+
+The `AssertExists()` call is therefore now optional. It is still valid and documents intent:
 
 <!-- snippet: sample_assert_tenant_id_exists -->
-<a id='snippet-sample_assert_tenant_id_exists'></a>
-```cs
-app.MapWolverineEndpoints(opts =>
-{
-    // Configure your tenant id detection...
-
-    // Require tenant id some how, some way...
-    opts.TenantId.AssertExists();
-});
-```
-<sup><a href='https://github.com/JasperFx/wolverine/blob/main/src/Http/Wolverine.Http.Tests/Samples/MultiTenancy.cs#L67-L76' title='Snippet source file'>snippet source</a> | <a href='#snippet-sample_assert_tenant_id_exists' title='Start of snippet'>anchor</a></sup>
 <!-- endSnippet -->
 
-At runtime, this is going to return a status code of 400 with a [ProblemDetails](https://learn.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.mvc.problemdetails?view=aspnetcore-7.0) specification 
-response stating that the tenant id was missing. 
+There are two ways to opt out of this default:
+
+**Preferred — register a fallback tenant.** If your application has a meaningful default tenant for unauthenticated or header-omitted traffic, register it with `DefaultIs(...)`. When a `DefaultIs(...)` strategy is registered the framework treats it as your explicit fallback and does not assert:
+
+```csharp
+opts.TenantId.IsRequestHeaderValue("X-Tenant-Id");
+opts.TenantId.DefaultIs("public");
+```
+
+**Discouraged — disable the assertion.** For non-multi-tenant scenarios (or to restore the 5.x behavior of silently coercing missing tenant ids to the master tenant) call `DoNotAssertExists()`. This is a footgun in multi-tenant deployments because cascaded messages and downstream side effects will execute against the master tenant store:
+
+<!-- snippet: sample_do_not_assert_tenant_id_exists -->
+<!-- endSnippet -->
+
+### Migrating from 5.x
+
+In 5.x, `AssertExists()` was opt-in and missing tenant ids were silently coerced to the master tenant store. In 6.0 the default is reversed. If your 5.x deployment relied on the silent coercion, either:
+
+1. Register `opts.TenantId.DefaultIs(...)` with the tenant id you want missing-id traffic to land on, or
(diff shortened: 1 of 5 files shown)


jeremydmiller:
@BlackChepo Nuh uh, this is a change in designed behavior. The approach from the beginning has been that you need to explicitly state you want the tenant id to be required, and some folks absolutely want the tenant id to be optional sometimes.

BlackChepo:
Thanks for the quick review, @jeremydmiller, and for the context on the original design intent, that makes sense.

PR #2844, opened 2026-05-19
Bound attacker-controlled length fields in EnvelopeSerializer and TCP transport

### LINKED ISSUE #2843: EnvelopeSerializer and TCP WireProtocol allocate attacker-controlled lengths — tiny-payload OOM
# `EnvelopeSerializer` and TCP `WireProtocol` allocate attacker-controlled lengths — tiny-payload OOM

## Summary

Three deserialize sinks read a 32-bit signed length directly off the wire and immediately allocate against that value with no bound check. `EnvelopeSerializer.ReadMany` reads `numberOfMessages` and does `new Envelope[numberOfMessages]` before reading any element. `readSingle` and `ReadEnvelopeData` pass the per-envelope `byteCount` straight into `BinaryReader.ReadBytes(byteCount)`, which allocates `new byte[byteCount]` up front. The TCP listener in `WireProtocol.ReceiveAsync` reads a 4-byte outer length prefix and immediately calls `stream.ReadBytesAsync(length)` before the envelope-level reader sees a byte. In every case the trigger is a 4-byte attacker-controlled field, so a single tiny request (4–17 bytes) can force a multi-gigabyte allocation and crash the process with `OutOfMemoryException`.

## Affected versions

- 6.0 / `main` — verified as of 2026-05-19.
- The TCP `WireProtocol.ReceiveAsync` path also exists in 5.x.

## Reproduction

All three scenarios require only that the corresponding listener is wired up. The HTTP transport endpoints are anonymous by default.

**1. HTTP batch — ~17 GiB allocation request from a 4-byte body**

Bootstrap with `app.MapWolverineHttpTransportEndpoints();`. Send a 4-byte body to `POST /_wolverine/batch/{queue}` with `Content-Type: binary/wolverine-envelope-batch` containing `0x7F 0xFF 0xFF 0xFF`. `ExecuteBatchAsync` reads the body and calls `EnvelopeSerializer.ReadMany(data)`, which immediately attempts `new Envelope[2_147_483_647]`.

**2. HTTP invoke — ~2 GiB allocation request from a ~17-byte body**

Send a ~17-byte body to `POST /_wolverine/invoke` shaped as 8 bytes `SentAt` + 4 bytes `headerCount=0` + 4 bytes `byteCount=0x7FFFFFFF`. `InvokeAsync` calls `EnvelopeSerializer.Deserialize`, the call path reaches `readSingle` → `BinaryReader.ReadBytes(int.MaxValue)` → `new byte[int.MaxValue]`.

**3. TCP listener — ~2 GiB allocation request from a 4-byte frame prefix**

Connect to any configured TCP listener and write `0x7F 0xFF 0xFF 0xFF` as the first four bytes. `WireProtocol.ReceiveAsync` reads the prefix into `length`, then `stream.ReadBytesAsync(length)` triggers a `new byte[int.MaxValue]` allocation before `EnvelopeSerializer.ReadMany` is ever called — i.e. the outer transport read happens before any envelope-level guard could apply.

In each case the process crashes with `OutOfMemoryException` after a single request.

## Affected code

- `src/Wolverine/Runtime/Serialization/EnvelopeSerializer.cs:151-163` — `ReadMany` reads `numberOfMessages` as a signed `int32` (line 155) and allocates `new Envelope[numberOfMessages]` (line 156) before reading any element.
- `src/Wolverine/Runtime/Serialization/EnvelopeSerializer.cs:188-203` — `readSingle` reads a per-envelope `byteCount` and passes it directly to `BinaryReader.ReadBytes(byteCount)` at line 203.
- `src/Wolverine/Runtime/Serialization/EnvelopeSerializer.cs:172-185` — `ReadEnvelopeData` does the same on the standalone envelope-data path at line 185.
- `src/Http/Wolverine.Http/Transport/HttpTransportExecutor.cs:27-51` — `/_wolverine/batch/{queue}` reads the whole body (line 41) and calls `EnvelopeSerializer.ReadMany(data)` (line 47).
- `src/Http/Wolverine.Http/Transport/HttpTransportExecutor.cs:72-90` — `/_wolverine/invoke` reads the whole body and calls `EnvelopeSerializer.Deserialize` (line 90).
- `src/Wolverine/Transports/Tcp/WireProtocol.cs:64-77` — `ReceiveAsync` reads a 4-byte length prefix (lines 70-71) and calls `stream.ReadBytesAsync(length)` (line 77) before the envelope-level reader can see any bytes.

## Why this matters

The attack is parametric in a 4-byte length field, not in body length. A 4–17 byte request triggers a multi-gigabyte allocation, so request-size limits (Kestrel's `MaxRequestBodySize`, gateway caps, etc.) do not defend against it — the body is well under any sensible cap. The OOM is unrecoverable in practice: the runtime cannot service a multi-GiB allocation request, and the exception only bubbles after the attempt has already aborted the process. The HTTP transport endpoints are anonymous by default, so the precondition for the HTTP variants is "the application has mapped the transport endpoints and is reachable." The TCP precondition is "a TCP listener is configured and reachable."

## Notes

A PR will follow that bounds all three sinks plus the outer TCP frame, exposes configurable ceilings on `WolverineOptions` (max envelopes per batch, max envelope data size, max incoming TCP frame size), and returns HTTP 400 from the transport endpoints when a malformed envelope is received instead of letting the deserializer crash the request. The patch is local to the deserialize side and to one TCP-receive call site; the serialize path is not touched.

BlackChepo:
# Bound attacker-controlled length fields in `EnvelopeSerializer` and TCP transport

Closes #2843.

## Summary

Three deserialize sinks in `EnvelopeSerializer` (`ReadMany`, `readSingle`, `ReadEnvelopeData`) and the outer TCP-frame read in `WireProtocol.ReceiveAsync` each took a 32-bit signed length straight off the wire and allocated against it with no bound check, so a 4–17 byte payload could request a multi-GiB allocation and crash the host with `OutOfMemoryException`. This PR validates each of those length fields against a configurable ceiling before the allocation, throws `InvalidEnvelopeException` on violation, exposes the ceilings on `WolverineOptions`, and maps malformed transport input to HTTP 400 / the existing TCP serialization-failure response instead of letting the deserializer crash the request. The serialize path is unchanged.

## What this PR changes

**Core guards in `EnvelopeSerializer`** (`src/Wolverine/Runtime/Serialization/EnvelopeSerializer.cs`)

Validates `numberOfMessages` (batch count), per-envelope `byteCount` (data length), and per-envelope `headerCount` against configurable caps before any size-dependent allocation. Cross-checks `numberOfMessages` against `data.Length / minBytesPerEnvelope` so a forged-but-under-cap value cannot allocate beyond the buffer that was actually received; cross-checks `byteCount` against bytes remaining in the buffer. Violations throw `InvalidEnvelopeException` rather than letting `new Envelope[…]` / `new byte[…]` fail with OOM. The active caps live on a single struct `EnvelopeReaderLimits` exposed through `EnvelopeSerializer.Limits`.

**Configurable limits on `WolverineOptions`** (`src/Wolverine/WolverineOptions.Serialization.cs`)

Four new properties:

- `MaxIncomingEnvelopeBatchSize` — default `1000` envelopes per batch
- `MaxIncomingEnvelopeDataSize` — default `4 * 1024 * 1024` (4 MiB) per envelope data segment
- `MaxIncomingEnvelopeHeaderCount` — default `128` headers per envelope
- `MaxIncomingTcpFrameSize` — default `32 * 1024 * 1024` (32 MiB) per outer TCP frame

`WolverineRuntime.HostService.StartAsync` publishes the configured values into `EnvelopeSerializer.Limits` and `WireProtocol.MaxFrameSize` after async extensions have had a chance to mutate options and before any listener is activated, so high-throughput batch consumers can raise the caps without touching the statics directly. The statics are process-global; the property docs call out that in a multi-host process the last host to start wins.

**HTTP 400 for malformed envelopes** (`src/Http/Wolverine.Http/Transport/HttpTransportExecutor.cs`)

`ExecuteBatchAsync` and `InvokeAsync` previously caught every deserialize exception and returned 500. Malformed payloads from external callers are client errors, so `InvalidEnvelopeException` now maps to 400 with a sanitized `"Invalid envelope: <reason>"` body and is logged at Warning. Genuine server-side deserialization failures continue to log at Error and return 500. The previously unwrapped `EnvelopeSerializer.Deserialize` call in `InvokeAsync` is now inside the same try/catch so it cannot escape into the ASP.NET pipeline.

**TCP outer-frame guard** (`src/Wolverine/Transports/Tcp/WireProtocol.cs`)

`ReceiveAsync` reads its 4-byte length prefix before `EnvelopeSerializer` ever sees the bytes, so the central guards above cannot protect it. The wire-supplied length is now validated against `WireProtocol.MaxFrameSize` (sourced from `WolverineOptions.MaxIncomingTcpFrameSize`) before `stream.ReadBytesAsync(length)`; out-of-range values throw `InvalidEnvelopeException`, which the existing `catch` in `ReceiveAsync` handles by returning the standard `SerializationFailureBuffer` to the sender — no protocol change, no new client-visible failure code. A new `WireProtocol.DefaultMaxFrameSize` const keeps the option's initializer and the static's default in sync so they can't drift.

**Test hardening**

Tests that touch the process-global `EnvelopeSerializer.Limits` and `WireProtocol.MaxFrameSize` are pinned into a single non-parallel xUnit collection (`EnvelopeSerializerLimitsCollection`) so they cannot race each other under default cross-class parallelism. `EnvelopeSerializerGuardTests` implements `IDisposable` to restore `Limits` after each test so a future test that lowers the caps cannot leak state to its neighbours in the collection.

## Configuration example

```csharp
builder.UseWolverine(opts =>
{
    // Raise caps for a high-throughput batch consumer
    opts.MaxIncomingEnvelopeBatchSize = 10_000;
    opts.MaxIncomingEnvelopeDataSize = 16 * 1024 * 1024; // 16 MiB
    opts.MaxIncomingTcpFrameSize     = 256 * 1024 * 1024; // 256 MiB
});
```

## Behaviour changes (potentially user-visible)

- Inbound envelope batches with `numberOfMessages > MaxIncomingEnvelopeBatchSize`, envelope data segments larger than `MaxIncomingEnvelopeDataSize`, or envelopes with more than `MaxIncomingEnvelopeHeaderCount` headers are now rejected with `InvalidEnvelopeException` rather than being deserialized. Existing applications that exchanged envelopes above these defaults need to raise the corresponding option.
- The HTTP transport endpoints `POST /_wolverine/batch/{queue}` and `POST /_wolverine/invoke` now return `400 Bad Request` (instead of `500 Internal Server Error`) when the body is not a valid envelope.
- TCP frames whose 4-byte length prefix exceeds `MaxIncomingTcpFrameSize` (default 32 MiB) cause the listener to return the standard `FailDesr` response to the sender instead of crashing. Hosts that previously relied on accepting larger frames need to raise the option.

## Out of scope

- The serialize side of `EnvelopeSerializer` is intentionally untouched — sending hosts continue to write whatever shape they did before, so the change is wire-compatible.
- No transport protocol bump and no new client-visible failure codes; existing senders see the existing failure responses.

### DIFF
--- a/src/Http/Wolverine.Http.Tests/Transport/HttpTransportExecutorTests.cs
+++ b/src/Http/Wolverine.Http.Tests/Transport/HttpTransportExecutorTests.cs
@@ -53,6 +53,55 @@ public class HttpTransportExecutorTests : IntegrationContext
     }
 
     [Fact]
+    public async Task execute_batch_returns_400_for_oversize_envelope_batch_claim()
+    {
+        // Wire-claim of int.MaxValue envelopes in a 4-byte payload — well over
+        // the configured MaxIncomingEnvelopeBatchSize cap.
+        var body = BitConverter.GetBytes(int.MaxValue);
+
+        await Scenario(s =>
+        {
+            s.Post.ByteArray(body).ToUrl("/_wolverine/batch/test")
+                .ContentType(HttpTransport.EnvelopeBatchContentType);
+            s.StatusCodeShouldBe(400);
+        });
+    }
+
+    [Fact]
+    public async Task invoke_returns_400_for_oversize_envelope_data_claim()
+    {
+        // Construct a minimal binary envelope that claims int.MaxValue bytes of
+        // payload data — well over the configured MaxIncomingEnvelopeDataSize cap.
+        // Layout: SentAt (int64) + headerCount=0 (int32) + byteCount=int.MaxValue (int32).
+        using var ms = new MemoryStream();
+        using (var bw = new BinaryWriter(ms, System.Text.Encoding.UTF8, leaveOpen: true))
+        {
+            bw.Write(DateTime.UtcNow.ToBinary());
+            bw.Write(0);            // headerCount
+            bw.Write(int.MaxValue); // byteCount
+        }
+        var body = ms.ToArray();
+
+        await Scenario(s =>
+        {
+            s.Post.ByteArray(body).ToUrl("/_wolverine/invoke")
+                .ContentType(HttpTransport.EnvelopeContentType);
+            s.StatusCodeShouldBe(400);
+        });
+    }
+
+    [Fact]
+    public async Task invoke_returns_500_for_invalid_envelope_data()
+    {
+        await Scenario(s =>
+        {
+            s.Post.ByteArray(new byte[] { 0xFF, 0xFF, 0xFF }).ToUrl("/_wolverine/invoke")
+                .ContentType(HttpTransport.EnvelopeContentType);
+            s.StatusCodeShouldBe(500);
+        });
+    }
+
+    [Fact]
     public async Task invoke_returns_415_for_wrong_content_type()
     {
         var result = await Scenario(s =>
--- a/src/Http/Wolverine.Http/Transport/HttpTransportExecutor.cs
+++ b/src/Http/Wolverine.Http/Transport/HttpTransportExecutor.cs
@@ -46,6 +46,11 @@ internal class HttpTransportExecutor
         {
             envelopes = EnvelopeSerializer.ReadMany(data);
         }
+        catch (InvalidEnvelopeException e)
+        {
+            _logger.LogWarning(e, "Rejected malformed envelope batch in Http Transport ExecuteBatchAsync()");
+            return Results.Problem($"Invalid envelope: {e.Message}", statusCode: 400);
+        }
         catch (Exception e)
         {
             _logger.LogError(e, "Error trying to read Envelope[] in the Http Transport ExecuteBatchAsync()");
@@ -87,7 +92,21 @@ internal class HttpTransportExecutor
         Envelope? envelope = null;
         if (values[0] == HttpTransport.EnvelopeContentType)
         {
-            var data = await httpContext.Request.Body.ReadAllBytesAsync(); envelope = EnvelopeSerializer.Deserialize(data);
+            var data = await httpContext.Request.Body.ReadAllBytesAsync();
+            try
+            {
+                envelope = EnvelopeSerializer.Deserialize(data);
+            }
+            catch (InvalidEnvelopeException e)
+            {
+                _logger.LogWarning(e, "Rejected malformed envelope in Http Transport InvokeAsync()");
+                return Results.Problem($"Invalid envelope: {e.Message}", statusCode: 400);
+            }
+            catch (Exception e)
+            {
+                _logger.LogError(e, "Error trying to deserialize envelope in Http Transport InvokeAsync()");
+                return Results.Problem("Error trying to deserialize envelope", statusCode: 500);
+            }
         }
         else
         {
--- a/src/Testing/CoreTests/Serialization/EnvelopeSerializerGuardTests.cs
+++ b/src/Testing/CoreTests/Serialization/EnvelopeSerializerGuardTests.cs
@@ -5,6 +5,7 @@ using Xunit;
 
 namespace CoreTests.Serialization;
 
+[Collection("EnvelopeSerializerLimits")]
 public class EnvelopeSerializerGuardTests
 {
     [Fact]
--- a/src/Testing/CoreTests/Serialization/EnvelopeSerializerGuardTests.cs
+++ b/src/Testing/CoreTests/Serialization/EnvelopeSerializerGuardTests.cs
@@ -1,0 +1,99 @@
+using System.IO;
+using Shouldly;
+using Wolverine.Runtime.Serialization;
+using Xunit;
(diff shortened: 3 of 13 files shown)


jeremydmiller:
@BlackChepo I'm too lazy to check, but I'd bet that WireProtocol hasn't changed in about 5 years.

This will be in 6.0 unless you want it in 5.* soon

BlackChepo:
Thanks @jeremydmiller, 6.0 is perfectly fine, no backport to 5.x needed. This is a hardening fix rather than an urgent hotfix, so shipping it with 6.0 works for me.

PR #2873, opened 2026-05-20
Polecat 4.0.0-rc.1 + cut 6.0.0-rc.1 + fix EF Core saga concurrency test (follow-up to #2872)

jeremydmiller:
## Why

PR #2872 was merged at its **first commit only**, before three follow-up fixes landed. This PR carries the stranded work and fixes `main`'s currently-red `efcore` CI.

## Contents

1. **Polecat `4.0.0-alpha.10` → `4.0.0-rc.1`** + version cut to **`6.0.0-rc.1`** (first 6.0 release candidate; changelog entry added in `Directory.Build.props`). Polecat rc.1 carries the same constant relocation as the rest of the stack — `Polecat.Tenancy.DefaultTenantId` is gone, so both `Wolverine.Polecat` subscription relays now use `JasperFx.StorageConstants.DefaultTenantId`.

2. **Fix `EfCoreTests.Optimistic_concurrency_with_ef_core`** (currently failing on `main`). The test set the concurrency-token `OriginalValue` to a boxed `long` (`999L`); EF Core unboxes against the property's CLR type, which is `int` again after the permanent `Saga.Version` revert, so the `long` failed to unbox. Reverted to `999`. Verified passing locally against SQL Server.

3. **Migration-docs correction.** `docs/guide/migration.md` documented a `Saga.Version` int→long widening and told users to drop `new int Version` shadows / make wire-format changes. `Saga.Version` is `int` in both 5.x and 6.0 — no type change — so the table row and section were removed and the stale `RevisionedSaga.cs` comment fixed.

## ⚠️ Known blocker (NOT fixed here — lives in Marten)

The `marten` CI job fails on a **Marten 9.0.0-rc.1 regression**, independent of this PR:

```
System.ArgumentNullException: Value cannot be null. (Parameter 'distributor')
  at JasperFx.Events.Daemon.ProjectionCoordinatorBase..ctor(IProjectionDistributor distributor, ...)
  at Marten.Events.Daemon.Coordination.ProjectionCoordinator..ctor(DocumentStore store, ILogger logger)
  at MartenTests.AncillaryStores.bootstrapping_ancillary_marten_stores_with_wolverine.projection_coordinator_from_marten()
```

Resolving `IProjectionCoordinator<TAncillaryStore>` constructs Marten's `ProjectionCoordinator` with a null `IProjectionDistributor`. Deterministic (fails both retries), and `9.0.0-rc.1` is the newest Marten rc, so there's no published fix. **This blocks the 6.0.0-rc.1 NuGet publish** until Marten ships a fix.

Full `wolverine.slnx` builds clean (0/0). `efcore` goes green with this PR; `marten` stays red on the dependency bug above.

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -61,8 +61,21 @@
                      IEventSlice<T>.Aggregate -> IEventSlice<T>.Snapshot
                      ProjectionBase.ProjectionName -> Name
                      ProjectionBase.CacheLimitPerTenant -> Options.CacheLimitPerTenant
+          rc.1:    First 6.0 release candidate. Critter Stack 2026 rc pin matrix
+                   (PR #2872):
+                     JasperFx / .RuntimeCompiler / .SourceGeneration / .Events
+                       (+ .Events.SourceGenerator)     alpha.* -> 2.0.0-rc.2 / 5.0.0-rc.2
+                     Marten / .AspNetCore / .Newtonsoft alpha.4 -> 9.0.0-rc.1
+                     Weasel.* (7 packages)             alpha.7 -> 9.0.0-rc.1
+                     Polecat                           alpha.10 -> 4.0.0-rc.1
+                   JasperFx 2.0 rc split numeric versioning into IRevisioned.Version
+                   (int) and ILongVersioned.Version (long); Saga.Version reverts to
+                   int (permanent) so sagas implement IRevisioned directly, and
+                   event-stream-versioned aggregates implement ILongVersioned. rc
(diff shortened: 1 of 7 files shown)


jeremydmiller:
The `marten` CI red on this PR is the Marten 9.0.0-rc.1 `ProjectionCoordinator` null-distributor regression (ancillary stores). Fix is up at JasperFx/marten#4535 — needs a Marten 9.0.0-rc.2 before this can re-pin and the 6.0.0-rc.1 publish can go out.

PR #2920, opened 2026-05-27
Flush EF Core outbox before writing the HTTP response (GH-2917)

### LINKED ISSUE #2917: TrackActivity with Transactional Outbox doesn't work with Wolverine.Http
### Discussed in https://github.com/JasperFx/wolverine/discussions/2916

<div type='discussions-op-text'>

<sup>Originally posted by **LodewijkSioen** May 27, 2026</sup>
**Describe the bug**
When using Wolverine.Http and transactions, the response is closed before the outbox is flushed. Alba stops tracking the session as soon as the response arrives, so the TrackedSession is empty.
I already tried to raise this issue in #1750, but with Wolverine 6 it actually got worse. Previously I got the `Sent` event in the tracked session, but now event that event is missing.

**To Reproduce**
See this unit test: https://github.com/LodewijkSioen/wolverine/blob/c68bf5fc6ac3b4ffa6729b0fa787a9ac02fa8d67/src/Persistence/EfCoreTests/trackactivity_outbox_transactional.cs

**Expected behavior**
Some way to track these kind of Activity in Wolverine.Http

**Workaround**
`Thread.Sleep(100)` ugh

**Additional context**
The generated code looks a bit like this. The problem is that `WriteJsonAsync` will close the TrackedSession on the Alba side of things while it's `FlushOutgoingMessagesAsync` that starts the tracking of the OutgoingMessages.

Strangely enough, if you're not using transactions, the tracking of the message starts in `EnqueueCascadingAsync` and everything works as intended.

```csharp
// The actual HTTP request handler execution
(var acceptResponse_response, var outgoingMessages) = await executeDeviceActionEndpoint.ExecuteDeviceAction(action, device, locationId, ducoSystemService).ConfigureAwait(false);


// Outgoing, cascaded message
await messageContext.EnqueueCascadingAsync(outgoingMessages).ConfigureAwait(false);

// This response type customizes the HTTP response
ApplyHttpAware(acceptResponse_response, httpContext);

// Added by EF Core Transaction Middleware
var result_of_SaveChangesAsync = await userAppDbContext.SaveChangesAsync(httpContext.RequestAborted).ConfigureAwait(false);

// Writing the response body to JSON because this was the first 'return variable' in the method signature
await WriteJsonAsync(httpContext, acceptResponse_response);

// Have to flush outgoing messages just in case Marten did nothing because of https://github.com/JasperFx/wolverine/issues/536
await messageContext.FlushOutgoingMessagesAsync().ConfigureAwait(false);

await efCoreEnvelopeTransaction.CommitAsync(httpContext.RequestAborted).ConfigureAwait(false);
```</div>

jeremydmiller:
Closes #2917.

## Problem

With Wolverine.Http + EF Core's transactional outbox, a transactional endpoint that returns cascading messages would leave an `IHost.TrackActivity().ExecuteAndWaitAsync(...)` session **empty** (no `Sent`/`MessageSucceeded` records) unless the caller added an explicit `WaitForMessageToBeReceivedAt` condition.

**Root cause (deterministic ordering, not a flaky race):** EF Core's Eager transaction middleware `EnrollDbContextInTransaction` wraps the rest of the chain in `try { … }` and emits `efCoreEnvelopeTransaction.CommitAsync()` at the **end** of the try block. `CommitAsync` commits the DB transaction and **then** flushes the outbox (sends the cascading messages). But the response-writing postprocessor (`WriteJsonAsync`) is *inside* that try block, **before** `CommitAsync`. So the generated endpoint wrote the HTTP response → the Alba client received it and `ExecuteAndWaitAsync` saw no in-flight activity and returned → *then* the outbox was flushed. Tracking missed the send. (Marten is unaffected because it flushes via a `SaveChanges` session listener, before the response.)

## Fix

Reorder so the transaction commit + outbox flush run **before** the response is written:

- `EnrollDbContextInTransaction` no longer emits `CommitAsync()` at the end of its wrap — it just begins the transaction and provides the `try/catch` rollback wrapper.
- A new `CommitEfCoreEnvelopeTransaction` postprocessor performs the commit + flush. `EFCorePersistenceFrameProvider` adds it **right after `SaveChanges`**, which is before the response writer Wolverine.Http appends during codegen. The generated try block now reads:

  ```
  try {
      handler;
      SaveChanges;
      commit + flush;     // CommitEfCoreEnvelopeTransaction (was at end of wrap)
      WriteResponse;       // now after commit + flush
  } catch { rollback; throw; }
  ```

This preserves existing guarantees:
- The flush still runs **after** the EF Core commit → does **not** reintroduce the flush-before-commit row-stranding bug (`Bug_efcore_outbox_flush_before_commit` stays green).
- A failed commit now rolls back and **never writes a success response** — strictly more correct than before (previously the client got `200` before the transaction committed). Rollback-after-commit is a no-op because EF Core clears `CurrentTransaction` on commit.

**Scope:** the non-multi-tenant Eager path (`EnrollDbContextInTransaction`). Lightweight mode (standalone `FlushOutgoingMessages`) and message-handler chains are unaffected. The multi-tenant `StartDatabaseTransactionForDbContext` path still has the old ordering — a follow-up commit will extend the same reordering there.

## Validation

- New regression test `issue_2917_trackactivity_transactional_outbox` passes deterministically (no explicit wait condition).
- `Bug_efcore_outbox_flush_before_commit` still passes (no flush-before-commit regression).
- Full **`Wolverine.Http.Tests`: 771 passed, 10 skipped, 0 failed** (EF Core + Marten HTTP).
- `EfCoreTests`, `EfCoreTests.MultiTenancy`, `PersistenceTests` compile clean.

> The SQL Server-backed EF Core **message-handler** suites couldn't be run locally (emulated SQL Server too slow) — relying on CI here. The change is functionally equivalent for handlers (no response writer, so the commit postprocessor remains the last frame).

### DIFF
--- a/src/Http/Wolverine.Http.Tests/issue_2917_trackactivity_transactional_outbox.cs
+++ b/src/Http/Wolverine.Http.Tests/issue_2917_trackactivity_transactional_outbox.cs
@@ -7,12 +7,14 @@ using Xunit.Abstractions;
 
 namespace Wolverine.Http.Tests;
 
-// Reproduction for https://github.com/JasperFx/wolverine/issues/2917.
-// /issue2917 is [Transactional] (Marten outbox) and returns a cascading Issue2917Message.
-// Under the outbox that message is only sent in FlushOutgoingMessagesAsync, which the generated
-// endpoint runs AFTER WriteJsonAsync. With Alba the client receives the response before the flush
-// records the "sent" envelope, so a TrackActivity session that relies on the default
-// "wait until activity quiesces" behavior can finish empty.
+// Regression test for https://github.com/JasperFx/wolverine/issues/2917.
+// /issue2917 is an EF Core [Transactional] endpoint that returns a cascading Issue2917Message.
+// EF Core's Eager transaction middleware now commits the transaction and flushes the outbox (via
+// the CommitEfCoreEnvelopeTransaction postprocessor) BEFORE the HTTP response is written. Previously
+// the flush happened inside EnrollDbContextInTransaction.CommitAsync at the end of the wrap, AFTER
+// WriteJsonAsync wrote the response - so with Alba the client received the response before the
+// "sent" envelope was recorded and a TrackActivity session (with no explicit wait condition) could
+// finish empty.
 public class issue_2917_trackactivity_transactional_outbox : IntegrationContext
 {
     private readonly ITestOutputHelper _output;
@@ -23,7 +25,7 @@ public class issue_2917_trackactivity_transactional_outbox : IntegrationContext
     }
 
     [Fact]
-    public async Task tracked_session_should_capture_the_transactional_outbox_message()
+    public async Task tracked_session_captures_the_transactional_outbox_message()
     {
         var tracked = await Host
             .TrackActivity()
@@ -35,12 +37,10 @@ public class issue_2917_trackactivity_transactional_outbox : IntegrationContext
             .Name.ShouldBe("test");
     }
 
-    // Demonstrates the current workaround: an explicit wait condition keeps the tracked session
-    // open until the cascading message is actually received, past the HTTP response. If this PASSES
-    // while the test above FAILS, it proves the message IS sent — just after the response is
-    // written (and after ExecuteAndWaitAsync's default quiescence check has already returned).
+    // An explicit wait condition is no longer required (the test above passes without one), but it
+    // must of course still work.
     [Fact]
-    public async Task workaround_with_explicit_wait_condition()
+    public async Task tracked_session_with_explicit_wait_condition()
     {
         var tracked = await Host
             .TrackActivity()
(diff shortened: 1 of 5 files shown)


jeremydmiller:
Follow-up commit `66da0ce7` extends the same reordering rule to the **multi-tenant** EF Core path (`StartDatabaseTransactionForDbContext`), which had the same response-before-flush ordering plus a latent flush-before-commit (its raw `CommitTransactionAsync` ran at the end of the wrap, after the response writer, and the standalone `FlushOutgoingMessages` ran after the response and before that commit).

- `StartDatabaseTransactionForDbContext` no longer commits at the end of its wrap.
- New `CommitTenantedDbContextTransaction` postprocessor commits the DbContext transaction then flushes the MessageContext, added right after `SaveChanges` (before the response writer). It implements `IFlushesMessages` so the chain doesn't also add a standalone `FlushOutgoingMessages` (no double flush, no flush-before-commit).

Net generated order for multi-tenant chains is now `handler; SaveChanges; commit; flush; WriteResponse` — matching the single-DbContext path. Relying on the `EfCoreTests.MultiTenancy` CI suite to validate the multi-tenant runtime paths (tenant databases aren't easily provisioned locally).

PR #2922, opened 2026-05-27
No message in wolverine_incoming_envelopes table when using aggregate handler

JurJean:
I think I found a bug where no message is stored in wolverine_incoming_envelopes when using an aggregate handler using the marten integration. I have reproduced my expectation in a unit test.

My colleague found out that when you put `await documentSession.SaveChangesAsync().ConfigureAwait()` at the end of the generated handler, the message is does actually appear in the table.

### DIFF
--- a/src/Persistence/MartenTests/Bugs/Bug_aggregate_does_not_publish.cs
+++ b/src/Persistence/MartenTests/Bugs/Bug_aggregate_does_not_publish.cs
@@ -1,0 +1,129 @@
+﻿using IntegrationTests;
+using JasperFx.Resources;
+using Marten;
+using JasperFx.Events.Projections;
+using MartenTests.AggregateHandlerWorkflow;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Hosting;
+using Npgsql;
+using Shouldly;
+using Weasel.Postgresql;
+using Wolverine;
+using Wolverine.Marten;
+using Wolverine.RDBMS;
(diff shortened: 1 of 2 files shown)


jeremydmiller:
@JurJean Sorry man, this was one giant red herring and it would have helped if I'd paid more attention to you in Discord.

Rename your handler from `AggregateHandler` (which has specific meaning to Wolverine) to AlmostAnythingElseHandler. The most common idiom is to name it [MessageTypeName]Handler

jeremydmiller:
Thanks for the detailed repro, @JurJean! I dug in — this turned out to be the `*AggregateHandler` naming convention rather than an outbox bug: because the handler type is named `AggregateHandler`, Wolverine auto-promotes it into Marten's aggregate event-sourcing workflow, so the return value is appended to the event stream as an event instead of being published as a cascading message. Renaming the handler so it doesn't end in `AggregateHandler` (or returning events intentionally) resolves it.

I've opened #2924 (with you retained as co-author) that supersedes this PR: it turns your repro into a deterministic regression test, adds a startup warning when a `[ReadAggregate]` handler is promoted purely by the naming convention, and documents the interaction.

JurJean:
Thanks for looking into it! I don't think it's necessarily related to the handler name though. I updated my branch,it's now another name.

Also, I'd expect (as a noob) that at least a DeliveryMessage would still be 'published' regardless.

I will probably debug a little more tomorrow.

PR #2941, opened 2026-05-28
Schedule in future reading aggregate

JurJean:
I expanded on the improved version of my initial proposed test, to try to prove my point again that when using `[ReadAggregate]`, published messages do not end up in the `wolverine_incoming_envelopes` table.

Since non-scheduled messages are published right away the issue is not appearing in the current test. The table is still omitted though. By scheduling the messages in the future I believe I expose the issue.

When stepping through the code, I see the StoreIncomingEnvelope in the unit of work:
afbeelding

But it never gets saved.


jeremydmiller:
@JurJean Definite problem this time. Apparently a permutation that hadn't been addressed before. There'll be a 6.1.1 release within a couple hours to address that.

Issue #536, opened 2023-09-08
In Memory Queues don't seem to work when IDocumentSession is injected into EndPoint's method


JoshuaNitschke:
**STEPS TO REPRODUCE**
1) Clone https://github.com/JoshuaNitschke/wolverine-issue-reproduction
2) Add postgres database as a user secret
2) Start the web api
3) Use the swagger page to execute all four end points

**RESULT**
The handlers for the first two endpoints will not execute (as determined by the handlers never firing logs nor hitting break points when running in debug mode).  However, the second two endpoints work just fine.  The only difference between the first set of endpoints and the second is that the first two have an unused dependency injected `IDocumentSession session`, which I wouldn't expect to affect the message bus behavior.

```
No Wolverine extensions are detected
info: Wolverine.Runtime.WolverineRuntime[0]
      Exporting Open Telemetry metrics from Wolverine with name Wolverine:WolverineIssueReproduction.WebApi, version 1.6.1.0

info: Wolverine.Runtime.WolverineRuntime[0]
      Starting Wolverine messaging for application assembly WolverineIssueReproduction.Application, Version=1.0.0.0, Culture=neutral, PublicKeyToken=null
info: Wolverine.Configuration.HandlerDiscovery[0]
      Searching assembly WolverineIssueReproduction.WebApi, Version=1.0.0.0, Culture=neutral, PublicKeyToken=null for Wolverine message handlers
info: Wolverine.Configuration.HandlerDiscovery[0]
      Searching assembly WolverineIssueReproduction.Application, Version=1.0.0.0, Culture=neutral, PublicKeyToken=null for Wolverine message handlers
info: Wolverine.Transports.ListeningAgent[0]
      Started message listening at dbcontrol://76b2943d-c3ab-437d-89d6-e0b21b42203a/
info: Wolverine.Runtime.Agents.NodeAgentController[0]
      Starting agents for Node 76b2943d-c3ab-437d-89d6-e0b21b42203a with assigned node id 45
info: Wolverine.Runtime.Agents.NodeAgentController[0]
      Found no elected leader on node startup, requesting node 7395fdbe-ea92-410c-952d-0c77c493ad39 to be the new leader
info: Microsoft.Hosting.Lifetime[14]
      Now listening on: http://localhost:5210
info: Microsoft.Hosting.Lifetime[0]
      Application started. Press Ctrl+C to shut down.
info: Microsoft.Hosting.Lifetime[0]
      Hosting environment: Development
[...]

PR #2959, opened 2026-05-29
Fix duplicate var messageContext codegen in Wolverine.Http (#2958)

### LINKED ISSUE #2958: Codegen emits duplicate 'var messageContext' in Wolverine.Http handlers using Wolverine.Marten outbox (regression in 6.0.2)
## Summary

Wolverine.Http endpoint codegen emits `var messageContext = new MessageContext(_wolverineRuntime);` **twice** in the same generated `Handle` method when the chain needs both Wolverine.Marten's `OutboxedSessionFactory.OpenSession(messageContext)` and the trailing `messageContext.FlushOutgoingMessagesAsync()` call (the workaround for #536). This produces `CS0128: A local variable or function named 'messageContext' is already defined in this scope` at first request.

Reproduces on **6.0.2, 6.1.0, 6.2.0, 6.2.1**. Does **not** reproduce on **6.0.0** or **6.0.1**. The regression appeared between 6.0.1 and 6.0.2.

Independent of Marten version — confirmed against Marten 9.0.1, 9.2.0, and 9.3.1.

## Repro shape

Any Wolverine.Http endpoint whose handler takes an `IDocumentSession` (so the Wolverine.Marten `OutboxedSessionFactory.OpenSession(messageContext)` contribution participates) on a host where `IntegrateWithWolverine()` is enabled (so the post-save `messageContext.FlushOutgoingMessagesAsync()` from the #536 workaround also participates):

```csharp
[WolverineGet("/things")]
public static async Task<IReadOnlyList<Thing>> Handle(
    IDocumentSession session,
    CancellationToken ct)
{
    return await session.Query<Thing>().ToListAsync(ct);
}
```

## Generated code

```csharp
public override async System.Threading.Tasks.Task Handle(Microsoft.AspNetCore.Http.HttpContext httpContext)
{
    var messageContext = new Wolverine.Runtime.MessageContext(_wolverineRuntime);
    var messageContext = new Wolverine.Runtime.MessageContext(_wolverineRuntime);  // <-- CS0128
    await using var documentSession = _outboxedSessionFactory.OpenSession(messageContext);
    System.Diagnostics.Activity.Current?.SetTag("handler.type", "...");

    var response = await Handler.Handle(documentSession, httpContext.RequestAborted).ConfigureAwait(false);

    await documentSession.SaveChangesAsync(httpContext.RequestAborted).ConfigureAwait(false);

    // Have to flush outgoing messages just in case Marten did nothing because of https://github.com/JasperFx/wolverine/issues/536
    await messageContext.FlushOutgoingMessagesAsync().ConfigureAwait(false);

    await WriteJsonAsync(httpContext, response);
}
```

Two contributions independently declare `messageContext`:
- Wolverine.Marten's `OutboxedSessionFactory.OpenSession(messageContext)` needs one
- Wolverine core's `messageContext.FlushOutgoingMessagesAsync()` (per #536) needs one

Their codegen frames each emit the `var messageContext = new MessageContext(_wolverineRuntime);` line. 6.0.0 / 6.0.1 deduped these; 6.0.2 onward doesn't.

## Exception

```
System.InvalidOperationException: Compilation failures!

CS0128: A local variable or function named 'messageContext' is already defined in this scope

   at JasperFx.RuntimeCompiler.AssemblyGenerator.Generate(String code)
   at JasperFx.RuntimeCompiler.AssemblyGenerator.Compile(GeneratedAssembly, IServiceVariableSource, String& code)
   at JasperFx.CodeGeneration.DynamicTypeLoader.Initialize(...)
   at Wolverine.Http.HttpChain.buildHandler()
```

## Bisection

| WolverineFx | Marten | Result |
|---|---|---|
| 6.0.0 | 9.0.1 | ✅ boots |
| 6.0.1 | 9.0.1 | ✅ boots |
| 6.0.2 | 9.0.1 | ❌ CS0128 |
| 6.0.2 | 9.3.1 | ❌ CS0128 |
| 6.1.0 | 9.3.1 | ❌ CS0128 |
| 6.2.1 | 9.2.0 | ❌ CS0128 |
| 6.2.1 | 9.3.1 | ❌ CS0128 |

## Workaround

Pin `WolverineFx*` at `6.0.1`. Compatible with Marten 9.0.x.

## Environment

- .NET 10
- WolverineFx.Http + WolverineFx.Marten with `IntegrateWithWolverine()` (Postgres-backed outbox)
- TypeLoadMode: Dynamic (development)

kentcooper:
Closes #2958.

## Summary

When a Wolverine.Http chain takes an `IDocumentSession` parameter, requires the post-save outbox flush per #536, and does NOT participate in tenant-id detection, the generated handler declares `var messageContext = new MessageContext(_wolverineRuntime);` twice and fails to compile with CS0128.

Two consumers of `MessageContext` ask the chain for different request types: `CreateDocumentSessionFrame` requests `IMessageContext`, and `FlushOutgoingMessages` (a `MethodCall(typeof(MessageContext), ...)`) requests `MessageContext`. `MessageBusSource.Matches` accepts both, but `Create` returned a fresh `CreateMessageContextWithMaybeTenantFrame` per call.

## Why now?

The bug is latent on the 6.0.x line (JasperFx 2.0.1) because `MethodVariables` deduped on the produced Variable's actual type. JasperFx 2.1.3+ (bumped in 8206b9603, rolled into 6.1.0) keys the cache on the request type, so `IMessageContext` and `MessageContext` became separate cache entries even though `MessageBusSource` returns a `MessageContext`-typed Variable in both cases. The tenant-detection codepath happens to dedupe via a shared frame emitted at the top of the method, which is why the committed test handlers (`POST_todo_create`, `GET_todo_id`) never reproduced it.

## Fix

Cache the produced frame inside `MessageBusSource`. Scope is per-source, and `HttpChain.Codegen.cs` registers a fresh `MessageBusSource` per HTTP chain, so the cache is exactly one generated handler method. Always return the concrete `MessageContext` variable rather than a `CastVariable` for the interfaces, because downstream consumers like `CreateDocumentSessionFrame` invoke methods (`EnqueueCascadingAsync`, `FlushOutgoingMessagesAsync`) that are not visible through `IMessageContext` — matches the pre-fix behavior of the source always producing the concrete variable.

## Verification

`Bug_marten_outbox_duplicate_message_context_codegen` asserts on the compiled generated source for `POST /todoitems` (existing sample endpoint that matches the failing shape) — forces codegen via `InitializeSynchronously` and counts occurrences of the declaration line.

- PASSES against the 6.0.1 source tree (no fix) — 1 declaration
- FAILS against current main without the fix — CS0128, 2 declarations
- PASSES against current main with the fix — 1 declaration

Full `Wolverine.Http.Tests` suite with the fix: **773 passed, 10 skipped, 0 failed**.

### DIFF
--- a/src/Http/Wolverine.Http.Tests/Bug_marten_outbox_duplicate_message_context_codegen.cs
+++ b/src/Http/Wolverine.Http.Tests/Bug_marten_outbox_duplicate_message_context_codegen.cs
@@ -1,0 +1,89 @@
+using JasperFx.CodeGeneration;
+using JasperFx.Core.Reflection;
+using Shouldly;
+
+namespace Wolverine.Http.Tests;
+
+/// <summary>
+/// Reproducer for the duplicate <c>var messageContext = new MessageContext(_wolverineRuntime);</c>
+/// codegen bug surfaced by an HTTP endpoint that:
+///
+///   (1) takes an <c>IDocumentSession</c> parameter (triggers Wolverine.Marten's
+///       <c>OutboxedSessionFactory.OpenSession(messageContext)</c> contribution), AND
+///   (2) requires the post-save outbox flush per
+///       https://github.com/JasperFx/wolverine/issues/536 (a cascading message return,
+///       <c>IMessageBus</c>/<c>IMessageContext</c> dependency, or a post-processor that
+///       MaySendMessages), AND
+///   (3) does NOT participate in tenant-id detection (so the MessageContext is not
+///       pre-emitted by tenant-aware codegen at the top of the method).
+///
+/// In that combination, <c>CreateDocumentSessionFrame.FindVariables</c> resolves its
+/// context via <c>chain.TryFindVariable(typeof(IMessageContext), VariableSource.NotServices)</c>
+/// — request type IMessageContext. Wolverine.Http's
+/// <see cref="Wolverine.Http.CodeGen.MessageBusSource"/> matches IMessageBus /
+/// IMessageContext / MessageContext. Later in the same chain,
+/// <c>FlushOutgoingMessages : MethodCall(typeof(MessageContext), ...)</c> asks the
+/// chain for typeof(MessageContext) — a DIFFERENT request type — and the source
+/// previously returned <c>new CreateMessageContextWithMaybeTenantFrame().Variable</c>
+/// on every call, producing two distinct frame instances that each emitted
+/// <c>var messageContext = new MessageContext(_wolverineRuntime);</c>. CS0128 at
+/// compile time on the second declaration prevented the chain from ever booting.
+///
+/// JasperFx.CodeGeneration's MethodVariables caches resolved variables by REQUEST
+/// type, so requests for IMessageContext and MessageContext live in separate cache
+/// entries — even though MessageBusSource produces a variable typed as
+/// MessageContext in both cases. The fix caches the produced frame inside
+/// MessageBusSource so that any of the three matched types resolves to the same
+/// frame instance. The cache is per-source, and a fresh MessageBusSource is added
+/// to every HTTP chain in <c>HttpChain.Codegen.cs</c>, so the cache scope is
+/// exactly one generated handler method.
+///
+/// We assert on the compiled generated source code (forced by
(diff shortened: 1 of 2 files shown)


jeremydmiller:
I pulled the changes into a different PR, but kept the co-authorship

PR #2961, opened 2026-05-29
Correct analysis

dmytro-pryvedeniuk:
This PR addresses the following issues with code analysis

*  `Microsoft.CodeAnalysis.NetAnalyzers` is used with condition having a typo that resolves to `false`. It means that SDK version is used instead. **This reference is removed to avoid confusion.**
*  In `Analysis.Build.props` `AnalysisModeReliability` is set to `true`. It's not a valid value, valid ones are listed here - https://learn.microsoft.com/en-us/dotnet/core/project-sdk/msbuild-props#analysismode. Invalid value is ignored.  **Removed to avoid confusion.**
* `Analysis.Build.props` is not referenced by all projects. Only some of them reference it directly in csproj. It means for many projects warnings are skipped. - **Imported Analysis.Build.props in Directory.Build.prop.**
* New warnings discovered are addressed - either fixed or disabled with pragma.


jeremydmiller:
@dmytro-pryvedeniuk Hey man, I appreciate all this kind of thing, but if the analysis stuff causes friction, I will rip it out later

dmytro-pryvedeniuk:
@jeremydmiller I don't add new analyzers, just let the existing ones to analyze all projects. Consistency matters. If with "friction" you mean false positives it's inevitable with any code analysis I am afraid.

jeremydmiller:
"Consistency matters." -- *some* things matter, other things are noise and friction. I'll try it for awhile.

Issue #2876, opened 2026-05-21
Make JasperFx.RuntimeCompiler a development-only dependency (decouple Roslyn from core WolverineFx)


jeremydmiller:
Tracking the final step of the cold-start / AOT-readiness work (#1577, #2715 master, #2746 AOT pillar): stop shipping `JasperFx.RuntimeCompiler` (Roslyn, ~6 MB) as a transitive **runtime** dependency of core `WolverineFx`, so `TypeLoadMode.Static` / AOT apps ship without Roslyn and cold start faster.

## Current state (verified on `main` / 6.0.0-rc.1)

- `src/Wolverine/Wolverine.csproj:58` — `<PackageReference Include="JasperFx.RuntimeCompiler" />`, **no `PrivateAssets`**, so Roslyn flows to every consumer.
- `src/Wolverine/HostBuilderExtensions.cs:128-130` — `UseWolverine()` registers the Roslyn `AssemblyGenerator` **by default** (with an `[UnconditionalSuppressMessage]` IL2026 suppression).
- The opt-in package **`WolverineFx.RuntimeCompilation`** already exists: `UseRuntimeCompilation()` does `Services.TryAddSingleton<IAssemblyGenerator, AssemblyGenerator>()` (idempotent), plus `AddWolverineRuntimeCompilation(IServiceCollection)`.
- `IAssemblyGenerator` is defined in **`JasperFx.CodeGeneration`** (already a core dependency). Only the concrete `AssemblyGenerator` lives in `JasperFx.RuntimeCompiler`. Core's *entire* dependency on the RuntimeCompiler package is the one registration line + the `using JasperFx.RuntimeCompiler;` at `HostBuilderExtensions.cs:21`.

## Goal

Core `WolverineFx` depends only on the `IAssemblyGenerator` **abstraction** (`JasperFx.CodeGeneration`). The Roslyn implementation ships only in `WolverineFx.RuntimeCompilation`. Static-mode apps carry no Roslyn; Dynamic/Auto-mode apps add the package.

## Plan

1. **Drop the core dependency.** Remove the `JasperFx.RuntimeCompiler` `PackageReference` from `src/Wolverine/Wolverine.csproj` and the `using JasperFx.RuntimeCompiler;` + default `RegisterAssemblyGenerator(...)` block from `HostBuilderExtensions.cs`.
[...]

PR #2963, opened 2026-05-29
Bump JasperFx.* 2.2.1 -> 2.2.3

jeremydmiller:
Pin-bump only — picks up upstream JasperFx 2.2.3 fixes. **No Wolverine code changes, no `<Version>` bump.**

## Bumped pins

| Package | From | To |
|---|---|---|
| `JasperFx` | 2.2.1 | **2.2.3** |
| `JasperFx.Events` | 2.2.1 | **2.2.3** |
| `JasperFx.Events.SourceGenerator` | 2.2.1 | **2.2.3** |
| `JasperFx.SourceGenerator` | 2.2.1 | **2.2.3** |

## Unchanged

| Package | Version | Reason |
|---|---|---|
| `JasperFx.RuntimeCompiler` | 5.0.0 | Own 5.x line — the Roslyn runtime-compiler package was decoupled from the JasperFx 2.x family in #2876. |
| Marten family | 9.2.0 | Latest. |
| Polecat | 4.2.1 | Latest. |
| `Directory.Build.props` `<Version>` | 6.2.1 | Pin-bump only; Wolverine version bump intentionally deferred. |

## Skipping 2.2.2

The 2.2.2 revision of the JasperFx family shipped `JasperFx.Events` but not the rest of the sibling set, so there was never a coherent 2.2.2 pin-set to land here. 2.2.3 is the first complete sibling-set after 2.2.1.

## Test plan

- `dotnet build wolverine.slnx -c Release` — clean locally (0 warnings, 0 errors, ~27 s).
- CI `dotnet.yml` gates merge.

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -140,8 +140,33 @@
                      JasperFx.RuntimeCompiler              unchanged (own 5.x line, stays 5.0.0)
                      Marten / Polecat                      unchanged (9.2.0 / 4.2.1)
                    No Wolverine code changes — pure dependency bump.
+
+          6.2.2: Patch release combining a JasperFx pin bump with two related
+                 codegen fixes for the EF Core transaction middleware + post-save
+                 cascade flush path (GH-536):
+                     JasperFx (+ .Events / .Events.SourceGenerator
+                                / .SourceGenerator)        2.2.1 -> 2.2.3 (#2963)
+                     JasperFx.RuntimeCompiler              unchanged (5.0.0)
+                     Marten / Polecat                      unchanged (9.2.0 / 4.2.1)
+                   Codegen fixes (both addressing handler compilation failures
+                   under the EF Core transaction middleware + outbox path):
+                     - CS0128 duplicate `var messageContext` declaration in the
+                       generated Wolverine.Http handler. Cherry-picked from
+                       #2959, originally authored by @kentcooper.
+                     - CS1061 'IMessageContext' missing EnqueueCascadingAsync.
+                       Latent since 020e79013e (2022-09-16); only surfaced once
+                       the variable-cache fix from #2959 made the source produce
+                       a concrete MessageContext-typed Variable. Fixed by
+                       re-targeting CaptureCascadingMessages' MethodCall to the
+                       concrete MessageContext type (matches the pre-existing
+                       FlushOutgoingMessages shape).
+                   Also rolls up #2960 (diagnostics docs from outofrange-consulting)
+                   and skips a long-broken Marten distribution test pending
+                   #2965 (SharedMemory transport cross-node AssignAgent
+                   serialization — bug predates this release line, was failing
+                   on every main commit through V6.2.0 and V6.2.1).
         -->
-        <Version>6.2.1</Version>
+        <Version>6.2.2</Version>
         <RepositoryUrl>$(PackageProjectUrl)</RepositoryUrl>
         <PublishRepositoryUrl>true</PublishRepositoryUrl>
         <EmbedUntrackedSources>true</EmbedUntrackedSources>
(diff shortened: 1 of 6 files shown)


jeremydmiller:
Force-pushed: cherry-picked the codegen fix from #2959 (`2edfb342e` → `9df45bfde` here, authorship preserved as @kentcooper) on top of the JasperFx 2.2.3 pin bump.

The original CI failure on this PR was 2 tests in `Wolverine.Http.Tests.issue_2917_trackactivity_transactional_outbox` — exactly the surface #2959 addresses (`CS0128` duplicate `var messageContext` declaration in the generated HTTP handler for `IDocumentSession` + outbox-flush chains). Bumping JasperFx 2.2.1 → 2.2.3 doesn't itself change `MethodVariables` cache-keying, but the failing test was already latent against JasperFx 2.1.3+ per #2959's analysis — surfacing more reliably under whatever 2.2.3 changed (likely a serializer/options touch that altered codegen ordering enough to expose the dedupe gap).

If #2959 lands first, this branch will rebase down to just the original pin-bump commit. If this lands first, #2959's commit is preserved verbatim with the original author so the credit shows up in both PRs.

CI re-running on `9df45bfde`.

jeremydmiller:
Pushed `195797c54` to address the http CI failures.

## Root cause

The two failing tests in `issue_2917_trackactivity_transactional_outbox` were not the same CS0128 `var messageContext` duplicate-declaration that #2959 fixes — they're a sibling **CS1061** in the same area:

```
'IMessageContext' does not contain a definition for 'EnqueueCascadingAsync'

var messageContext = new Wolverine.Runtime.MessageContext(_wolverineRuntime);
...
await ((Wolverine.IMessageContext)messageContext).EnqueueCascadingAsync(outgoingMessages).ConfigureAwait(false);
```

[`CaptureCascadingMessages`](https://github.com/JasperFx/wolverine/blob/main/src/Wolverine/Runtime/Handlers/CaptureCascadingMessages.cs) declared its `MethodCall` target as `typeof(IMessageContext)` but `EnqueueCascadingAsync` only exists on the concrete `MessageContext` class. The mismatch told JasperFx to emit the `((IMessageContext)messageContext)` cast that then refuses to compile.

The frame has been wired this way since 2022-09-16 (`020e79013e`). It only surfaces on chains that combine the EF Core transaction middleware + post-save cascade flush (GH-536). #2959 made `MessageBusSource` always produce a concrete `MessageContext` Variable — which is the prerequisite for the fix to work, but without ALSO flipping this frame's target type the cast was still being emitted.

The fix matches how the sibling [`FlushOutgoingMessages`](https://github.com/JasperFx/wolverine/blob/main/src/Wolverine/Persistence/FlushOutgoingMessages.cs) already declares its `MethodCall` (`typeof(MessageContext)`, not the interface).

## Verification

Full `wolverine.slnx` build clean (0 warnings, 0 errors). CI re-running on `195797c54`.

jeremydmiller:
Pushed `141fee293` to skip the long-broken `MartenTests.Distribution.with_ancillary_stores.spread_out_over_multiple_hosts` test pending #2965.

Investigation summary (full details in #2965):
- Marten CI has been red on this test for every main commit including the V6.2.0 and V6.2.1 release tags
- Not a regression from #2963; bisected and confirmed
- All 6 event-subscription agents enumerate + start correctly on a single host (verified via DI/family diagnostics)
- Root cause is in `SharedMemory` transport: `SendAsync` posts envelopes directly to the receiver block without serializing, so `AssignAgent` system commands arrive at other simulated nodes with empty `Data`/`MessageType` and the receive pipeline can't deserialize them
- Out of scope for a JasperFx pin-bump release; tracked as #2965 with two proposed fix shapes

CI re-running on `141fee293`. Release flow continues from here.

PR #2975, opened 2026-05-29
Fix flaky leader_switchover_between_nodes test

dmytro-pryvedeniuk:
This PR fixes flaky test from `CompianceTests` project (used by multiple descendants). Flakiness is observed on RavenDB tests.

The test assumes the first started extra host wins leadership, but actually all hosts compete for it.


jeremydmiller:
@dmytro-pryvedeniuk That assumption *should* be correct though. It *should* be using the nodes in the order they were registered. I'd want a closer look first.

Issue #2989, opened 2026-05-30
Inline HTTP request/reply: full integration coverage + CritterWatch control mechanism for HTTP-only services


jeremydmiller:
Follow-up to #2966 (optimized inline request/reply over the Wolverine.Http transport, where `InvokeAsync<T>` reads the reply straight off the HTTP response). #2966 ships the sender path (`MessageRoute` → `IInlineRequestReplyEndpoint` → `HttpEndpoint.InvokeRemoteAsync`), the receiver failure mapping, and component-level tests (sender happy/failure paths with a fake transport client). This issue tracks the remaining hardening + the CritterWatch use case.

## 1. Full integration test coverage

The #2966 tests stub `IWolverineHttpTransportClient` to exercise the sender routing/serialization logic in-process. Add real cross-process (Kestrel) coverage:

- [ ] End-to-end `InvokeAsync<T>` from a real sender host to a real receiver host over HTTP — happy path and handler-failure (`WolverineRequestReplyException`) path.
- [ ] Assert the sender allocates **no `ReplyListener<T>`** and runs **no listening endpoint** for the reply (the acceptance criterion from #2966).
- [ ] **Outbox-commit-before-response**: with a transactional outbox on the receiver, confirm the HTTP response is withheld until the outbox commits, and that cascading messages emitted by the handler are in the same transactional unit.
- [ ] Run the existing request/reply **compliance suite** against the Wolverine.Http sender transport.
- [ ] **`Result<T>` composition** (#2221): a handler returning `Result<T>` over inline HTTP request/reply still unwraps end-to-end on the caller.
- [ ] CloudEvents-encoded request with a binary reply, and content-negotiation edge cases.
- [ ] Timeout / cancellation behavior (the `DefaultRemoteInvocationTimeout` and a passed `CancellationToken`).

## 2. CritterWatch control mechanism for HTTP-only services

Inline HTTP request/reply lets CritterWatch issue **request/reply control commands to Wolverine services that only expose HTTP** — services with no broker and no inbound listening endpoint of their own. Today CritterWatch's control path assumes a brokered/queue control endpoint; HTTP-only services can't participate.

- [ ] Document (and, where needed, wire up) using `/_wolverine/invoke` + inline request/reply as the **control channel** for HTTP-only services: how CritterWatch sends a control/query command and reads the structured reply over plain HTTP.
[...]

Issue #2221, opened 2026-02-21
Formal support for the stupid Result pattern in "Wolverine as MediatR" usage


jeremydmiller:
It comes up a lot, and I'm feeling like you almost can't fight it off. Make it easy for the Wolverine "Railway Programming" model to be usable with the stupid `Result<T>` thing. Maybe automatically map the real message w/ Result<T> on success. Try to get this done w/o the stupid high ceremony code that people using MediatR love in LinkedIn posts on "Exceptions bad, clumsy Result<T> good" posts

nkosi23:
I just hope it'll be opt-in (or at least transparent) for users not interested in dealing with Result<T>.

dmytro-pryvedeniuk:
@jeremydmiller You used "stupid" word twice. Where is old good "it depends"? :) IMO it should be up to developers how to implement their API: throw an exception or return "failed" result. Both work as long as the contract is clear for the callers. 
On the other hand why should Wolverine care? In the example below Wolverine is used with OneOf result. Is something else meant in the request?
  
```
public record CompleteTask(TaskId Id);

public static class CompleteTaskHandler
{
  public static async ValueTask<Domain.Task?> LoadAsync(
    CompleteTask command, ITaskRepository taskRepository, CancellationToken ct)
  {
    return await taskRepository.GetByIdAsync(command.Id.Value, ct);
  }

  public static IEnumerable<object> Handle(CompleteTask command, Domain.Task? task)
  {
    if (task is null)
      return [new NotFound()];

    var result = task.Complete(DateTimeOffset.UtcNow);

    return result.Match<object[]>(
      success => [
        new Success(), // Result A
        new TaskCompleted(command.Id)], // Event A
      alreadyCompleted => [
        new TaskIsAlreadyCompleted(), // Result B
        new TaskFailedToComplete(command.Id, "Already completed")] // Event B
    );
  }
}

[GenerateOneOf]
public partial class CompleteTaskResult : OneOfBase<Success, NotFound, TaskIsAlreadyCompleted> { }
```

jeremydmiller:
@nkosi23 100% optional, and honestly, what I want is something that will wrap the Result stuff for you outside your handler so your handler code can remain as simple as possible.

[...]

PR #2990, opened 2026-05-30
Optimized inline request/reply over the Wolverine.Http transport (GH-2966)

### LINKED ISSUE #2966: Optimized inline request/reply over the Wolverine.Http sender transport
## Context

`IMessageBus.InvokeAsync<T>(message)` today routes every external-transport call through `MessageRoute.RemoteInvokeAsync<T>` (`src/Wolverine/Runtime/Routing/MessageRoute.cs:197`):

```csharp
var waiter = _replyTracker.RegisterListener<T>(envelope, cancellation, timeout.Value);
// send the envelope...
return await waiter.Task;   // separate envelope, correlated by Id, on a listener loop
```

That's the right model for brokered transports (Rabbit, SQS, ASB, Kafka) where requests and replies are independent messages. It's wasteful for **HTTP**, where the protocol already provides a response slot:

- Sender must run a listening endpoint just to receive its own replies.
- A `ReplyListener<T>` + cross-loop `TaskCompletionSource<T>` is allocated for every call, even though the HTTP response body could have *been* the reply.
- Failures round-trip as separate envelopes rather than using HTTP status semantics.

## Goal

When `InvokeAsync<T>` routes to a `Wolverine.Http` transport endpoint, the HTTP response body **is** the reply. No `ReplyListener`, no listener loop on the sender, no cross-loop `TaskCompletionSource`.

## Design decisions (settled in scoping)

1. **Endpoint-by-endpoint via transport capability — baked into `MessageRoute` at routing-compile time, no runtime `if/then` branching.** The routing layer picks a different `MessageRoute` shape when the target transport declares an inline-request/reply capability. *Verify during the spike whether the existing internal model already exposes a hook for this — long-planned, but confirm before adding one.*
2. **Header-switched on a single endpoint URL.** Same receiver URL as today's batch endpoint, distinguished by a `Wolverine-Reply: inline` header (exact name TBD). Lower surface area than a second URL.
3. **HTTP-native failure mapping.** Handler exception → HTTP 500 with an envelope-shaped error body. The HTTP-flavoured `MessageRoute` translates that back into a `FailureAcknowledgement` on the sender side so the existing `WolverineRequestReplyException` flow is preserved for the caller. Stay with HTTP idioms when possible.
4. **Outbox safety on the receiver: hold the HTTP response until the outbox commits**, then write the reply. Prioritises caller-observable consistency over latency. Cascades emitted during the handler are part of the same transactional unit.
5. **Wire format**: envelope-in-body both ways (preserves headers, MessageId, tenancy, correlation, OTel context). Standard `EnvelopeSerializer`.
6. **Version target**: 6.x additive. Auto-activates per-endpoint when the transport declares the capability; no user-facing opt-in flag since the routing decision is structural, not configurable.

### Sender
- New `MessageRoute` variant (e.g. `InlineReplyMessageRoute`) selected at compile time when the route's `Endpoint` is HTTP and declares the inline-reply capability. Its `InvokeAsync<T>` does **not** call `_replyTracker.RegisterListener`; it `POST`s the request envelope, deserialises the response envelope from the response body, extracts the message, and returns.
- HTTP-flavoured `MessageRoute` handles 500 + envelope-shaped error → throws `WolverineRequestReplyException` matching today's broker behaviour (same exception type, same callers).
- Existing fire-and-forget HTTP sending (`HttpSenderProtocol.SendBatchAsync`) continues to handle non-`InvokeAsync<T>` traffic via the same URL.

### Receiver
- The HTTP transport receive endpoint inspects `Wolverine-Reply: inline`. When present:
  - Deserialise a *single* envelope (not a batch).
  - Run the handler chain to completion.
  - Capture the handler's return via `HandlerChain.UseForResponse(MethodCall)` (already exists — this is the same seam used for broker request/reply).
  - Wait for the receiver outbox to commit, then serialise the response envelope into the HTTP response body and complete with 200.
  - On handler exception, write the envelope-shaped error and respond 500.

### Codegen
- The receiver's handler chain doesn't need new codegen — `HandlerChain.UseForResponse` already produces the response capture. The transport layer reads the captured response variable and inlines it into the HTTP response.

## Acceptance criteria

- [ ] An HTTP endpoint exposed by the Wolverine.Http sender transport declares an inline-request/reply capability that `MessageRoute` consumes at routing-compile time (no runtime `if/then`).
- [ ] `InvokeAsync<T>` against an HTTP-routed message reads the response body directly — verified by asserting the sender allocates no `ReplyListener<T>`.
- [ ] Sender does not require a listening endpoint to receive replies.
- [ ] Handler exceptions surface as HTTP 500 with an envelope-shaped body; the sender re-throws `WolverineRequestReplyException`, matching existing broker semantics.
- [ ] Receiver holds the HTTP response until the outbox commits when an outbox is configured. Cascades emitted during the handler are part of the same transactional unit.
- [ ] Fire-and-forget HTTP sends (existing batch path) continue to work unchanged.
- [ ] Compliance tests in the request/reply suite pass against the Wolverine.Http sender transport.

## Open questions (to settle during the spike)

- Does the existing `Endpoint`/`MessageRoute` internal model already expose a hook for "this transport supports inline request/reply"? (Long-planned; verify before adding one.)
- Exact name for the header: `Wolverine-Reply: inline`, `Wolverine-Mode: invoke`, or something else.
- Should the response envelope flow through the receiver's outbox at all (i.e., enqueue the response envelope and let the outbox dispatcher write the HTTP response), or is the response a pure in-process payload that the transport writes itself once the outbox commits?

## Related

- gRPC companion: #2967 — same architecture, different protocol. Priority is HTTP first (this issue), gRPC follows.
- Shares the `HandlerChain.UseForResponse` codegen seam discussed in #2221 (custom `Result<T>` support). Worth verifying the two compose cleanly so a handler returning `Result<T>` over inline HTTP request/reply still unwraps end-to-end.

jeremydmiller:
Closes #2966.

When `IMessageBus.InvokeAsync<T>(message)` routes to a Wolverine.Http transport endpoint, the HTTP response body **is** the reply — no `ReplyListener`, no listener loop on the sender, no cross-loop `TaskCompletionSource`.

## Sender
- **New `IInlineRequestReplyEndpoint` capability** (Wolverine core). `MessageRoute` caches `endpoint as IInlineRequestReplyEndpoint` once **at construction** (routing-compile time); when present, `RemoteInvokeAsync<T>` sends the request and reads the reply envelope straight back instead of registering a `ReplyListener` — **no per-message branching**, null for every other transport.
- **`HttpEndpoint`** implements the capability: serializes the request, POSTs to its `OutboundUri`, deserializes the reply envelope from the HTTP response body.
- **`IWolverineHttpTransportClient.InvokeAsync`** returns `(status, body)` so a 500 carrying an envelope-shaped error reply is read regardless of status.
- A receiver handler failure comes back as a `FailureAcknowledgement` reply and is rethrown as **`WolverineRequestReplyException`** — identical caller semantics to brokered request/reply.

## Receiver
The existing `/_wolverine/invoke` endpoint already ran the handler and wrote `Envelope.Response` to the body. Now a handler exception maps to **HTTP 500 + a `FailureAcknowledgement` envelope body** (using `IntrinsicSerializer.Write` — `WriteMessage` intentionally throws). Outbox commit already happens (via `InvokeInlineAsync`'s flush) before the response is written.

## Design notes
- **Auto-activates per endpoint** via the transport capability — no user-facing opt-in flag, since the routing decision is structural.
- The brokered request/reply path is untouched (`_inlineRequestReply` is null for non-HTTP endpoints).
- Fire-and-forget HTTP sends (batch/inline publish) are unchanged.

## Tests
`inline_request_reply_sender` drives the full sender path (`InvokeAsync<T>` → `MessageRoute` inline branch → `HttpEndpoint` → reply) with a fake transport client standing in for the receiver. The host configures **no listening endpoint**, so the tests only pass because the reply is read off the response slot (otherwise they'd time out):
- happy path returns the typed reply;
- a receiver-side failure surfaces as `WolverineRequestReplyException`.

(Pre-existing unrelated flake noted: `HttpTransportExecutorTests.batch_with_multiple_queues_routes_to_correct_queue` fails on clean `main` too.)

## Verification
- New sender tests green; HTTP transport suite otherwise green.
- `dotnet build wolverine.slnx -c Release` clean.

## Docs & follow-up
- Docs: new **"Inline Request/Reply"** section in `guide/http/transport.md`.
- **#2989** tracks the remaining hardening — full cross-process (Kestrel) integration + compliance-suite coverage, the outbox-commit-before-response assertion, `Result<T>` composition (#2221), and **documenting/using this as the control mechanism for CritterWatch against HTTP-only services**.

### DIFF
--- a/docs/guide/http/transport.md
+++ b/docs/guide/http/transport.md
@@ -159,6 +159,58 @@ When using `.UseDurableOutbox()`, messages are first persisted to your configure
 (PostgreSQL, SQL Server, etc.) and then delivered in the background. This guarantees that messages
 are never lost, even if the sending application crashes or the receiver is temporarily unavailable.
 
+## Inline Request/Reply
+
+When you call `IMessageBus.InvokeAsync<T>(message)` and the message is routed to an HTTP transport
+endpoint, Wolverine uses the HTTP response body **as** the reply. The request envelope is POSTed to the
+receiver's `/_wolverine/invoke` endpoint and the response envelope is read straight back off the HTTP
+response:
+
+```cs
+// Sender — route the request to the receiver's /_wolverine/invoke endpoint
+opts.PublishMessage<PriceQuoteRequest>()
+    .ToHttpEndpoint("https://pricing.example.com/_wolverine/invoke");
+```
+
+```cs
+// Anywhere with an IMessageBus
+var quote = await bus.InvokeAsync<PriceQuote>(new PriceQuoteRequest("SKU-123"));
+```
+
+This is materially cheaper than request/reply over a brokered transport (Rabbit MQ, SQS, Azure Service
+Bus): because HTTP already provides a response slot, there is
+
+- **no `ReplyListener` and no listener loop on the sender** — the reply is the HTTP response, not a
+  separate correlated message, so the sender does not need to run a listening endpoint just to receive
+  its own replies;
+- **no cross-loop `TaskCompletionSource`** allocated per call.
+
+The optimization is **structural and automatic** — there is no opt-in flag. Any `InvokeAsync<T>` routed
+to an HTTP endpoint takes this path; the decision is baked into the route when it is compiled, so there
+is no per-message branching. Fire-and-forget sends (`PublishAsync` / `SendAsync`) continue to use the
+batched/inline send path described above.
+
+### Failure semantics
+
+If the receiver's handler throws, the receiver responds with HTTP 500 and an envelope-shaped error
+body, and the sender re-throws a `WolverineRequestReplyException` — exactly the same caller-visible
+behavior as brokered request/reply:
+
+```cs
+try
+{
+    var quote = await bus.InvokeAsync<PriceQuote>(new PriceQuoteRequest("SKU-123"));
+}
+catch (WolverineRequestReplyException e)
+{
+    // The receiver's handler failed; e.Message carries the failure detail.
+}
+```
+
+When the receiver is configured with a transactional outbox, the HTTP response is held until the outbox
+commits, so any cascading messages emitted by the handler are part of the same transactional unit
+before the caller observes the reply.
+
 ## CloudEvents Support
 
 The HTTP transport supports the [CloudEvents](https://cloudevents.io/) specification for
--- a/src/Http/Wolverine.Http.Tests/Transport/HttpScheduledMessageTests.cs
+++ b/src/Http/Wolverine.Http.Tests/Transport/HttpScheduledMessageTests.cs
@@ -101,6 +101,11 @@ public class TestWolverineHttpTransportClient : IWolverineHttpTransportClient
 
         return Task.CompletedTask;
     }
+
+    public Task<InlineHttpReply> InvokeAsync(string uri, Envelope envelope, JsonSerializerOptions serializerOptions)
+    {
+        return Task.FromResult(new InlineHttpReply(200, Array.Empty<byte>()));
+    }
 }
 
 public record ScheduledTestCommand(string Id);
--- a/src/Http/Wolverine.Http.Tests/Transport/inline_request_reply_sender.cs
+++ b/src/Http/Wolverine.Http.Tests/Transport/inline_request_reply_sender.cs
@@ -1,0 +1,94 @@
+using System.Text.Json;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Hosting;
+using Shouldly;
+using Wolverine.Http.Transport;
+using Wolverine.Runtime.RemoteInvocation;
+using Wolverine.Runtime.Serialization;
+using Wolverine.Transports;
+using Xunit;
+
+namespace Wolverine.Http.Tests.Transport;
+
+// GH-2966: the sender side of optimized inline request/reply over the Wolverine.Http transport.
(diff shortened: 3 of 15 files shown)


jeremydmiller:
Updated per review: the sender path is now **polymorphic instead of a runtime `if`**.

- The per-call `if (_inlineRequestReply != null)` branch in `MessageRoute.RemoteInvokeAsync<T>` is gone. The "send the request and resolve the reply" step is a `protected virtual sendAndAwaitReplyAsync<T>` seam — the base keeps the brokered reply-tracker/listener path; a new **`InlineReplyMessageRoute : MessageRoute`** overrides it to read the reply off the transport response.
- The shape is chosen **once at routing-compile time** by a new `MessageRoute.For(messageType, endpoint, runtime)` factory (HTTP / `IInlineRequestReplyEndpoint` → the subclass, everything else → base). All route-construction sites now go through `MessageRoute.For`; it's the only place that `new`s a route. No per-message branching anywhere.

Verified: the two inline request/reply tests pass, 41 core request/reply + routing tests pass, and `dotnet build wolverine.slnx -c Release` is clean.

Also filed #2994 to spec using the HTTP transport as a Wolverine cluster's **node control endpoint** (control queue), which builds directly on this inline request/reply work.

PR #3017, opened 2026-06-03
Fix NATS JetStream native scheduled send (err 10190) and add configurable schedule-subject suffix.

midub:
## Problem

`Wolverine.Nats` fails to send any scheduled message over NATS JetStream native scheduling (NATS Server 2.12+). The transport publishes the scheduling control message to the **same** subject it sets as the `Nats-Schedule-Target` header, but NATS requires the control message to be published to a subject **different** from the target. The server rejects the publish:

```json
{"error":{"code":400,"err_code":10190,"description":"message schedules target is invalid"}}
```

The send throws, nothing is materialized, and the consumer never receives the message.

## Fix

`JetStreamPublisher` now keeps `Nats-Schedule-Target` = the real destination subject (where the listener's exact-`FilterSubject` consumer reads and the server materializes the message at fire time) and publishes the **control message** to a derived subject:

- Derived subject defaults to `<subject>.scheduled`.
- Configurable per endpoint via `UseScheduleSubjectSuffix(...)`.
- Both the target and the schedule subject must be covered by the same stream → use a `prefix.>` filter (a single-token `*` / exact-subject filter will not cover the derived subject).
- An empty/whitespace suffix falls back to `.scheduled`, so the control subject can never collapse onto the target again.

## Tests

- **`schedule_send_over_native_jetstream_scheduling`** — drives the native path by putting `.UseJetStream(stream)` on the publishing endpoint. Times out before the fix, passes after. Asserts the native path is engaged (not the durable fallback). Skips on NATS < 2.12.
- **`overridden_suffix_is_used_for_native_scheduling`** — provisions the stream with explicit subjects covering only the target and the `.override-scheduled` subject (default `.scheduled` deliberately not covered), so delivery only succeeds if the configured suffix is actually applied.

### DIFF
--- a/docs/guide/messaging/transports/nats.md
+++ b/docs/guide/messaging/transports/nats.md
@@ -360,9 +360,12 @@ derived **schedule subject** — the destination subject plus a suffix (default 
 (`orders.created`). At the scheduled time the server materializes a new message onto the target subject,
 where your listener's consumer receives it; the control message itself is never delivered to consumers.
 
-Both the schedule subject and the target must be covered by the **same stream**, so use a wildcard subject
-filter such as `orders.>` (a single-token `*` filter or an exact-subject filter will **not** cover
-`<subject>.scheduled`). Override the suffix per publishing endpoint when needed:
+Both the target and the derived schedule subject must be covered by the **same stream**. The schedule
+subject is the target plus an extra suffix token (`orders.created` → `orders.created.scheduled`), so it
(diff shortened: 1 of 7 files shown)


jeremydmiller:
@midub I'll try to get this reviewed and in today

jeremydmiller:
@midub Looks good to me, I'll be getting this in today. I think that will be 6.4.2, but might be 6.5. Depends on what else comes in in the meantime.

PR #3049, opened 2026-06-08
fix: Preserve custom dead letter exchange arguments

frankvdb7:
**Summary**
Fix RabbitMQ queue declaration so an explicitly configured x-dead-letter-exchange is not overwritten by Wolverine’s global CustomizeDeadLetterQueueing(...) default.

**What changed**
Updated RabbitMQ queue declaration logic to preserve a user-supplied x-dead-letter-exchange argument.
The transport now only applies the default dead-letter exchange when the queue does not already define one.
Added a regression test covering the override scenario.

**Why**
PublishMessage<T>().ToRabbitQueue(..., configure => configure.Arguments["x-dead-letter-exchange"] = "...") was not effective when a global native dead-letter queue was configured with CustomizeDeadLetterQueueing(new DeadLetterQueue("default-dlx")).

The queue declaration code was always writing the default DLX argument last, so the explicit per-queue value was lost.

**Verification**
dotnet test src/Transports/RabbitMQ/Wolverine.RabbitMQ.Tests/Wolverine.RabbitMQ.Tests.csproj --filter FullyQualifiedName~RabbitMqQueueTests
Added regression coverage in RabbitMqQueueTests

### DIFF
--- a/src/Persistence/Oracle/Wolverine.Oracle/OracleMessageStore.Incoming.cs
+++ b/src/Persistence/Oracle/Wolverine.Oracle/OracleMessageStore.Incoming.cs
@@ -215,8 +215,8 @@ internal partial class OracleMessageStore
         insertCmd.With("messageType", envelope.MessageType ?? string.Empty);
         insertCmd.With("receivedAt", envelope.Destination?.ToString() ?? string.Empty);
         insertCmd.With("source", envelope.Source ?? string.Empty);
-        insertCmd.With("exceptionType", exception?.GetType().FullNameInCode() ?? string.Empty);
-        insertCmd.With("exceptionMessage", exception?.Message ?? string.Empty);
+        insertCmd.With("exceptionType", exception.DeadLetterExceptionType() ?? string.Empty);
+        insertCmd.With("exceptionMessage", exception.DeadLetterExceptionMessage() ?? string.Empty);
         insertCmd.Parameters.Add(new OracleParameter("sentAt", OracleDbType.TimeStampTZ) { Value = envelope.SentAt.ToUniversalTime() });
         insertCmd.With("replayable", 0); // Oracle stores bool as NUMBER(1)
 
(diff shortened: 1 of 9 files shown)


jeremydmiller:
@frankvdb7 Why is this necessary? You can already override the DLQ exchange per queue as is in the fluent interface.  And the tests break anyway, so this isn't working.

frankvdb7:
Let me explain:
We had 2 seperate services - a publisher - a consumer. We share the way Wolverine is setup between the publisher and consumer. Autoprovisioning, dead letter queuing on 'default-dlx', etc. We set the publisher and consumer up in a way to use a 'custom-dlx' to override the 'default-dlx'.

The problem we had is that when the publisher started first, it created a queue with the 'default-dlx' binding. If we had started the consumer first, it would have created it correctly with the 'custom-dlx' binding. Our workaround for the publisher, was to adjust the 'x-dead-letter-exchange' argument to 'default-dlx', as there is no fluent api to adjust the dead letter queuing for publishing messages (for autoprovision).

But the 'x-dead-letter-exchange' value was then getting overwritten during AutoProvision() in the publisher when Wolverine declared the queue, because RabbitMqQueue.DeclareAsync() always applied the transport default DLQ ('default-dlx') exchange again.


Example:
```
// Service A: the consumer service
services.UseWolverine(opts =>
{
    opts.UseRabbitMq()
        .AutoProvision()
        .CustomizeDeadLetterQueueing(new DeadLetterQueue("default-dlx"));

    opts.ListenToRabbitQueue("orders")
        .DeadLetterQueueing(new DeadLetterQueue("orders-dlx"));
});
```

```
// Service B: the publisher service
services.UseWolverine(opts =>
{
    opts.UseRabbitMq()
        .AutoProvision()
        .CustomizeDeadLetterQueueing(new DeadLetterQueue("default-dlx"));

    opts.PublishMessage<OrderPlaced>()
        .ToRabbitQueue("orders", q =>
        {
            // No dedicated PublishMessage<T>().DeadLetterQueueing(...) API exists here
            // so this is the only way to try to set a per-queue DLQ on the publish side
            q.Arguments["x-dead-letter-exchange"] = "orders-dlx";
        });
});
```

Depending on the order of the execution it will either create 'orders' queue with dlx 'orders-dlx' (good) or 'default-dlx' (wrong)

jeremydmiller:
@frankvdb7 I completely get your need, I'm asking you why you need all different syntax instead of using what we already have in the fluent interface to override the DLQ exchange on a single queue? If there's a bug, sure, fix that.

And you need some real test coverage here too, not just the mock heavy test.

frankvdb7:
@jeremydmiller I am going to open a new PR with a better way to solve our issue and align with the fluent API.

PR #3050, opened 2026-06-08
Fix EF Core durable outbox ignoring overridden message storage

Ferchke7:
Fixes `EfCoreEnvelopeTransaction` so `IDbContextOutbox<T>` respects `MessageContext.OverrideStorage()` when persisting outgoing/incoming envelopes through the raw EF Core outbox path.

Previously, `EfCoreEnvelopeTransaction` captured the `IMessageDatabase` once in its constructor. If code later called `OverrideStorage(ancillaryStore)`, the message context changed, but the transaction kept writing SQL against the original main store schema.

## Why

This matters for cases where a high-volume workflow needs its own ancillary outbox table while still using EF Core transactional outbox semantics.

Example:

```csharp
var outbox = scope.ServiceProvider
    .GetRequiredService<IDbContextOutbox<AppDbContext>>();

var context = (MessageContext)outbox;

context.OverrideStorage(ancillaryStore);

await context.PublishAsync(new BulkRecordApproved(recordId));
await outbox.SaveChangesAndFlushMessagesAsync();
```

Expected:

```text
bulk_outbox.wolverine_outgoing_envelopes
```

Actual before this fix:

```text
wolverine.wolverine_outgoing_envelopes
```

That meant bulk operations could flood the main outbox table and delay unrelated outgoing messages.

## Fix

`EfCoreEnvelopeTransaction` now resolves the active `IMessageDatabase` at persistence time instead of always using the constructor-captured database.

It only switches to the override database when the override store shares the same `DbDataSource` as the original database. That keeps the SQL schema switch safe for same-physical-database ancillary stores, while avoiding accidentally generating SQL for one database and executing it on another `DbContext` connection.

Affected paths:

- `PersistOutgoingAsync(Envelope)`
- `PersistOutgoingAsync(Envelope[])`
- `PersistIncomingAsync(Envelope)`
- `CommitAsync`

### DIFF
--- a/src/Persistence/EfCoreTests/Bugs/Bug_DurableLocalQueue_ancillary_store_routing.cs
+++ b/src/Persistence/EfCoreTests/Bugs/Bug_DurableLocalQueue_ancillary_store_routing.cs
@@ -69,6 +69,12 @@ public class MainLocalQueueDbContext(DbContextOptions<MainLocalQueueDbContext> o
     }
 }
 
+/// <summary>
+/// Raw DbContext without Wolverine's envelope mappings. This forces
+/// EfCoreEnvelopeTransaction to use raw SQL against the active message database.
+/// </summary>
+public class RawAncillaryOutboxDbContext(DbContextOptions<RawAncillaryOutboxDbContext> options) : DbContext(options);
+
 public class AncillaryLocalQueueDoc
 {
     public Guid Id { get; set; }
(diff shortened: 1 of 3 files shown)


jeremydmiller:
@Ferchke7 Hey man, can you fix the broken tests, but first, **why** are you doing that? That's not idiomatic usage at all. Can you give me a little more context about what you're trying to accomplish here?

Ferchke7:
@jeremydmiller Sorry, I missed that one, on a way fixing the error! This one i missed somehow. 

- The goal is to isolate a high-volume EF Core transactional outbox workload from the main Wolverine outbox table.We have a bulk workflow that can enqueue ~200k Kafka events. Those events need to be committed atomically with the EF Core database changes, so we’re using IDbContextOutbox<T>. But if all of those messages go into the main wolverine_outgoing_envelopes table, they can delay unrelated outgoing messages.

- Wolverine already has ancillary message stores, so we’re trying to route this specific outbox workload to an ancillary store in the same PostgreSQL database, just a different schema. The problem is that OverrideStorage updates the MessageContext, but EfCoreEnvelopeTransaction captured the original IMessageDatabase in its constructor, so it still writes to the main schema.

jeremydmiller:
@Ferchke7 I don't want to take this in and have to support it. I still don't think this is necessary or beneficial in the way you were hoping anyway. I think this is just more complexity.

PR #3052, opened 2026-06-08
Route message types to ancillary durable stores

Ferchke7:
## Why

Wolverine already supports ancillary message stores through provider registrations like:

```csharp
opts.PersistMessagesWithPostgresql(
        connectionString,
        "bulk_outbox",
        role: MessageStoreRole.Ancillary)
    .Enroll<BulkOutboxStore>();
```

But routing specific message types to those stores was not first-class. Users had to combine several lower-level mechanisms:

- set `chain.AncillaryStoreType` through a custom `IHandlerPolicy`
- manually call `MessageContext.OverrideStorage()` before publishing
- cast outbox abstractions like `IDbContextOutbox<T>` back to `MessageContext`
- duplicate that logic across handlers or wrapper services

That works, but it is fragile and easy to miss in high-volume publishing paths.

This PR adds a single policy API for routing selected message types to an enrolled ancillary store for both durable inbox and durable outbox persistence.

## Example

```csharp
public interface BulkOutboxStore;

builder.Host.UseWolverine(opts =>
{
    opts.PersistMessagesWithPostgresql(connectionString, "wolverine");

    opts.PersistMessagesWithPostgresql(
            connectionString,
            "bulk_outbox",
            role: MessageStoreRole.Ancillary)
        .Enroll<BulkOutboxStore>();

    opts.Policies.RouteMessagesToAncillaryStore<BulkOutboxStore>(
        typeof(BulkEvent),
        typeof(AnotherHighVolumeMessage));
});
```

## What this does

For matching message types, the policy:

- sets `HandlerChain.AncillaryStoreType` so durable inbox persistence targets the enrolled ancillary store
- stamps outgoing envelopes with the ancillary message store before durable outbox persistence
- works with normal publishing APIs, including `IMessageBus`, `IMessageContext`, and `IDbContextOutbox<T>`
- removes the need for user code to manually call `OverrideStorage()`

### DIFF
--- a/docs/guide/durability/index.md
+++ b/docs/guide/durability/index.md
@@ -142,6 +142,44 @@ using var host = await Host.CreateDefaultBuilder()
 <sup><a href='https://github.com/JasperFx/wolverine/blob/main/src/Persistence/PersistenceTests/Samples/DocumentationSamples.cs#L52-L61' title='Snippet source file'>snippet source</a> | <a href='#snippet-sample_make_all_subscribers_be_durable' title='Start of snippet'>anchor</a></sup>
 <!-- endSnippet -->
 
+### Routing Message Types to Ancillary Stores <Badge type="tip" text="6.0" />
+
+If one or more message types create a much higher volume of durable outgoing messages than the rest of your
+application, you can isolate those messages into a separate ancillary message store. This gives the selected message
+types their own `wolverine_outgoing_envelopes` table and durability agents instead of making every outgoing message
+compete for the main outbox table.
+
+First register the main message store, then register an ancillary store and enroll it with a marker type. Finally,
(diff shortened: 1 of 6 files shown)


jeremydmiller:
@Ferchke7 Is there any reason why you didn't base any of this off the existing Marten support for this kind of thing? And *what are you doing that you require Wolverine to work completely differently than for everybody else?*. I'm deeply unenthusiastic about supporting any of this stuff going forward until I can hear a better rationale for what you're doing

jeremydmiller:
@Ferchke7 Yeah, sorry, same thing. This is complexity that I don't want to have to support. And you could have based this off the Marten implementation just to be more consistent anyway.

PR #3068, opened 2026-06-11
feat(efcore): add MainDbContext configuration to resolve multiple DbContext conflicts

KhaledZaabat:
### Problem
When a message handler chain depends on multiple `DbContext` types (either directly or transitively via other services), Wolverine's EF Core transaction middleware throws a `System.InvalidOperationException` because it cannot determine which `DbContext` to enroll in the transaction and the Wolverine Outbox:

> `System.InvalidOperationException: Cannot determine the DbContext type for Message Handler for ..., multiple DbContext types detected: AppDbContext, CommunDbContext`

Currently, there is no simple way to resolve this conflict during Wolverine's EF Core configuration.

### Solution
This PR introduces a `UseMainDbContext<TDbContext>()` configuration option that lets you designate a default/main `DbContext` to use when multiple `DbContext` types are detected in a chain. 

If a main `DbContext` is configured, Wolverine will resolve to that type to enroll in transactions/outbox, bypassing the ambiguity exception. If no main `DbContext` is configured, it falls back to the original behavior of throwing the `InvalidOperationException`.

### Usage
You can configure the main `DbContext` in two ways:

1. **Via the fluent transactions configuration:**
   opts.UseEntityFrameworkCoreTransactions()
       .UseMainDbContext<AppDbContext>();

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -155,6 +155,7 @@
     <PackageVersion Include="Microsoft.Extensions.Hosting" Version="10.0.0" />
     <PackageVersion Include="Microsoft.Extensions.Hosting.Abstractions" Version="10.0.0" />
     <PackageVersion Include="Microsoft.Extensions.ObjectPool" Version="10.0.2" />
+    <PackageVersion Include="Microsoft.EntityFrameworkCore.InMemory" Version="10.0.2" />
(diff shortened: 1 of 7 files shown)


jeremydmiller:
@KhaledZaabat Hey, you're going to need to go down a different path. This won't work for folks doing modular monoliths, and I don't think that "main DbContext" is the right concept. I think I'd recommend doing something on a handler by handler basis that declares with DbContext *for that handler or endpoint* is the transactional DbContext.

PR #3073, opened 2026-06-11
Corrected test discovery with flaky retries

dmytro-pryvedeniuk:
This PR changes the way how the tests are executed. 
* Fix. Because of the custom test discovery some tests were ignored (e.g. ones that extend parent test class). Test discovery is fixed.
* Flaky retries. Instead of using different scopes for test runs - per assembly, per class, per method - and retrying whole scope I run all tests per assembly (Round 1) and retry failed ones per method up to 2 times (Round 2). It provides uniform approach for all test projects, improves performance (no need to rerun successful tests) and protects better from flaky tests (existing or new).
* Fix: CosmosDB and AzureServiceBus tests. New discovered tests required fixes to make runs green.
* Marked some tests as flaky to skip for now. E.g. Pulsar compliance or RabbitMq CircuitBreaker tests. We need to address them separately.

Possible improvements:
* For now flaky-during-run tests are just logged but later we can handle them better, e.g. report somehow or auto-create an issue.
* If there are many failing tests, e.g. if DB is not available during run flaky retry approach can be slow. Some restriction can be added, e.g. no more than N failed tests to retry.


jeremydmiller:
@dmytro-pryvedeniuk That's a big win! Thank you.

PR #3082, opened 2026-06-12
Coverage: composite + side-effect publishing under managed distribution + tenant-partitioned events

erdtsieck:
### What

Adds a test for a composite projection whose stage-2 member publishes side-effect messages
(`RaiseSideEffects` -> `slice.PublishMessage`) running under **Wolverine-managed event-subscription
distribution** on a `Conjoined` + `Quick` + `UseTenantPartitionedEvents` store. It asserts the
composite catches up for every managed tenant.

### Why

The existing `tenant_partitioned_distribution_*` tests cover standalone `Snapshot` projections under
managed distribution, but not `CompositeProjectionFor` + side-effect publishing. That exact
combination is the production configuration behind JasperFx/marten#4727 (the composite optimized
rebuild deadlock on `ProjectionUpdateBatch.CurrentMessageBatch`).

### Note on sequencing

The deterministic root-cause reproduction and the fix live in the Marten PR for #4727. This is
**combination coverage** for the Wolverine-managed-distribution side; it passes against released
Marten in this configuration and guards the combination going forward. Best landed alongside / after
the Marten #4727 fix release.

### DIFF
--- a/src/Persistence/MartenTests/Distribution/composite_side_effects_under_managed_distribution.cs
+++ b/src/Persistence/MartenTests/Distribution/composite_side_effects_under_managed_distribution.cs
@@ -1,0 +1,182 @@
+using System;
+using System.Collections.Generic;
+using System.Threading;
+using System.Threading.Tasks;
+using IntegrationTests;
+using JasperFx;
+using JasperFx.CodeGeneration;
+using JasperFx.Core;
+using JasperFx.Events;
+using JasperFx.MultiTenancy;
+using JasperFx.Resources;
+using Marten;
+using Marten.Events;
+using Marten.Events.Aggregation;
+using Marten.Events.Projections;
+using Microsoft.Extensions.DependencyInjection;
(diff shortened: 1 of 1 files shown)


erdtsieck:
## Closing — the premise (a Wolverine-layer composite bug) is disproven

This PR was opened to guard what looked like a Wolverine-hosted-daemon defect (#3083). Root-cause analysis shows there is no Wolverine bug: the composite stall was **Marten #4712** (high-water `Timestamp` under `UseTenantPartitionedEvents`), surfacing only because this repo pins **Marten 9.6.0**, which predates the fix. On **Marten 9.7.3** the same scenario passes in ~4s (details in #3083).

Closing this PR as it no longer reflects an actual Wolverine issue. The coverage test itself stays useful and can be re-introduced alongside a Marten-dependency bump (≥ 9.7.3) if the maintainers want a standing regression guard for composite catch-up under tenant-partitioned managed distribution.

Issue #3083, opened 2026-06-12
Composite projection does not catch up under the Wolverine-hosted daemon with UseTenantPartitionedEvents (Marten's own daemon does)


erdtsieck:
## Summary

A **composite projection** (`Projections.CompositeProjectionFor`) does **not catch up under the Wolverine-hosted projection daemon** when `UseTenantPartitionedEvents = true`, while:

- a **plain `SingleStreamProjection`** in the same store **does** catch up, and
- the **same composite catches up fine under Marten's own daemon** (`store.BuildProjectionDaemonAsync()` + `StartAllAsync()`).

So the composite shard is left stuck at its initial mark under Wolverine-managed/hosted distribution, while everything else advances.

## Isolation (single variable: the daemon host)

Identical Marten config in both — `Conjoined` + `QuickWithServerTimestamps` + `UseTenantPartitionedEvents` + `AllDocumentsAreMultiTenanted`, two managed tenants (`AddMartenManagedTenantsAsync`), one plain projection + one 2-stage `CompositeProjectionFor`, ~10 streams/tenant appended **after** the daemon starts (continuous). Marten 9.7.3 / JasperFx 2.9.9 / WolverineFx 6.8.0, PostgreSQL.

| Daemon | plain projection | composite |
|---|---|---|
| Marten `BuildProjectionDaemonAsync().StartAllAsync()` | catches up | **catches up ✅** |
| Wolverine `IntegrateWithWolverine(m => m.UseWolverineManagedEventSubscriptionDistribution = true)` | catches up | **stuck ❌** |

Observed DB state after the Wolverine run (2 tenants × 10 streams × 3 events = 60 events; per-tenant sequences so `seq_id` 1..30 appears once per tenant → `count=60, distinct=30, max=30`):

```
HighWaterMark        = 30      <- high-water reached the real max(seq_id)
CsPlain:All          = 30      <- plain projection caught up
CsComposite:All      = 1       <- composite stuck
CsTrip:All           = 1
CsTripNotice:All     = 1
```

Daemon log for the composite: `Starting single-pass composite rebuild for CsComposite:All over events 0 to 1` then `Finished ... proceeding to normal, continuous operation`, followed by repeated `High Water agent is stale at 60 ... not able to determine a safe harbor sequence`. The composite never advances in continuous mode even though the high-water and the plain projection reach 30.

[...]

PR #3085, opened 2026-06-12
Bump Marten to 9.7.3 (+ JasperFx 2.9.9) — fix tenant-partitioned composite catch-up under managed distribution

### LINKED ISSUE #3084: Composite projection stalls under UseTenantPartitionedEvents + managed distribution on the pinned Marten 9.6.0 (missing marten#4712)
### Summary

With `UseWolverineManagedEventSubscriptionDistribution = true` and Marten's `UseTenantPartitionedEvents` across multiple managed tenants, a **composite projection never catches up** under the Wolverine-hosted daemon — it stays stuck at its premature `0 → 1` optimized rebuild — while a plain `SingleStreamProjection` in the same store catches up. Marten's own daemon (`StartAllAsync` / `StartAgentAsync`) catches the composite up fine in the identical scenario.

### Root cause

Not a Wolverine distribution defect. This repo's `Directory.Packages.props` pins **Marten 9.6.0**, which predates **[marten#4712](https://github.com/JasperFx/marten/pull/4714)** (high-water `Timestamp` under tenant-partitioned events). On 9.6.0 the store-global `HighWaterStatisticsDetector` left `Timestamp` at `default(DateTimeOffset)` = `0001-01-01`, so `HighWaterAgent` computed `safeHarborTime = 0001-01-01 + StaleSequenceThreshold` and could never determine a safe harbor:

```
Starting single-pass composite rebuild for CsComposite:All over events 0 to 1
Shard 'CsPlain:All': Executed updates for Event range ... 0 to 30
Finished ... CsComposite:All, proceeding to normal, continuous operation
High Water agent is stale at 30 ... skipping gap to events marked after 01/01/0001 00:00:03
Daemon projection high water detection was not able to determine a safe harbor sequence, will try again soon   (repeats forever)
```

The composite finishes its `0 → 1` rebuild, goes continuous, and is never told to advance to the real high-water. The plain projection only caught up because it happened to consume the first good mark before the high-water agent wedged.

### Proof (clean A/B, only the Marten version changed)

| Marten | Result |
|---|---|
| 9.6.0 (current pin) | hang ~90s, composite stuck at 1, `stale … 01/01/0001` |
| 9.7.3 | PASS in ~4s |

Marten 9.7.3 already contains #4712.

### Fix

Bump the pinned Marten to **9.7.3** (and the JasperFx packages to **2.9.9**, which 9.7.3 depends on), plus a regression test that pins composite catch-up under managed distribution + tenant partitioning. The full `wolverine.slnx` builds clean against the bump (0 warnings / 0 errors). PR follows.

Supersedes #3083 (which assumed a Wolverine-layer bug — disproven).

erdtsieck:
## What

Bump the pinned **Marten 9.6.0 → 9.7.3** and **JasperFx 2.8.2 → 2.9.9** (the version 9.7.3 depends on), and add a regression test for composite projection catch-up under `UseWolverineManagedEventSubscriptionDistribution` + Marten `UseTenantPartitionedEvents`.

## Why

On the pinned Marten 9.6.0, a composite projection never catches up under the Wolverine-hosted daemon with tenant-partitioned events — it stays stuck at a premature `0 → 1` optimized rebuild, while a plain `SingleStreamProjection` in the same store catches up and Marten's own daemon handles the composite fine.

Root cause is **not** in Wolverine's distribution layer. Marten 9.6.0 predates **marten#4712** (high-water `Timestamp` under tenant-partitioned events): the store-global `HighWaterAgent` reads a `default(DateTimeOffset)` = `0001-01-01` timestamp, computes a bogus safe-harbor (`0001-01-01 + StaleSequenceThreshold`), and never advances:

```
Starting single-pass composite rebuild for CsComposite:All over events 0 to 1
Shard 'CsPlain:All': Executed updates for Event range ... 0 to 30
Finished ... CsComposite:All, proceeding to normal, continuous operation
High Water agent is stale at 30 ... skipping gap to events marked after 01/01/0001 00:00:03
Daemon projection high water detection was not able to determine a safe harbor sequence, will try again soon   (forever)
```

Clean A/B, only the Marten version changed:

| Marten | Result |
|---|---|
| 9.6.0 (current pin) | hang ~90s, composite stuck at 1, `stale … 01/01/0001` |
| 9.7.3 | PASS in ~4s |

The full `wolverine.slnx` builds clean against the bump (0 warnings / 0 errors).

Closes #3084. Supersedes #3083 (which assumed a Wolverine-layer bug).

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -32,19 +32,19 @@
     <PackageVersion Include="Grpc.StatusProto" Version="2.76.0" />
     <PackageVersion Include="Grpc.Tools" Version="2.76.0" />
     <PackageVersion Include="HtmlTags" Version="9.0.0" />
-    <PackageVersion Include="JasperFx" Version="2.8.2" />
-    <PackageVersion Include="JasperFx.Events" Version="2.8.2" />
-    <PackageVersion Include="JasperFx.Events.SourceGenerator" Version="2.8.2" />
+    <PackageVersion Include="JasperFx" Version="2.9.9" />
+    <PackageVersion Include="JasperFx.Events" Version="2.9.9" />
+    <PackageVersion Include="JasperFx.Events.SourceGenerator" Version="2.9.9" />
     <!-- RuntimeCompiler is on its own 5.x line (the Roslyn compiler package) — not the 2.1.x
          family; it stays at 5.0.0. -->
     <PackageVersion Include="JasperFx.RuntimeCompiler" Version="5.0.0" />
-    <PackageVersion Include="JasperFx.SourceGenerator" Version="2.8.2" />
-    <PackageVersion Include="Marten" Version="9.6.0" />
+    <PackageVersion Include="JasperFx.SourceGenerator" Version="2.9.9" />
+    <PackageVersion Include="Marten" Version="9.7.3" />
     <PackageVersion Include="Microsoft.Data.SqlClient" Version="6.1.3" />
     <PackageVersion Include="Polecat" Version="4.2.1" />
     <PackageVersion Include="Microsoft.Azure.Cosmos" Version="3.46.1" />
-    <PackageVersion Include="Marten.AspNetCore" Version="9.6.0" />
-    <PackageVersion Include="Marten.Newtonsoft" Version="9.6.0" />
+    <PackageVersion Include="Marten.AspNetCore" Version="9.7.3" />
+    <PackageVersion Include="Marten.Newtonsoft" Version="9.7.3" />
     <PackageVersion Include="MemoryPack" Version="1.21.3" />
     <PackageVersion Include="MessagePack" Version="3.1.3" />
     <PackageVersion Include="Meziantou.Extensions.Logging.Xunit" Version="1.0.15" />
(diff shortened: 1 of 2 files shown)


erdtsieck:
### Note on the failing CI checks — unrelated to this bump

The red `build` and `Run Marten Tests` checks fail at **restore**, not in any test:

```
error NU1903: Warning As Error: Package 'MessagePack' 3.1.3 has a known high severity
vulnerability, https://github.com/advisories/GHSA-hv8m-jj95-wg3x
```

This is a newly-published advisory for **MessagePack 3.1.3**, which is pinned repo-wide in `Directory.Packages.props` and is **not touched by this PR**. With `TreatWarningsAsErrors=true` the NuGet audit warning becomes a build error. `main`'s CI was green on 2026-06-11 and would now hit the same NU1903 if re-run, since the audit evaluates live advisory data at restore time — only the jobs that pull in `Wolverine.MessagePack` fail; every job that doesn't passes.

So the failures are orthogonal to the Marten 9.6.0 → 9.7.3 bump. Locally the full `wolverine.slnx` builds clean and the new test passes in ~4s with `-p:NuGetAudit=false`. Resolving the advisory (e.g. MessagePack 3.1.3 → 3.1.7) is a separate, repo-wide change and is intentionally left out of this PR.

jeremydmiller:
Rebased onto current main and merged via #3105 (your authorship preserved via cherry-pick). main was already at JasperFx 2.9.10, so only the Marten 9.6.0 → 9.7.3 bump and your regression test were carried over. Thank you @erdtsieck!

PR #3167, opened 2026-06-19
Map Wolverine TenantId from incoming MassTransit messages

outofrange-consulting:
## What

Adds a `MapTenantIdFrom<T>` hook to the MassTransit interop so applications consuming MassTransit messages can populate Wolverine's `Envelope.TenantId` from each incoming message:

```csharp
opts.ListenToRabbitQueue("orders")
    .UseMassTransitInterop(mt =>
    {
        // from the strongly-typed message body...
        mt.MapTenantIdFrom<OrderPlaced>(env => env.Message?.TenantId);

        // ...or from MassTransit envelope metadata (headers, addresses, ids)
        mt.MapTenantIdFrom<OrderShipped>(env =>
            env.Headers.TryGetValue("tenant-id", out var v) ? v?.ToString() : null);
    });
```

The lambda receives a new public read-only view, `IMassTransitEnvelope<T>`, exposing the deserialized `Message` alongside the surrounding MassTransit metadata. The concrete `MassTransitEnvelope<T>` stays `internal`, so the mutable serialization DTO is not leaked onto the public surface.

Behavior:
- Multiple message types compose — each mapper only fires for its own `T` (re-registering the same `T` replaces the prior mapping).
- A `null`/empty result never overwrites an existing `TenantId`.
- Inbound (deserialization) path only.

## Drive-by fix

While wiring this up I found that `UseMassTransitInterop(configure)` dropped the `configure` lambda on the **Rabbit MQ** and **Amazon SQS** listeners: Rabbit MQ called `InteropWithMassTransit()` with no argument, and the SQS `MassTransitMapper` never received it. As a result any serializer customization — including the existing `UseSystemTextJsonForSerialization`, and now `MapTenantIdFrom` — silently no-opped on those transports while only Azure Service Bus honored it. This is threaded through on both, in its own commit so it can be reviewed (or split) independently.

## Tests (no broker required)

- `MassTransitTenantMappingTests` (CoreTests) — body mapping, header mapping, unregistered-type pass-through, empty-no-overwrite, multi-type dispatch, full-chain fall-through.
- `masstransit_interop_map_tenant_id` (RabbitMQ.Tests) — drives the full deserialization pipeline with a real MassTransit wire payload through `TryDeserializeEnvelope`, proving the `configure` lambda reaches the listener. Uses stubbed transports.
- `masstransit_mapper_tenant_mapping` (AmazonSqs.Tests) — confirms the SQS mapper applies the `configure` lambda.

Docs updated under `docs/guide/messaging/transports/rabbitmq/interoperability.md`. `dotnet build wolverine.slnx -c Release` is clean locally.

### DIFF
--- a/docs/guide/messaging/transports/rabbitmq/interoperability.md
+++ b/docs/guide/messaging/transports/rabbitmq/interoperability.md
@@ -221,8 +221,34 @@ Wolverine = await Host.CreateDefaultBuilder().UseWolverine(opts =>
         // Tell Wolverine to make this endpoint interoperable with MassTransit
         .UseMassTransitInterop(mt =>
         {
-            // optionally customize the inner JSON serialization
+            // optionally customize the inner JSON serialization, or map the Wolverine
+            // tenant id from each incoming MassTransit message (see below)
         })
         .DefaultIncomingMessage<ResponseMessage>().UseForReplies();
 }).StartAsync();
 ```
+
+### Mapping the Tenant Id
+
+When consuming messages from MassTransit, you can derive Wolverine's tenant id for each incoming
+message from either the message body or the MassTransit envelope metadata (headers, addresses,
+correlation ids). Register one or more `MapTenantIdFrom<T>` mappers inside `UseMassTransitInterop`:
+
+```cs
+opts.ListenToRabbitQueue("orders")
+    .UseMassTransitInterop(mt =>
+    {
+        // Pull the tenant id straight off the strongly-typed message body
+        mt.MapTenantIdFrom<OrderPlaced>(env => env.Message?.TenantId);
+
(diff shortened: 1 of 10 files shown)


jeremydmiller:
Thanks @outofrange-consulting! Superseded by #3192, which retains your original commits (and authorship) and adds a small public-API rework on top: rather than introducing a new `IMassTransitEnvelope<T>` interface, it exposes the concrete `MassTransitEnvelope`/`MassTransitEnvelope<T>` type publicly and removes the interface, so `MapTenantIdFrom<T>` takes `Func<MassTransitEnvelope<T>, string?>`.

Closing in favor of #3192 — please follow along there. 🙏

PR #3189, opened 2026-06-22
Stable node identity + lifecycle for storeless Solo hosts (#3188)

### LINKED ISSUE #3188: Lightweight Solo-mode heartbeat/liveness process (stable node identity + lifecycle without the agent controller)
## Summary

In **Solo** mode, node liveness (stable node identity + `NodeStarted`/`NodeStopped` + heartbeats) is welded to the heavyweight `NodeAgentController` path, which is both **store-dependent** and overkill for a single node. As a result a Solo host with **no durable message store** (`NullMessageStore`) gets a random, restart-unstable `AssignedNodeNumber`, never fires `NodeStarted()`/`NodeStopped()`, and emits heartbeats under that churning id.

Proposal: give Solo mode a **lightweight, store-independent liveness process** that establishes a stable identity and lifecycle bookends without spinning up the agent-controller / durable-coordination machinery at all.

## Mechanism today (on `main`)

`StartSoloModeAsync()` is the *only* place a Solo host gets `AssignedNodeNumber = 1` and a `NodeStarted()` — but it does durable work and lives behind the storeless early-return:

```csharp
// src/Wolverine/Runtime/WolverineRuntime.Agents.cs:148
private async Task startAgentsAsync()
{
    if (Storage is NullMessageStore)
    {
        return;                       // <-- storeless Solo bails here, before NodeController exists
    }

    switch (Options.Durability.Mode)
    {
        case DurabilityMode.Solo:
            await startDurableScheduledJobs();
            startNodeAgentController();
            await NodeController!.StartSoloModeAsync();   // sets AssignedNodeNumber=1, fires NodeStarted()
            break;
        // ...
    }
}
```

```csharp
// src/Wolverine/Runtime/Agents/NodeAgentController.StartLocally.cs:9
public async Task StartSoloModeAsync()
{
    await _runtime.Storage.Nodes.ClearAllAsync(_cancellation.Token);   // durable
    await _runtime.Storage.Admin.ReleaseAllOwnershipAsync();           // durable
    var current = WolverineNode.For(_runtime.Options);
    _runtime.Options.Durability.AssignedNodeNumber = current.AssignedNodeNumber = 1;
    await _observer.NodeStarted();
    await startAllAgentsAsync();
    _soloCheckingTask = Task.Run(/* assignment-check polling loop */);
}
```

When the early-return fires (storeless), `NodeController` is never even created, so:
- `AssignedNodeNumber` keeps its constructor default — a random per-process value, different every restart:
  ```csharp
  // src/Wolverine/DurabilitySettings.cs:197
  public int AssignedNodeNumber { get; internal set; } = Guid.NewGuid().ToString().GetDeterministicHashCode();
  ```
- `NodeStarted()` never fires.
- On shutdown, `teardownAgentsAsync()` only calls `NodeStopped()` via the (null) `NodeController`, so `NodeStopped()` never fires either:
  ```csharp
  // src/Wolverine/Runtime/WolverineRuntime.Agents.cs:250
  private async Task teardownAgentsAsync()
  {
      _healthCheckLoop?.SafeDispose();
      if (NodeController != null)            // null for storeless Solo
      {
          var bus = new MessageBus(this);
          await NodeController.StopAsync(bus);   // -> _observer.NodeStopped()
      }
  }
  ```

Meanwhile `HeartbeatBackgroundService` emits regardless (gated only by `Heartbeat.Enabled`), carrying that random id:

```csharp
var heartbeat = new WolverineHeartbeat(
    _runtime.Options.ServiceName ?? string.Empty,
    _runtime.Options.Durability.AssignedNodeNumber,   // GUID-hash for storeless Solo
    now, now - _startedAt);
```

| Host shape | `AssignedNodeNumber` | `NodeStarted()` / `NodeStopped()` |
|---|---|---|
| Solo **+ durable store** | `1` (stable) ✅ | fired ✅ |
| Solo **+ `NullMessageStore`** | **GUID-hash default — random per process** ❌ | never fired ❌ |
| Balanced | assigned via persistence/election | fired |

`AssignedNodeNumber` has an `internal set`, so consumers can't stabilize this themselves — it must be fixed inside Wolverine.

## Why "lightweight" (not just hoist the assignment)

A Solo node is by definition *the* single node: no leader election, no peer rebalancing, no node-table coordination. The durable bits of `StartSoloModeAsync` (`ClearAllAsync`, `ReleaseAllOwnershipAsync`, the assignment-check polling loop) exist to coordinate *against other nodes/storage* — none of which applies to a single Solo node, and none of which is even possible storeless. Liveness shouldn't depend on any of it.

## Proposed design — Option B (lead): dedicated `SoloHeartbeatService`

A small `IHostedService` registered **only when `Durability.Mode == Solo`** that owns the lightweight liveness contract, independent of `NodeAgentController` and of any durable store:

- **On start:** `Options.Durability.AssignedNodeNumber = 1;` then `await Observer.NodeStarted();`
- **On stop:** `await Observer.NodeStopped();`
- Heartbeat emission stays in `HeartbeatBackgroundService` (unchanged) — now riding a stable id `1`.
- Storeless Solo never instantiates `NodeAgentController`, leader/listener families, ownership management, or the assignment-poll loop.
- Stored Solo keeps its existing durable coordination through the current path; we only lift the **identity + lifecycle bookends** into the lightweight service so they no longer sit behind the `NullMessageStore` gate.

**Ordering constraint:** identity must be established in the Solo branch of `WolverineRuntime.HostService` **before** `startMessagingTransportsAsync()` and before the first heartbeat emits, because envelope ownership and the heartbeat both read `AssignedNodeNumber`.

### Alternative — Option A: factor `StartSoloModeAsync`
Split into `establishSoloIdentity()` (store-independent: set `=1`, fire `NodeStarted()`, start local agents — hoisted ahead of the `NullMessageStore` gate) + `coordinateDurableSolo()` (store-dependent: the durable calls + assignment loop). Symmetric teardown fires `NodeStopped()` for storeless Solo too. Reuses the proven path (one code path) but still threads through `NodeAgentController`. Recorded as the fallback if a separate hosted service is judged heavier than folding into the existing method.

## Non-goals
- No change to Balanced behavior.
- No durable `wolverine_nodes` row required for Solo (a single ephemeral node doesn't need one; optional when a store is present).
- Not a change to the heartbeat *emitter* — `HeartbeatBackgroundService` is fine; this is about giving it a stable identity + lifecycle to ride on.

## Acceptance / tests
- Boot a storeless Solo host (`NullMessageStore`), heartbeats enabled:
  - `Durability.AssignedNodeNumber == 1`.
  - Observer receives `NodeStarted()` at boot and `NodeStopped()` on a clean stop.
  - The node id is stable across a stop/start cycle (no GUID-hash churn).
- Solo **+ durable store** retains today's behavior (still node 1, still fires lifecycle, durable coordination unchanged).
- Balanced is unaffected.

## Impact / motivation
Surfaced from CritterWatch monitoring (JasperFx/CritterWatch#510): storeless Solo publishers heartbeat under a random node id that churns every restart, producing phantom node rows that never reconcile with a `NodeStarted` lifecycle record. More broadly, any feature that assumes "Solo == node 1 / a stable node identity" (envelope-ownership debugging, `{NodeId}` logs, per-node dashboards) gets a confusing random integer for storeless Solo today.

## Repro
Boot a `UseWolverine` host with `Durability.Mode = Solo` and **no** durable message store (no Marten/Polecat/Postgres integration), enable heartbeats, and observe `Durability.AssignedNodeNumber` — it is a large random integer that differs across restarts rather than `1`, and the observer never sees `NodeStarted`/`NodeStopped`.

jeremydmiller:
Fixes #3188.

## Problem

A **Solo** host with **no durable message store** (`NullMessageStore`) never runs `StartSoloModeAsync()` — it sits behind the storeless early return in `startAgentsAsync()` (`WolverineRuntime.Agents.cs:148`). `StartSoloModeAsync()` is the *only* place a Solo host gets a stable `AssignedNodeNumber = 1` and a `NodeStarted()` notification, so a storeless Solo host instead keeps the constructor default:

```csharp
// DurabilitySettings.cs
public int AssignedNodeNumber { get; internal set; } = Guid.NewGuid().ToString().GetDeterministicHashCode();
```

…a random value that differs on every process start. Its `WolverineHeartbeat` pings (and any node-keyed signal) carry that churning id, and it never emits `NodeStarted()`/`NodeStopped()`. Downstream monitors (CritterWatch JasperFx/CritterWatch#510) see phantom, restart-churning nodes that never reconcile with a lifecycle record. `AssignedNodeNumber` has an `internal set`, so this has to be fixed inside Wolverine.

| Host shape | `AssignedNodeNumber` | `NodeStarted()`/`NodeStopped()` |
|---|---|---|
| Solo + durable store | `1` ✅ | fired ✅ |
| Solo + `NullMessageStore` | **random per process** ❌ | never fired ❌ |
| Balanced | assigned via election | fired |

## Fix

A lightweight, store-independent liveness path for Solo, split so the timing is correct (identity must be set before transports start; the `NodeStopped()` bookend must fire while transports are still up, since a publish-based observer like CritterWatch emits a record there):

- **Identity (runtime).** The runtime now assigns `AssignedNodeNumber = 1` for a storeless Solo host in `startAgentsAsync()`, before `startMessagingTransportsAsync()` — envelope ownership and the heartbeat both read it.
- **Lifecycle bookends (`SoloHeartbeatService`).** A new `IHostedService` owns `NodeStarted()`/`NodeStopped()` for storeless Solo only. It is registered **after** the runtime hosted service, so `NodeStarted()` fires once transports are up and — because hosted services stop in reverse order — `NodeStopped()` fires **before** the runtime tears the transports down. No-ops unless the host is `Solo` with a `NullMessageStore`.
- **Default observer tolerance.** `PersistenceWolverineObserver.NodeStarted()`/`NodeStopped()` now catch `NotSupportedException` (matching `LostLeadership`), so the default observer doesn't throw when a storeless Solo node fires the bookends against a `NullMessageStore`.

Stored Solo (`StartSoloModeAsync`) and Balanced are **untouched** — the new service no-ops for them, so there is no double-`NodeStarted`/`NodeStopped`.

> Design note: the issue floated a single host-ordered `IHostedService` that also sets identity. That can't satisfy both constraints at once — a service registered *before* the runtime stops *after* it (transports already gone → `NodeStopped` publish fails), and one registered *after* starts *after* the runtime (identity set too late). Splitting identity (runtime, pre-transport) from the lifecycle bookends (service, registered after the runtime) is what makes the timing correct. Confirmed with the issue author before implementing.

## Tests

`solo_storeless_node_identity` (CoreTests, no infra):
- `storeless_solo_gets_node_1_and_fires_lifecycle_bookends` — `AssignedNodeNumber == 1`, `NodeStarted()` once at boot, `NodeStopped()` once on clean stop.
- `storeless_solo_node_number_is_stable_across_restart` — two boot/stop cycles both yield `1` (no GUID-hash churn).
- `does_not_fire_for_a_storeless_non_solo_host` — a storeless Balanced host gets neither bookend.

The two positive tests **fail without the fix** (random node number; bookends never fire) and pass with it. Full CoreTests suite green.

### DIFF
--- a/src/Testing/CoreTests/Runtime/Heartbeat/solo_storeless_node_identity.cs
+++ b/src/Testing/CoreTests/Runtime/Heartbeat/solo_storeless_node_identity.cs
@@ -1,0 +1,87 @@
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Hosting;
+using NSubstitute;
+using Shouldly;
+using Wolverine;
+using Wolverine.Runtime;
+using Wolverine.Runtime.Agents;
+using Xunit;
+
+namespace CoreTests.Runtime.Heartbeat;
+
+/// <summary>
+/// #3188 — a Solo host with no durable message store (NullMessageStore) never runs the
+/// NodeAgentController path, so before the fix it kept a random per-process AssignedNodeNumber
+/// and never emitted NodeStarted()/NodeStopped(). The stable identity is now assigned by the
+/// runtime before transports start, and SoloHeartbeatService owns the lifecycle bookends.
+/// </summary>
+public class solo_storeless_node_identity
+{
+    private static IHost BuildHost(Action<WolverineOptions> configure)
+    {
+        return Host.CreateDefaultBuilder()
+            .UseWolverine(configure)
+            .Build();
+    }
+
+    [Fact]
+    public async Task storeless_solo_gets_node_1_and_fires_lifecycle_bookends()
+    {
+        using var host = BuildHost(opts => opts.Durability.Mode = DurabilityMode.Solo);
+
+        var runtime = host.Services.GetRequiredService<IWolverineRuntime>();
+        var observer = Substitute.For<IWolverineObserver>();
+        runtime.Observer = observer;
+
+        await host.StartAsync();
+
+        // Stable identity, set before the messaging transports start.
+        runtime.Options.Durability.AssignedNodeNumber.ShouldBe(1);
+
+        // NodeStarted fires once, while transports are up; NodeStopped not yet.
+        await observer.Received(1).NodeStarted();
+        await observer.DidNotReceive().NodeStopped();
+
+        await host.StopAsync();
+
+        await observer.Received(1).NodeStopped();
+    }
+
+    [Fact]
+    public async Task storeless_solo_node_number_is_stable_across_restart()
+    {
+        async Task<int> bootAndReadNodeNumber()
+        {
+            using var host = BuildHost(opts => opts.Durability.Mode = DurabilityMode.Solo);
+            await host.StartAsync();
+            var number = host.Services.GetRequiredService<IWolverineRuntime>()
+                .Options.Durability.AssignedNodeNumber;
+            await host.StopAsync();
+            return number;
+        }
+
+        var first = await bootAndReadNodeNumber();
+        var second = await bootAndReadNodeNumber();
+
+        first.ShouldBe(1);
+        second.ShouldBe(1);
+    }
+
+    [Fact]
+    public async Task does_not_fire_for_a_storeless_non_solo_host()
+    {
+        // Default mode is Balanced; a storeless Balanced host has no NodeAgentController and must
+        // NOT get the Solo lifecycle bookends.
+        using var host = BuildHost(_ => { });
+
+        var runtime = host.Services.GetRequiredService<IWolverineRuntime>();
+        var observer = Substitute.For<IWolverineObserver>();
+        runtime.Observer = observer;
+
+        await host.StartAsync();
+        await host.StopAsync();
+
+        await observer.DidNotReceive().NodeStarted();
+        await observer.DidNotReceive().NodeStopped();
+    }
+}
--- a/src/Transports/AWS/Wolverine.AmazonSqs/Internal/AmazonSqsTransport.cs
+++ b/src/Transports/AWS/Wolverine.AmazonSqs/Internal/AmazonSqsTransport.cs
@@ -166,8 +166,17 @@ public class AmazonSqsTransport : BrokerTransport<AmazonSqsQueue>
         // and SQS queue names are case-sensitive. Without this, the sender creates
         // "wolverine-response-MyApp-123" but the receiver resolves the reply URI
         // to "wolverine-response-myapp-123" (lowercased by Uri), creating a different queue.
+        // The per-node response queue must be unique to this running node. In Solo mode the assigned
+        // node number is always 1 (#3188), and the service name is not unique per host, so multiple
+        // Solo hosts (e.g. the request/reply compliance sender + receiver, or successive fixtures on
+        // one broker) would share one response queue and cross-deliver each other's replies. Use the
+        // always-unique UniqueNodeId in Solo; Balanced gets a unique AssignedNodeNumber via election.
+        // See #3189.
+        var responseNode = runtime.Options.Durability.Mode == DurabilityMode.Solo
+            ? runtime.Options.UniqueNodeId.ToString("N")
+            : runtime.DurabilitySettings.AssignedNodeNumber.ToString();
         var responseName = SanitizeSqsName(
-            $"wolverine.response.{runtime.Options.ServiceName}.{runtime.DurabilitySettings.AssignedNodeNumber}")
+            $"wolverine.response.{runtime.Options.ServiceName}.{responseNode}")
             .ToLowerInvariant();
 
         var queue = Queues[responseName];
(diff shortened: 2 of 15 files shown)


jeremydmiller:
Pushed reply-routing regression fixes uncovered while validating this change.

The #3188 "Solo == node 1" change means every storeless Solo host gets `AssignedNodeNumber = 1`. Transports that key their per-node reply endpoint on the node number **with no service discriminator** then collide when two Solo services share a broker, cross-delivering each other's replies (request/reply compliance: "Expected one message … but detected 2"). Fixed in Solo mode by keying on the always-unique `UniqueNodeId` (`"N"` format — dash-free hex, safe for subject/topic names); Balanced is unchanged (its `AssignedNodeNumber` is unique via election):

- **NATS** — `NatsTransport` (validated against broker: request/reply 12/12, full suite 121/121)
- **MQTT** — `MqttTransport` + updated `has_response_topic_automatically` assertions (validated: request/reply 12/12)
- **GCP Pub/Sub** — `PubsubTransport` (reproduced against the emulator: 7 request/reply failures → 16/16 green after fix)

Redis / Azure Service Bus / AWS SQS already namespace their response endpoint by `ServiceName`, so they were unaffected.

GCP Pub/Sub has **no CI job**, which is why its collision wasn't caught automatically — tracked in #3191.

Issue #3280, opened 2026-06-30
Managed event-subscription distribution: no per-tenant agents under MultiTenantedWithShardedDatabases + UseTenantPartitionedEvents (Phase 3) → co-located tenants share one high-water, lagging tenant's events skipped


erdtsieck:
### Summary

Under **`UseWolverineManagedEventSubscriptionDistribution = true`** with a Marten store that uses **`MultiTenantedWithShardedDatabases` + `Events.UseTenantPartitionedEvents = true`**, async projection agents are distributed **per (shard × database)**, never **per tenant**. When two or more tenants are co-located on the same shard database, they each draw `seq_id` from their **own** `mt_events_sequence_<tenant>` (independent, overlapping ranges starting at 1), but a single store-global `<Projection>:All` agent + a single store-global `HighWaterMark` row tries to span all of them. Once the global high-water advances to the busier tenant's height, the lagging tenant's later appends fall **below** it and are **silently skipped** by the async daemon → projection data loss for the lagging co-located tenant.

This is the explicitly-deferred **"Phase 3"** of the per-tenant-partitioned-events distribution matrix (#3021). It's documented as a known caveat in `tenant_partitioned_distribution_granularity.cs`:

> Wolverine-managed event-subscription agents are distributed per (shard × DATABASE), NOT per tenant. … ("One agent per tenant" is only true under sharded databases, **which is Phase 3**.)

### Root cause (code path)

- `Wolverine/Runtime/Agents/EventStoreAgents.cs` → `SupportedAgentsAsync` enumerates agent URIs from `usage.Subscriptions.Where(Async).SelectMany(x => x.ShardNames)` — i.e. the **store-global** shard names (`<proj>:All`), one per (shard × database). No per-tenant expansion.
- The per-tenant continuous-agent fan-out **does** exist, but only in the **non-managed** path: `JasperFx.Events/Daemon/JasperFxAsyncDaemon.cs` → `StartAllAsync` → `buildPerTenantContinuousAgents` (gated on `detector.SupportsTenantPartitioning` + `Database is ICrossTenantRebuildSource`, enumerating tenants via `ICrossTenantRebuildSource.FindRebuildTenantsAsync` = Marten's `mt_tenant_partitions`). Under managed distribution this fan-out is bypassed, because Wolverine assigns/starts agents individually from `SupportedAgentsAsync`.
[...]

PR #3281, opened 2026-06-30
Managed distribution: per-tenant agents + database-affine assignment for sharded stores (#3280, marten#4806)

erdtsieck:
Wolverine piece of #3280 and JasperFx/marten#4806. **Depends on JasperFx/jasperfx#482** and pairs with JasperFx/marten#4799.

### Per-tenant agents (#3280)
`EventStoreAgents.SupportedAgentsAsync` fans out one agent per (shard, tenant) when the store reports `DistributesAgentsPerTenant` — fixing sharded + tenant-partitioned stores where co-located tenants shared one high-water mark and a lagging tenant's events were skipped. Store-global and database-per-tenant stores keep one agent per (shard, database), unchanged.

### Database-affine assignment (marten#4806, opt-in)
With per-tenant agents across many shard databases, the default even spread makes every node open connection pools to (nearly) every database — pools grow as nodes × databases and exhaust a shared server's `max_connections` (observed in production-scale testing: Postgres 53300 at ~9% DB CPU — the ceiling was connections, not hardware). When a store reports `GroupAgentAssignmentsByDatabase`:

- New `AssignmentGrid.DistributeByGroupAffinity(scheme, groupKey)`: every agent sharing a group key (the `[store type]/[store name]/[database]` prefix of the agent Uri) is placed whole on the least-loaded node, largest group first, with deterministic tie-breaks — so a node only connects to the databases it owns and the connection budget scales with the topology (databases), not the deployment (replicas).
- `EventSubscriptionAgentFamily.EvaluateAssignmentsAsync` uses it when any registered store opts in; otherwise the existing `DistributeEvenlyWithBlueGreenSemantics` path is untouched. (The affine path does not apply blue/green capability matching — called out in code + docs.)
- Convenience: `IntegrateWithWolverine(m => m.UseDatabaseAffineAgentAssignment = true)` flows through to `StoreOptions.Events.UseDatabaseAffineAgentAssignment` for the main store.

### Tests & docs
Unit tests for the per-tenant fan-out, grid affinity (grouping, spreading, balancing around a heavy database), family branch selection, the Uri group key, and the IntegrateWithWolverine bridge; the distribution guide documents the opt-in.

**Merge order:** jasperfx#482 → Marten #4799 → this.

### DIFF
--- a/docs/guide/durability/marten/distribution.md
+++ b/docs/guide/durability/marten/distribution.md
@@ -59,6 +59,45 @@ Some other facts about this integration:
   new versions of a projection or subscription on some application nodes in order to do try ["blue/green deployment."](https://en.wikipedia.org/wiki/Blue%E2%80%93green_deployment)
 * This capability does depend on Wolverine's built-in [leadership election](https://en.wikipedia.org/wiki/Leader_election) -- which fortunately got a _lot_ better in Wolverine 3.0
 
+## Database-Affine Distribution for Sharded Stores <Badge type="tip" text="6.x" />
+
+By default Wolverine spreads subscription and projection agents *evenly* across the cluster (with blue/green
+capability matching). That is the right choice for a store-global event store or database-per-tenant multi-tenancy,
+where there is one agent per database anyway.
+
+It is *not* the right choice for a Marten store that combines [sharded multi-tenancy](https://martendb.io/configuration/multitenancy.html#sharded-multi-tenancy-with-database-pooling)
+with [per-tenant event partitioning](https://martendb.io/events/multitenancy.html#per-tenant-event-partitioning). There,
+many tenants are co-located in one shard database and each draws its own event sequence, so Wolverine fans agents out
+one-per-`(shard, tenant)` rather than one-per-database. With hundreds of tenants scattered across many shard databases,
+an even per-agent spread makes *every* node open a connection pool to *nearly every* shard database — so the pool count
+grows as `nodes × databases` and quickly exhausts a shared server's `max_connections`.
+
+To bound that, opt into **database-affine assignment** on the Marten side. Wolverine then keeps every agent for a given
+shard database together on a single node, so a node only opens pools to the databases it actually owns and the pool
+count scales with the number of shard databases, not `nodes × databases`:
+
+```cs
+services.AddMarten(opts =>
+{
+    // ... sharded tenancy + per-tenant event partitioning ...
+    opts.Events.UseTenantPartitionedEvents = true;
+
+    // Keep a shard database's per-tenant agents together on one node
+    opts.Events.UseDatabaseAffineAgentAssignment = true;
+
+    // The "mix": allow a single shard database's agents to fan out across up to N nodes
+    // for more parallelism (default 1 = strict affinity). Server-side connections for one
+    // database then stay bounded at roughly N × per-node pool size.
+    opts.Events.DatabaseAffineAgentFanout = 2;
+})
+.IntegrateWithWolverine();
+```
+
+Wolverine reads these off the store through `IEventStore.GroupAgentAssignmentsByDatabase` and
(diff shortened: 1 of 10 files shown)


erdtsieck:
Addressed both Copilot findings in a38bce882:

1. **Tenant-id normalization** — the per-tenant fan-out now dedupes tenant ids with `StringComparer.OrdinalIgnoreCase` (consistent with `TryResolveTenantDatabaseIdAsync`) and skips blank/whitespace ids, so a malformed usage descriptor can't produce duplicate or invalid per-tenant agents. Covered by a new unit test (`dedupes_tenants_case_insensitively_and_skips_blank_ids`).

2. **The #3280 integration test** — new `sharded_tenant_partitioned_distribution`: `MultiTenantedWithShardedDatabases` + `UseTenantPartitionedEvents` with two tenants co-located on one shard database under managed distribution, asserting one `:all/<tenant>` agent per tenant and that both tenants' events project against their own high-water marks (the exact failure mode of #3280). It runs green locally against the companion jasperfx#482 + marten#4799 builds and will go green in CI once those merge, per the stated merge order.

One observation from writing the test, noted there as a known transition edge: when a database gains its *first* tenants while the host is running, the store-global agent from the empty-database phase keeps running alongside the new per-tenant agents until assignments reconcile. The test provisions tenants before host start; retiring the stale agent on that transition may deserve a small follow-up.

jeremydmiller:
Thank you for this work @erdtsieck — the per-(shard, tenant) fan-out and the test coverage here are solid, and we're going to reuse them. Closing this PR only because the design underneath it is shifting on the JasperFx side and it's cleaner to restart than rebase:

- jasperfx#482 is being revised to **drop `IEventStore.GroupAgentAssignmentsByDatabase` entirely**. Database-affine agent grouping will not be an opt-in flag: Wolverine's managed distribution will apply it **always, per store, for any event store reporting a multi-database `DatabaseCardinality`** (`StaticMultiple`/`DynamicMultiple`), while single-database stores keep today's even + blue/green distribution. That removes the app-wide flag semantics and the Marten/Wolverine configuration surface (`UseDatabaseAffineAgentAssignment`) altogether — less for users to configure, and it aligns Wolverine's placement with what Marten's native `MultiTenantedProjectionDistributor` already does (whole database per node).
- The successor PR will also address the runtime first-tenant transition (retiring the store-global agent once per-tenant agents fan out) and add the affine end-to-end + failover coverage.

The `EventStoreAgents` fan-out and the test suites from this branch will be carried into the successor with authorship retained. Tracked under the per-tenant partitioning epic: JasperFx/jasperfx#486.

erdtsieck:
One addition for the successor PR, from running this at production scale (500 tenants / 512 sharded databases, field notes on jasperfx#486): a **complement** to the "retire stale store-global agents on the first-tenant transition" bullet in the close-out.

The transition edge has two sides, and never-creating is cleaner than retiring-after-the-fact:

- **Retire** (already noted): a store-global agent that was started while a database was empty must be stopped once per-tenant agents appear.
- **Never create** (this change): under `DistributesAgentsPerTenant`, an empty database should yield **no agent at all** — not a store-global one. With per-tenant event partitioning every event lives in a tenant partition, so an empty database has nothing to process, and the store-global agent started for it is exactly the one that then lingers and double-processes next to the per-tenant agents once tenants arrive (both materialize the same per-tenant progression rows → sustained 23505 storms — we hit this hard during provisioning). If it's never created, there's nothing to retire; the tenant's agent simply appears on the next `SupportedAgentsAsync` evaluation.

Isolated commit (fan-out change + unit test that asserts an empty per-tenant database yields zero agents): erdtsieck/wolverine@dfeccbe97 on branch `feat/per-tenant-empty-db-no-agents`. Depends on the `DistributesAgentsPerTenant` surface, so it slots into the successor rather than standing alone. Reuse freely — authorship isn't the point, just flagging the edge so it doesn't get lost.

PR #3284, opened 2026-07-02
Allow [Transactional] to select the DbContext for multi-DbContext han…

KhaledZaabat:
# Allow `[Transactional]` to select the DbContext for multi-DbContext handlers

## Problem

A Wolverine handler chain can only have one *transactional* DbContext. But a
handler is often perfectly valid with **two** DbContext-shaped dependencies —
one it writes through (must be enrolled in the transaction + outbox), and
one it only reads from (e.g. a shared read-only lookup database, or a
different bounded context/module in a modular monolith).

Today, `EFCorePersistenceFrameProvider.DetermineDbContextType` throws the
moment it sees more than one DbContext-shaped dependency on a chain:

```
Cannot determine the DbContext type for ..., multiple DbContext types
detected: AppDbContext, CommonDbContext
```

There was no way to tell Wolverine which one is "the" transactional context
for that handler.

## Solution

`[Transactional]` gains an optional `DbContextType`:

```csharp
public class GrantAccessHandler
{
    [Transactional(typeof(AppDbContext))]
    public static void Handle(GrantAccess message, AppDbContext users, CommonDbContext lookup)
    {
        // users.SaveChanges() is enrolled in the transaction + outbox.
        // lookup is just an ordinary injected read-only dependency.
    }
}
```

- Only needed when a chain has more than one DbContext-shaped dependency —
  single-DbContext handlers are unaffected.
- The type can also be a DbContext **abstraction** registered via
  `WithDbContextAbstraction<TAbstraction, TDbContext>()`, so Clean
  Architecture / Application-layer handlers never need to reference the
  concrete EF Core type at all — see `USAGE.md`.
- A tag that doesn't match any of the chain's actual dependencies fails
  loudly at startup, naming the offending type, rather than silently
  picking a default.

## Changes

- `Wolverine/Attributes/TransactionalAttribute.cs` — new `DbContextType`
  property + `Type`-based constructor, stored as a chain tag (same pattern
  already used for `Mode`).
- `Wolverine.EntityFrameworkCore/Codegen/EFCorePersistenceFrameProvider.cs`
  — `DetermineDbContextType` resolves that tag before the ambiguity check,
  including resolving through registered DbContext abstractions.
- `EfCoreTests/transactional_dbcontext_selection_tests.cs` — coverage for:
  disambiguation with a concrete type, disambiguation through a registered
  abstraction, and the mismatched-tag error path.

## Backward compatibility

Fully additive. Existing single-DbContext handlers and existing
`[Transactional(Mode = ...)]` usages are unaffected; the ambiguity error
for un-tagged multi-DbContext chains is unchanged.

### DIFF
--- a/docs/guide/durability/efcore/transactional-middleware.md
+++ b/docs/guide/durability/efcore/transactional-middleware.md
@@ -384,3 +384,103 @@ opts.UseEntityFrameworkCoreTransactions()
 ```
 <sup><a href='https://github.com/JasperFx/wolverine/blob/main/src/Persistence/EfCoreTests/dbContext_abstraction_scenarios.cs#L62-L83' title='Snippet source file'>snippet source</a> | <a href='#snippet-sample_register_mixed_dbcontexts' title='Start of snippet'>anchor</a></sup>
 <!-- endSnippet -->
+
+## Selecting the Transactional DbContext <Badge type="tip" text="6.17" />
+
+A handler chain can only have **one** transactional `DbContext` — the one Wolverine enrolls in the
+transaction and uses for the outbox. But a handler is often legitimately given more than one
+`DbContext`-shaped dependency: one it writes through, plus one it only *reads* from — a shared
+read-only lookup database, or another module's context in a modular monolith.
+
+When a chain depends on more than one `DbContext`-shaped service, Wolverine resolves which one is
+transactional in this order:
+
+1. A `[Transactional(typeof(TDbContext))]` attribute on the handler (most specific).
+2. A registration-side selection made with `WithTransactionalDbContext<T>()`.
+3. **Automatic**: the single [Wolverine-enabled](#automatic-selection) candidate, if there is exactly one.
+
+If none of these resolves to a single context, Wolverine fails fast at startup and names the
+candidates.
+
+### Automatic selection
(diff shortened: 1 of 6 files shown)


jeremydmiller:
@KhaledZaabat Some of what you wrote for the documentation isn't right. Hey, I think I'm going to go for a superseding pull request based on this, but I'll leave you as a co-author and reuse the tests

jeremydmiller:
Thanks @KhaledZaabat — really solid work here, and the core capability is being merged. I've opened #3295 which **supersedes** this PR. It keeps your `[Transactional(typeof(TDbContext))]` mechanism and tests (and preserves you as a co-author on the commit), with two changes from review:

- Reworked to be **explicit-only, no "magic"**: dropped the registration-side `WithTransactionalDbContext<T>()` selection and the automatic "single Wolverine-enabled candidate" inference. When a chain has more than one DbContext and no designation, Wolverine now fails fast and tells the user to either drop the automatic transactional middleware or designate the context.
- Added `[Storage(typeof(TDbContext))]` as an equivalent, provider-agnostic way to designate the transactional DbContext.

Closing in favor of #3295. Much appreciated! 🙏

PR #3326, opened 2026-07-06
fix: CircuitWatcher.Dispose() does not stop the ping-until-reconnected loop

lahma:
## Summary

`CircuitWatcher` only stores the caller's `CancellationToken` and its `Dispose()` just disposes the `Task` wrapper -- neither actually stops `pingUntilConnectedAsync`, which keeps running against the still-live caller token. `SendingAgent`'s own `DisposeAsync()` never disposed the circuit watcher either, so once a sender's circuit trips, it keeps pinging a dead destination forever, even after the owning host/agent has been disposed.

I hit this concretely with the Kafka transport: a test host pointed at a permanently unreachable broker (used deliberately to exercise startup config validation) never exited, because the producer's circuit-breaker ping loop kept retrying in the background indefinitely regardless of how fast/slow each connection attempt failed -- disposing the `WebApplicationFactory` did not stop it. The same gap applies to any transport that goes through `SendingAgent`'s circuit-breaker path, not just Kafka.

## Fix

- `CircuitWatcher` now links the caller's token into its own `CancellationTokenSource` and cancels that source in `Dispose()`, so `Dispose()` can actually stop the loop instead of only releasing the `Task` wrapper.
- `SendingAgent.DisposeAsync()` now disposes the circuit watcher (mirroring the existing cleanup already done in `MarkSuccessAsync()`) before tearing down the sender.

## Test plan

- [x] Added `circuit_watcher_dispose_stops_the_ping_loop` (isolated `CircuitWatcher` + `StubCircuit`) and `disposing_a_sending_agent_stops_its_circuit_watcher` (end-to-end via `BufferedSendingAgent`) to `CoreTests/Transports/Sending/`.
- [x] Verified both new tests fail without the fix (ping count keeps climbing after `Dispose()`/`DisposeAsync()`) and pass with it.
- [x] `dotnet build src/Wolverine/Wolverine.csproj` and `dotnet build src/Testing/CoreTests/CoreTests.csproj`: clean.
- [x] Full `CoreTests` suite: 1856 passed, 4 pre-existing failures unrelated to this change (confirmed identical failures on unmodified `main`: `Bug_2896_mixed_lifetime_enumerable_dependency` and 3 `result_types_end_to_end` tests, all "Cannot resolve scoped service ... from root provider" -- a DI scoping issue unrelated to sending/circuit-breaker code).

### DIFF
--- a/src/Testing/CoreTests/Transports/Sending/CircuitWatcherTester.cs
+++ b/src/Testing/CoreTests/Transports/Sending/CircuitWatcherTester.cs
@@ -42,9 +42,11 @@ public class StubCircuit : ISenderCircuit
 
     public bool SupportsNativeScheduledSend => true;
 
+    public int CallCount => _count;
+
     public Task<bool> TryToResumeAsync(CancellationToken cancellationToken)
     {
-        _count++;
+        Interlocked.Increment(ref _count);
 
         if (_count < _failureCount)
         {
@@ -60,7 +62,7 @@ public class StubCircuit : ISenderCircuit
         return Task.CompletedTask;
     }
 
-    public TimeSpan RetryInterval { get; } = 50.Milliseconds();
+    public TimeSpan RetryInterval { get; set; } = 50.Milliseconds();
 
     public void Dispose()
     {
--- a/src/Testing/CoreTests/Transports/Sending/SendingAgentDisposalTests.cs
+++ b/src/Testing/CoreTests/Transports/Sending/SendingAgentDisposalTests.cs
@@ -23,6 +23,11 @@ public class SendingAgentDisposalTests
 
         watcher.Dispose();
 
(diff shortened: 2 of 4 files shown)


lahma:
Good catches, thanks — pushed a fix for both:
- `pingUntilConnectedAsync` now wraps the whole loop in a try/catch, not just the inner ping call, so `OperationCanceledException`/`ObjectDisposedException` from the timer wait or a token-registration race during `Dispose()` is swallowed as an expected shutdown rather than surfacing as a faulted/unobserved task.
- Both new tests now settle for one retry interval after `Dispose()`/`DisposeAsync()` before capturing the baseline count, so an already-in-flight ping can't tick over during the assertion window and cause a spurious failure. Ran both 5x locally with no flakes after the change.

PR #3329, opened 2026-07-06
Suppress misleading 'async daemon is disabled' warning under Wolverine-managed event subscription distribution (#3290)

### LINKED ISSUE #3290: Misleading 'async daemon is disabled' warning with UseWolverineManagedEventSubscriptionDistribution = true
### Summary

When a Marten store is integrated with Wolverine and configured with
`opts.UseWolverineManagedEventSubscriptionDistribution = true`, Marten logs a
`The async daemon is disabled` warning at startup for **every** async projection —
even though the projections *do* run under Wolverine's managed event-subscription
distribution.

The warning is misleading: it states the projections "will not be executed", which is
false in this configuration, and it reads like a misconfiguration to operators.

### Observed

At startup, one warning per projection group:

```
Warning: The async daemon is disabled.
Projections Marten.Events.Projections.CompositeProjection, ...,
Application.Journaling.InvoiceJournalEntriesProjection, ...
will not be executed without the async daemon enabled
```

The async projections are in fact executed (read models are built, per-tenant
progression advances), so the daemon is clearly running — it's just owned/started by
Wolverine's managed distribution rather than Marten's own `AddAsyncDaemon`.

### Expected

When Wolverine manages the event-subscription distribution (and therefore runs the
async projections), Marten should not emit "the async daemon is disabled / projections
will not be executed". Either configure Marten's daemon state so the warning is not
raised, or suppress this specific warning when
`UseWolverineManagedEventSubscriptionDistribution = true`.

### Configuration

```csharp
services.AddMarten(/* ... async projections registered ... */)
    .IntegrateWithWolverine(opts =>
    {
        opts.MainDatabaseConnectionString = referencesConnectionString;
        opts.UseWolverineManagedEventSubscriptionDistribution = true;
    });
```

### Versions

- Wolverine 6.16.x
- Marten 9.12.x
- .NET 10

### Notes

The warning appears consistently whenever `UseWolverineManagedEventSubscriptionDistribution = true`
is used with async projections. It is not blocking — projections run fine — but it is a
confusing operational signal.

jeremydmiller:
Closes #3290

## Root cause

Marten's `DocumentStore` constructor writes this console warning whenever async projections are registered but `StoreOptions.Projections.AsyncMode == DaemonMode.Disabled`:

```
Warning: The async daemon is disabled.
Projections ... will not be executed without the async daemon enabled
```

`UseWolverineManagedEventSubscriptionDistribution = true` deliberately replaces Marten's `AddAsyncDaemon(DaemonMode.HotCold)` registration with `WolverineProjectionCoordinator` — but Marten's only knowledge of the daemon state is `Projections.AsyncMode`, which the integration never set. So it stayed `Disabled`, the warning fired for every store, and it was flatly wrong: the async projections DO run under Wolverine's managed distribution.

## Fix

`MartenOverrides` — the `IConfigureMarten` hook the integration already registers for the main store and (via `MartenOverrides<T>`) every ancillary store — now records the real daemon state: when the resolved `MartenIntegration` has `UseWolverineManagedEventSubscriptionDistribution` enabled and `AsyncMode` is still `Disabled`, it sets `AsyncMode = DaemonMode.HotCold`, exactly what the `AddAsyncDaemon(HotCold)` call it replaces would have recorded.

Deliberately narrow:

- Only upgrades from `Disabled` — an explicit user `AddAsyncDaemon()` choice is never overwritten (regardless of call order relative to `IntegrateWithWolverine`).
- `Projections.AsyncMode` is purely descriptive here: Marten's own `ProjectionCoordinator` hosted service is only registered by `AddAsyncDaemon()` itself, so no Marten-side daemon gets started by this — the only Marten readers of the value are the warning gate and the coordinator `Mode` properties.
- Stores without managed distribution warn exactly as before.

## Tests

`Bug_3290_misleading_daemon_disabled_warning` (MartenTests) captures `Console.Out` around `DocumentStore` construction (the warning is a `Console.WriteLine`, not `ILogger`) and asserts:

- managed distribution → `AsyncMode == HotCold`, no "The async daemon is disabled" output
- no managed distribution → `AsyncMode == Disabled`, warning still emitted (behavior preserved)
- managed distribution + explicit `AddAsyncDaemon(Solo)` → stays `Solo`

Verified locally: new tests green on net9.0 + net10.0; `MartenTests.Distribution` suite (26 tests — the managed-distribution mechanics that now see `HotCold`) and `AncillaryStores` suite (19 tests) green on net10.0.

### DIFF
--- a/src/Persistence/MartenTests/Bugs/Bug_3290_misleading_daemon_disabled_warning.cs
+++ b/src/Persistence/MartenTests/Bugs/Bug_3290_misleading_daemon_disabled_warning.cs
@@ -1,0 +1,103 @@
+using IntegrationTests;
+using JasperFx.Core.Reflection;
+using JasperFx.Events.Daemon;
+using JasperFx.Events.Projections;
+using Marten;
+using MartenTests.Distribution.TripDomain;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Hosting;
+using Shouldly;
+using Wolverine;
+using Wolverine.Marten;
+
+namespace MartenTests.Bugs;
+
+// GH-3290: with UseWolverineManagedEventSubscriptionDistribution = true, Marten's
+// DocumentStore wrote a misleading "The async daemon is disabled ... will not be
+// executed" console warning at startup even though Wolverine's managed distribution
+// does run the async projections. The Marten integration now reflects the real daemon
+// state on StoreOptions.Projections.AsyncMode so the warning is not raised.
+public class Bug_3290_misleading_daemon_disabled_warning : PostgresqlContext
+{
+    private static IHostBuilder configureHost(bool useWolverineManagedDistribution,
+        DaemonMode? explicitDaemonMode = null)
+    {
+        return Host.CreateDefaultBuilder()
+            .UseWolverine(opts =>
+            {
+                var marten = opts.Services.AddMarten(m =>
+                    {
+                        m.DisableNpgsqlLogging = true;
+                        m.Connection(Servers.PostgresConnectionString);
+                        m.DatabaseSchemaName = "bug3290";
+
+                        m.Projections.Add<TripProjection>(ProjectionLifecycle.Async);
+                    })
+                    .IntegrateWithWolverine(m =>
+                    {
+                        m.UseWolverineManagedEventSubscriptionDistribution = useWolverineManagedDistribution;
+                    });
+
+                if (explicitDaemonMode.HasValue)
+                {
+                    marten.AddAsyncDaemon(explicitDaemonMode.Value);
+                }
+            });
+    }
+
+    // Marten writes the async-daemon warning with Console.WriteLine inside the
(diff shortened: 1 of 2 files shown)


jeremydmiller:
Closing per review: setting `AsyncMode = HotCold` from the integration is wrong — it records a mode that turns Marten's own daemon coordination on in parallel with Wolverine's managed distribution.

Correct design instead: a fourth `DaemonMode` enum value, **`ExternallyManaged`** — no warnings from the store, no store-hosted coordination, and Wolverine's `MartenIntegration` sets it explicitly (same for the Polecat integration, which most likely has the same issue). Chain: JasperFx enum + core handling → Marten (warning gate + coordinator registration) → Wolverine integration sets it → Polecat parity. The test coverage from this PR (console capture, explicit-user-choice preservation) carries into the redo. Tracked under JasperFx/jasperfx#486 / wolverine#3290.

PR #3333, opened 2026-07-07
fix(shared-memory): copy envelopes at transport handoff

### LINKED ISSUE #3015: Problems w/ the shared memory transport after introducing envelope pooling
## Summary

Confirming **6.4.1 fixes exactly the two paths called out in #3013** (BufferedReceiver retry blocks + MoveToErrorQueue.ExecuteAsync line 33). Thank you for the fast turnaround.

However, on the same CritterWatch integration sweep, the test still fails with the same shape — the cascade just exposes **three further unguarded `!` derefs** that the outer NREs were masking. The root issue is that the all-zero `#00000000-...` system / agent-handshake envelope from 6.4.0+ still flows through the inbound receiver → handler-pipeline → DLQ path even though it has no Message, no MessageType, no Destination, and no Data.

Filing this as a follow-up so the surface gets fully guarded (or, better, the empty envelope gets short-circuited at the receiver entry-point so none of the downstream paths see it).

## Versions

- `WolverineFx*` 6.4.1 (commit `7253408a8`)
- `JasperFx*` 2.8.0, `Marten` 9.4.0, `Polecat` 4.2.1
- net9.0

Same repro repo: CritterWatch PR https://github.com/JasperFx/CritterWatch/pull/277 — `feat/268-276-wolverine-6-4-and-pipeline-tab-cleanup` branch now pinned to 6.4.1; the cold_start integration test still aggregates ~5 NREs at `TrackedSession.AssertNoExceptionsWereThrown`.

## Newly-visible NRE sites (origin/main `7253408a8`)

**1. `src/Wolverine/Runtime/HandlerPipeline.cs:241`**

```csharp
var executor = _executors[envelope.Message!.GetType()];
```

NREs when `envelope.Message` is null — which is exactly the case for the all-zero envelope. The upstream block at lines 207-228 does check `envelope.Message != null` and short-circuits to `NullContinuation.Instance`, but the path that lands here in `executeAsync` skips that early-return.

**2. `src/Persistence/Wolverine.RDBMS/MessageDatabase.Incoming.cs:69`**

```csharp
builder.AppendParameter(envelope.Destination!.ToString());
```

Same null-Destination problem as the original #3013, just on the new path #3014 opened — `MoveToErrorQueue.ExecuteAsync:52` now successfully delegates to `MessageContext.MoveToDeadLetterQueueAsync` → `MessageDatabase.MoveToDeadLetterStorageAsync` which immediately NREs on `envelope.Destination!.ToString()`. The 6.4.1 fix guarded one Destination read; this is the other.

**3. `src/Wolverine/Runtime/Serialization/EnvelopeSerializer.cs:283`**

```csharp
writer.Write(env.Data!.Length);
writer.Write(env.Data);
```

`Data!.Length` triggers `Envelope.Data` getter which calls `assertMessage()` (Envelope.cs:216) which throws `WolverineSerializationException: "Cannot ensure data is present when there is no message. The Message Type Name is ''"`. This path runs during DLQ persistence (`DatabasePersistence.ConfigureDeadLetterCommands:153`) — once the envelope reaches the DLQ writer, it tries to serialize and trips on the empty payload.

## Full inner-trace from the failing cold_start test

(Edited for clarity — five inner exceptions in the aggregate; same three sites repeat as the envelope is retried 3× per occurrence.)

```
fail: Wolverine.Runtime.WolverineRuntime[0]
      Exception detected:
      System.NullReferenceException:
         at Wolverine.RDBMS.MessageDatabase`1.MoveToDeadLetterStorageAsync(Envelope, Exception)
            in Persistence/Wolverine.RDBMS/MessageDatabase.Incoming.cs:line 69
         at Wolverine.Runtime.MessageContext.MoveToDeadLetterQueueAsync(Exception)
            in Wolverine/Runtime/MessageContext.cs:line 401
         at Wolverine.ErrorHandling.MoveToErrorQueue.ExecuteAsync(IEnvelopeLifecycle, IWolverineRuntime, ...)
            in Wolverine/ErrorHandling/MoveToErrorQueue.cs:line 52    ← 6.4.1's guarded path
         at Wolverine.Runtime.HandlerPipeline.InvokeAsync(...)
            in Wolverine/Runtime/HandlerPipeline.cs:line 88

fail: Wolverine.Runtime.WolverineRuntime[0]
      Exception detected:
      System.NullReferenceException:
         at Wolverine.Runtime.HandlerPipeline.executeAsync(MessageContext, Envelope, Activity)
            in Wolverine/Runtime/HandlerPipeline.cs:line 241    ← envelope.Message null
         at Wolverine.Runtime.HandlerPipeline.InvokeAsync(...)
            in Wolverine/Runtime/HandlerPipeline.cs:line 87

fail: Wolverine.Postgresql.PostgresqlMessageStore[0]
      Error trying to serialize a dead letter envelope
      Wolverine.WolverineSerializationException: Cannot ensure data is present when there is no message.
         The Message Type Name is 'unknown/WolverineSerializationException'
         at Wolverine.Envelope.assertMessage() in Envelope.cs:line 216
         at Wolverine.Envelope.get_Data() in Envelope.cs:line 184
         at Wolverine.Runtime.Serialization.EnvelopeSerializer.writeSingle(BinaryWriter, Envelope)
            in EnvelopeSerializer.cs:line 283
         at Wolverine.Runtime.Serialization.EnvelopeSerializer.Serialize(Envelope)
            in EnvelopeSerializer.cs:line 262
         at Wolverine.RDBMS.DatabasePersistence.ConfigureDeadLetterCommands(...)
            in DatabasePersistence.cs:line 153

fail: Wolverine.Runtime.WolverineRuntime[108]
      Envelope Envelope #00000000-0000-0000-0000-000000000000 to shared-memory://critterwatch/critterwatch
      was moved to the error queue
      Wolverine.WolverineSerializationException: Cannot ensure data is present when there is no message.
         The Message Type Name is ''
         at Wolverine.Envelope.assertMessage() in Envelope.cs:line 216
         at Wolverine.Envelope.get_Data() in Envelope.cs:line 184
         at Wolverine.Runtime.HandlerPipeline.TryDeserializeEnvelope(Envelope)
            in HandlerPipeline.cs:line 139
```

## Suggested fix — two options

The point-fixes for the three new sites are mechanical:

**`HandlerPipeline.cs:241`**

```csharp
// The all-zero system envelope can reach here with no Message — bail early instead of NRE'ing
// on the dictionary lookup. The upstream TryDeserializeEnvelope path short-circuits via
// NullContinuation; mirror that here for the executeAsync path.
if (envelope.Message is null)
{
    return MessageSucceededContinuation.Instance;
}
var executor = _executors[envelope.Message.GetType()];
```

**`MessageDatabase.Incoming.cs:69`** — same shape as 6.4.1's MoveToErrorQueue fix:

```csharp
if (envelope.Destination is null)
{
    return;  // can't persist an envelope without a destination to the per-destination DLQ
}
builder.AppendParameter(envelope.Destination.ToString());
```

**`EnvelopeSerializer.cs:283`** — guard the serializer:

```csharp
if (env.Message is null && env.Data is null)
{
    // Empty system envelope reached the serializer — skip persistence entirely.
    return;
}
writer.Write(env.Data!.Length);
writer.Write(env.Data);
```

But the **cleaner** fix is upstream of all three: short-circuit the empty envelope at the receiver entry point so none of the downstream paths see it. Something like the following in `BufferedReceiver.executeAsync` (or wherever the receiver-side envelope first lands):

```csharp
// The all-zero startup / agent-handshake envelope has no semantic payload — discard
// it silently rather than running it through the full handler / DLQ pipeline, where
// every downstream `Envelope.X!` deref has to defend against it independently.
if (envelope.Id == Guid.Empty && envelope.Message is null && envelope.MessageType.IsNullOrEmpty())
{
    return;
}
```

That collapses the five-site defence-in-depth into a single check at the boundary — and avoids the next "fixed one path, three more unmasked" round when some future internal code path also calls `envelope.Data` / `envelope.Destination` / `envelope.Message`.

I'm happy to send the boundary-check fix as a PR if that's the direction you want to take it.

## Why this still matters

The 6.4.1 fix already made the symptoms quieter in CritterWatch's log (the BufferedReceiver retry-spam is gone), but the integration sweep stays red because the aggregated inner exceptions still trigger `TrackedSession.AssertNoExceptionsWereThrown`. Anyone bumping a Wolverine consumer past 6.0.x → 6.4.x is going to hit this until the empty envelope is short-circuited.

Refs: #3013 / #3014 (original NRE fix), CritterWatch#268, CritterWatch#276, CritterWatch#277.

Steve-XYZ:
## Summary
Fixes #3015 by giving Shared Memory receivers a new `Envelope` instance at the transport handoff boundary instead of retaining the sender's pooled envelope.
This follows the intended ownership model: the sender may recycle its pooled envelope after `SendAsync()` completes, while each asynchronous receiver owns an independent envelope wrapper.
## Changes
- add a Shared Memory-local envelope copy helper;
- create a distinct envelope copy per topic subscription;
- copy direct subscription sends before posting to the receiver;
- preserve message, transport metadata, headers, context fields, and custom metric tags;
- intentionally exclude sender, pool, persistence, batching, and receiver runtime lifecycle state;
- add a deterministic regression test proving receiver copies remain valid after the original sender envelope is recycled.
## Validation
- deterministic regression test: 10/10 independent runs passed;
- focused CoreTests Shared Memory coverage: 14 passed;
- scheduled/context Shared Memory regression: passed;
- CoreTests net10.0: 1861 passed, 2 skipped;
- focused request/reply and correlation tests passed when run independently.
The combined Shared Memory compliance selection has pre-existing cross-test interference in `can_send_from_one_node_to_another_by_publishing_rule`; both inline and buffered cases pass in isolation.

### DIFF
--- a/src/Testing/CoreTests/Transports/SharedMemory/shared_memory_envelope_pooling_3015.cs
+++ b/src/Testing/CoreTests/Transports/SharedMemory/shared_memory_envelope_pooling_3015.cs
@@ -1,0 +1,309 @@
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Hosting;
+using Microsoft.Extensions.Logging;
+using Shouldly;
+using Wolverine;
+using Wolverine.Configuration;
+using Wolverine.Runtime;
+using Wolverine.Transports;
+using Wolverine.Transports.Sending;
+using Wolverine.Transports.SharedMemory;
+using Xunit;
+
+namespace CoreTests.Transports.SharedMemory;
+
+public class shared_memory_envelope_pooling_3015
+{
+    private static readonly TimeSpan Timeout = TimeSpan.FromSeconds(10);
+
+    [Fact]
+    public async Task shared_memory_handoff_copies_pooled_envelope_before_sender_recycles_original()
+    {
+        var topicName = $"pooling-3015-{Guid.NewGuid():N}";
+
+        using var host = await Host.CreateDefaultBuilder()
+            .UseWolverine(opts =>
+            {
+                opts.Durability.Mode = DurabilityMode.Serverless;
+                opts.Discovery.DisableConventionalDiscovery();
+                opts.PublishAllMessages().ToSharedMemoryTopic(topicName);
+            })
+            .StartAsync();
+
+        var runtime = host.Services.GetRequiredService<IWolverineRuntime>()
+            .ShouldBeOfType<WolverineRuntime>();
+        var destination = new Uri($"shared-memory://{topicName}");
+        var configuredAgent = runtime.Endpoints.GetOrBuildSendingAgent(destination);
+        var sharedMemorySender = configuredAgent.Endpoint.ShouldBeOfType<SharedMemoryTopic>();
+
+        var receivers = new[] { new BlockingReceiver(), new BlockingReceiver() };
+        var subscriptions = receivers.Select((receiver, index) => new SharedMemorySubscription(
+            sharedMemorySender,
+            $"receiver-{index + 1}-{Guid.NewGuid():N}",
+            EndpointRole.Application)).ToArray();
+        for (var i = 0; i < subscriptions.Length; i++)
+        {
+            await subscriptions[i].BuildListenerAsync(runtime, receivers[i]);
+        }
+
+        var gatedSender = new GateSenderReturn(sharedMemorySender);
+        var diagnosticAgent = new BufferedSendingAgent(
+            runtime.LoggerFactory.CreateLogger<BufferedSendingAgent>(),
+            runtime.MessageTracking,
+            gatedSender,
+            runtime.DurabilitySettings,
+            sharedMemorySender,
+            runtime,
+            null);
+
+        Task? sendTask = null;
+
+        try
+        {
+            var envelope = runtime.AcquireOutgoingEnvelope(diagnosticAgent);
+            var message = new PoolingProbe("original payload");
+            envelope.Message = message;
+            envelope.Destination = destination;
+            envelope.Sender = diagnosticAgent;
+            envelope.SendAttempts = 3;
+            envelope.WasPersistedInOutbox = true;
+            envelope.ResponseType = typeof(string);
+            envelope.Response = "sender-side response";
+            envelope.DoNotCascadeResponse = true;
+            envelope.AlwaysPublishResponse = true;
+            envelope.Batch = [new Envelope(new PoolingProbe("batch member"))];
+            envelope.InBatch = true;
+            envelope.HasBeenAcked = true;
+            envelope.Failure = new InvalidOperationException("sender-side failure state");
+            envelope.SetMetricsTag("org.unit", "platform");
+            var originalId = envelope.Id;
+
+            assertValid(envelope, true, "before Shared Memory enqueue");
+
+            sendTask = envelope.StoreAndForwardAsync().AsTask();
+
+            var senderEnvelope = await waitFor(
+                gatedSender.SharedMemoryEnqueued,
+                "SharedMemoryTopic.SendAsync to enqueue the envelope");
+            var receiverEnvelopes = new[]
+            {
+                await waitFor(receivers[0].Entered, "the first Shared Memory receiver to retain its envelope"),
+                await waitFor(receivers[1].Entered, "the second Shared Memory receiver to retain its envelope")
+            };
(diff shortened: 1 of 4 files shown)


jeremydmiller:
@Steve-XYZ Thanks for taking that on! Frankly, that was probably never something I was going to get around to

Issue #3368, opened 2026-07-11
Wolverine.Grpc: server-side tenant-id detection (mirror of Wolverine.Http's ITenantDetectionPolicies)


erikshafer:
**Is your feature request related to a problem? Please describe.**

`Wolverine.Http` has a first-class tenant-id detection API (`ITenantDetectionPolicies`, exposed as `opts.TenantId` on `WolverineHttpOptions`) that lets an endpoint derive the tenant id from the route, query string, request header, claim, or subdomain, and wires the result into `MessageContext.TenantId` via `DetectTenantIdFrame` so downstream Marten/Polecat session-opening frames pick it up automatically (see `src/Http/Wolverine.Http/Runtime/MultiTenancy/TenantIdDetection.cs` and `docs/guide/http/multi-tenancy.md`).

`Wolverine.Grpc` doesn't have an equivalent yet. The client-side propagation interceptor (`WolverineGrpcClientPropagationInterceptor`, `src/Wolverine.Grpc/Client/WolverineGrpcClientPropagationInterceptor.cs:105`) will stamp an already-known `IMessageContext.TenantId` onto outgoing gRPC calls as a `tenant-id` metadata header, but there's no server-side counterpart that reads that header back off an incoming call and populates `Envelope.TenantId` / `MessageContext.TenantId`. Right now a user has to do that manually in every gRPC service method, which is easy to miss.

This came up as a question in our Discord: someone building on `Wolverine.Grpc` went looking for tenant handling like `Wolverine.Http` has, checked both the docs and the source, and came up empty, because the feature genuinely isn't there yet.

**Describe the solution you'd like**

- An `ITenantDetectionPolicies`-style API on the gRPC options (e.g. `opts.TenantId.IsRequestHeaderValue("tenant-id")`, `.IsClaimTypeNamed(...)` off `ServerCallContext.AuthContext`, `.DetectWith<T>()`), mirroring `src/Http/Wolverine.Http/Runtime/MultiTenancy/`.
- A server-side frame/interceptor (analogous to `DetectTenantIdFrame`) that sets the same `tenantId` codegen variable (`PersistenceConstants.TenantIdVariableName`) that Marten's `OpenMartenSessionFrame` and Polecat's `OpenPolecatSessionFrame` already look for, so tenant-scoped sessions "just work" for gRPC the same way they do for HTTP. No core Wolverine changes needed for this part.
[...]

PR #3369, opened 2026-07-11
gRPC: server-side envelope propagation + correct false docs/sample claims (GH-3368)

erikshafer:
## Summary

- `Wolverine.Grpc`'s client interceptor already stamped `correlation-id`/`tenant-id` onto outgoing calls, but nothing on the server side read them back — despite `docs/guide/grpc/client.md` and the `OrderChainWithGrpc` sample's XML comments claiming a `WolverineGrpcServicePropagationInterceptor` already existed and did exactly this. It didn't; grepping `src/Wolverine.Grpc` confirmed `AddWolverineGrpc()` only ever registered the exception interceptor.
- Adds the real `WolverineGrpcServicePropagationInterceptor` (registered automatically by `AddWolverineGrpc()`, positioned inside the exception interceptor to mirror the client-side ordering). It reads `correlation-id`/`tenant-id` off inbound gRPC metadata and applies them to the ambient `IMessageContext` **unconditionally** when present — matching the convention every other Wolverine transport uses on a freshly received envelope (`MessageContext.ReadEnvelope`), rather than a "set only if empty" guard (a fresh `IMessageContext.CorrelationId` is never actually empty — it's pre-seeded from `Activity.Current` — so that guard would have silently no-op'd, and an earlier draft of this change caught exactly that bug via its own tests).
- Adds `WolverineGrpcOptions.PropagateEnvelopeHeaders` (default `true`) as the server-side off-switch, mirroring the existing client-side option.
- `parent-id`/`conversation-id` are deliberately **not** propagated this way — they live on an `Envelope`, and there is no envelope yet at the point this interceptor runs (`IMessageContext.Envelope` is null outside of message handling). Cross-process parent/trace correlation for gRPC hops already happens separately via OpenTelemetry `Activity` propagation over HTTP/2.
- Corrects the false claims: `docs/guide/grpc/client.md` and the `OrderChainWithGrpc` sample's `ReserveStockHandler.cs` XML comment now describe what's actually implemented instead of a type that never existed.

Related to #3368 (the original Discord question was about tenant-id detection for external gRPC callers; this closes the Wolverine-to-Wolverine hop half of that gap and fixes the docs/sample along the way). External-caller *detection* strategies — analogous to `Wolverine.Http`'s `ITenantDetectionPolicies` (route/claim/subdomain-based) — are still open in #3368.

## Test plan

- [x] `dotnet build wolverine.slnx -c Release` — full solution, 0 errors/warnings
- [x] `dotnet test src/Wolverine.Grpc.Tests/Wolverine.Grpc.Tests.csproj` — 255/255 passing, including 8 new tests:
  - Full Wolverine-to-Wolverine hop: client stamps headers → server interceptor reads them → invoked handler sees them via `IMessageContext` → echoed back (proves the actual bug the docs/sample claimed was already fixed)
  - Tenant-id propagates independent of correlation-id
  - An arbitrary non-Wolverine caller manually setting the raw header also works (the original Discord scenario)
  - `PropagateEnvelopeHeaders = false` actually disables it, verified against a dedicated in-process host
  - DI-registration idempotency guards for the new interceptor (repeat `AddWolverineGrpc()` calls don't double-stack it) + interceptor ordering assertion (exception interceptor stays outermost)
- [x] Verified the `OrderChainWithGrpc` sample separately does *not* actually run today (`Dynamic` codegen mode throws — the sample projects don't reference `Wolverine.RuntimeCompilation`), confirming it was never actually executed/verified before — a pre-existing, unrelated issue left out of scope here. The in-process test suite exercises the identical code path (real Kestrel `TestServer`, real interceptors, real `Bus.InvokeAsync`).

### DIFF
--- a/docs/guide/grpc/client.md
+++ b/docs/guide/grpc/client.md
@@ -106,9 +106,23 @@ builder.Services.AddWolverineGrpcClient<IPingService>(o =>
 });
 ```
 
-On the server side of a Wolverine→Wolverine hop, the envelope headers are read back in the
-`WolverineGrpcServicePropagationInterceptor` already shipped with the adapter, so a call chain
-spanning multiple Wolverine services keeps a single correlation identity without any user wiring.
+On the server side of a Wolverine→Wolverine hop, `WolverineGrpcServicePropagationInterceptor`
+(registered automatically by `AddWolverineGrpc()`) reads the `correlation-id` and `tenant-id`
+headers back off the inbound call and applies them to the ambient `IMessageContext` before the
+service method runs. Only those two identifiers round-trip this way — `parent-id` and
+`conversation-id` live on an `Envelope`, and there is no envelope yet at this point in the
+pipeline (`IMessageContext.Envelope` is null outside of message handling). Once the service
+method calls `Bus.InvokeAsync(...)`, Wolverine copies the now-set `CorrelationId`/`TenantId` onto
+the freshly created envelope automatically, and Marten/Polecat tenant-scoped session frames pick
+it up the same way they would for any other transport — no further wiring needed.
+
+Cross-process parent/trace correlation for gRPC hops is handled separately by OpenTelemetry
+`Activity` propagation over HTTP/2 (W3C traceparent) — see [How gRPC Handlers Work](./handlers).
+The `parent-id`/`conversation-id` headers still go out on the wire for diagnostic/non-Wolverine
+consumers, but nothing on the Wolverine side reads them back in.
+
+Server-side propagation can be disabled with `WolverineGrpcOptions.PropagateEnvelopeHeaders =
+false`, the counterpart to the client-side `PropagateEnvelopeHeaders` option above.
 
 ## `RpcException` → typed-exception translation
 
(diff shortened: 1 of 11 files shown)


jeremydmiller (approved):
Reviewed — this is right, and the write-up is exemplary. Points that got specific scrutiny and passed:

- **Unconditional set on header presence** is the correct convention, and your reasoning is exactly why: a fresh `IMessageContext.CorrelationId` is pre-seeded (Activity root or GUID), so an "only if empty" guard would silently never fire. Matching `MessageContext.ReadEnvelope` semantics is the consistent choice.
- **Not propagating `parent-id`/`conversation-id`** is correct — no envelope exists at interceptor time, and trace lineage already rides W3C traceparent via Activity propagation.
- Interceptor ordering (exception translation outermost, symmetric to the client side) and the scoped-`IMessageContext` null-bail are both right.
- Correcting the docs/sample claims about a type that never existed: thank you for grepping instead of assuming.

Merging on green (24/24). Leaving #3368 open: this lands the propagation leg, and the remaining feature is the full `ITenantDetectionPolicies` mirror (claims-based + custom detectors + the codegen tenant-id variable so Marten/Polecat session-opening frames get the tenant without touching the ambient context). That work will build on this interceptor rather than replace it.

Issue #2903, opened 2026-05-26
Wolverine centric OpenAPI generation command that isn't as problematic as Microsoft's


jeremydmiller:
Here's the rub, the Microsoft version of creating OpenAPI documentation from a command line tool tries to call `IHost.StartAsync()`, which causes all kinds of hilarity with Wolveirne that tries to start itself in an IHostedService -- plus lots of other tools do too. Let's add a custom `openapi` command that does the same thing functionally, but doesn't cause the same problems w/ trying to connect to databases and what not. 

See this code sample as a starting point, but we'll make sure it works with just `IHost.Build()` or model this after the `codegen` command that can selectively turn off some of Wolverine bootstrapping. 

We're going to do this by reverse engineering Microsoft.Extensions.ApiDescription.Server.

Build the new command inside of Wolverine.Http as a JasperFx command.

PR #3373, opened 2026-07-11
Read HttpGraph instead of EndpointDataSource in WolverineApiDescriptionProvider. Closes GH-3371

### LINKED ISSUE #3371: Read to ApiExplorer before server startup permanently empties every OpenAPI document
**Describe the bug**

On a Wolverine.HTTP host using minimal hosting and `Microsoft.AspNetCore.OpenApi`, any ApiExplorer read that lands before server start permanently empties every OpenAPI document. After the hit `/openapi/v1.json` returns 200 with an empty document, e.g `paths: {}` for the lifetime of the host, while HTTP routing works normally. 

This was first noticed in our CI (2-core GitHub runners) this hits roughly 40% of runs; locally it is rare. Once it happens, the document never recovers without a restart.

With CritterWatch monitoring attached, Wolverine.Http ships such an early reader in the box (the capabilities snapshot, since #2720), so a plain Wolverine + CritterWatch + `AddOpenApi()` host can break its own OpenAPI documents with no user code involved.

**I will follow up with a PR containing both the repro (as a failing-then-passing regression test) and a fix** — the full `Wolverine.Http.Tests` (818) and `Wolverine.Http.AspVersioning.Tests` (66) suites already pass against it, including `generate_openapi_without_database`.

**To Reproduce**

Complete single-file repro, verified against WolverineFx.Http 6.17.1 (no CritterWatch needed — any early reader shows it). Packages: `WolverineFx.Http` 6.17.1, `WolverineFx.RuntimeCompilation` 6.17.1, `Microsoft.AspNetCore.OpenApi` 10.0.9.

```csharp
using Microsoft.AspNetCore.Mvc.ApiExplorer;
using Wolverine;
using Wolverine.Http;

var builder = WebApplication.CreateBuilder(args);

builder.Host.UseWolverine();

builder.Services.AddEndpointsApiExplorer();
builder.Services.AddOpenApi();
builder.Services.AddWolverineHttp();

// The early reader. Any hosted service registered before Build() starts ahead of the web
// server (GenericWebHostService is appended last), so its StartAsync runs pre-start.
builder.Services.AddHostedService<EarlyApiExplorerToucher>();

var app = builder.Build();

app.MapOpenApi();
app.MapWolverineEndpoints();

app.Run();

public class EarlyApiExplorerToucher(IApiDescriptionGroupCollectionProvider provider) : IHostedService
{
    public Task StartAsync(CancellationToken cancellationToken)
    {
        // First read caches for the lifetime of the host
        _ = provider.ApiDescriptionGroups;
        return Task.CompletedTask;
    }

    public Task StopAsync(CancellationToken cancellationToken) => Task.CompletedTask;
}

public static class HelloEndpoint
{
    [WolverineGet("/hello")]
    public static string Get() => "hello";
}
```

1. Run the app: `GET /hello` returns `hello` (routing works), but `GET /openapi/v1.json` returns `"paths": {}` — 100% of the time, for the lifetime of the host.
2. Remove the `AddHostedService<EarlyApiExplorerToucher>()` line: `/openapi/v1.json` contains `/hello`.
3. Moving the same read into an `IHostApplicationLifetime.ApplicationStarted` callback also returns the full description set — confirming the window is "after Wolverine endpoint mapping, before server pipeline build".

**Expected behavior**

OpenAPI documents contain all Wolverine endpoints regardless of when the first ApiExplorer read happens. Wolverine already treats pre-start OpenAPI generation as a supported scenario — the `openapi` command (GH-2903 / #2911, which you added following our request — thank you!) never calls `IHost.StartAsync()`, and `generate_openapi_without_database` asserts real routes appear in that regime. An in-process pre-start read should meet the same contract.

**Desktop (please complete the following information):**

- OS: linux-x64 (GitHub-hosted 2-core runners); reproduced deterministically on Windows 11
- Browser: n/a (server-side; observed via curl and the Scalar UI)
- Version: WolverineFx / WolverineFx.Http 6.16.0 (provider unchanged on current main, post-6.17.1); Wolverine.CritterWatch 1.0.0-beta.1; Microsoft.AspNetCore.OpenApi 10.0.9; .NET 10, minimal hosting

**Additional context**

Three behaviours interact:

1. ASP.NET Core's `ApiDescriptionGroupCollectionProvider` caches its result keyed on `ActionDescriptorCollection.Version`. Endpoint-routing endpoints are not action descriptors, so in an endpoint-routing-only host the version never changes — whatever the first read produced is served for the lifetime of the host.
2. Under minimal hosting (`WebApplication`), the host wires endpoints mapped with `app.MapWolverineEndpoints()` into the DI-level `EndpointDataSource` only when it builds the server's request pipeline — inside `GenericWebHostService.StartAsync`, which runs last among hosted services. (#2911 works around exactly this for the CLI path — `OpenApiCommand.ComposeEndpointDataSources` merges the data sources "exactly the way `UseEndpoints` does" because the real merge only runs once the host starts. That protection covers the CLI's one-shot process only.)
3. `WolverineApiDescriptionProvider` enumerates the injected DI `EndpointDataSource` (`src/Http/Wolverine.Http/WolverineApiDescriptionProvider.cs`, `OnProvidersExecuting`) rather than Wolverine's own `HttpGraph`, which is already complete when `MapWolverineEndpoints()` returns.

Consequence: any pre-start ApiExplorer read sees zero Wolverine endpoints, and (1) freezes that empty result forever. Every OpenAPI document generated later — `Microsoft.AspNetCore.OpenApi`, Swashbuckle, versioned documents via `WolverineFx.Http.AspVersioning` — stays permanently empty.

The in-box racing reader, when CritterWatch monitoring is attached (we run `Wolverine.CritterWatch` 1.0.0-beta.1): `ServiceCapabilities.ReadFrom` → `readHttpGraphs` → `HttpGraphUsageSource.TryCreateUsage` → `OpenApiDescriptorBuilder.TryBuildForWolverine` → `apiDescriptions.ApiDescriptionGroups.Items` — fires from the observer's boot-time batching loop after the Wolverine runtime starts, and can land before the web host wires endpoints. On slow or 2-core machines it wins the race often, and exposure grows with CritterWatch adoption. (Related: GH-3008/3009 touch the same snapshot path.)

The fix we will PR: `WolverineApiDescriptionProvider` doesn't need the DI `EndpointDataSource` — `WolverineHttpOptions.Endpoints` (the `HttpGraph`) holds every `HttpChain`, each with its `RouteEndpoint` already built by `MapWolverineEndpoints()`, and every value the provider puts into an `ApiDescription` already comes from `chain.Endpoint`, not from the enumerated data source. Enumerating the graph makes the provider's answer identical before and after server start, and ASP.NET's version-keyed cache becomes harmless — for Wolverine-sourced descriptions; minimal-API endpoints in a hybrid host still compose at server start, which only ASP.NET can fix. (`HttpGraphUsageSource` already takes exactly this approach for the chain walk itself.)

Interim mitigation for anyone hitting this before the fix ships: replace `IApiDescriptionGroupCollectionProvider` with a wrapper that serves pre-start reads from a throwaway `ApiDescriptionGroupCollectionProvider` instance (a fresh instance has a fresh cache, so the lifetime cache cannot be poisoned) and delegates to the default provider after `ApplicationStarted`. We run this in production today and can share it.

uniquelau:
Fixes GH-3371: ASP.NET Core caches the first ApiExplorer read for the lifetime of the host (keyed on `ActionDescriptorCollection.Version`, which endpoint routing never bumps), and Wolverine endpoints only reach the DI `EndpointDataSource` at server start — so any earlier read permanently empties every OpenAPI document.

## The change

`WolverineApiDescriptionProvider` enumerates `WolverineHttpOptions.Endpoints` (the `HttpGraph`) instead of the injected `EndpointDataSource`. The graph — including each chain's built `RouteEndpoint` — is complete as soon as `MapWolverineEndpoints()` returns, so the provider's answer is the same before and after server start.

This is behavior-preserving: every value in the produced `ApiDescription` already comes from `chain.Endpoint`; the data source enumeration only ever recovered the chain from endpoint metadata. The `ExcludeFromDescription` filter and the QUERY-verb skip are unchanged. Chains are enumerated rather than `graph.Endpoints` because the chain is what descriptions are built from — happy to switch if you'd rather keep the endpoint-shaped loop.

`OpenApiCommand.ComposeEndpointDataSources` is intentionally untouched: ASP.NET's own minimal-API provider still reads the composite source, so the CLI's merge is still needed for non-Wolverine endpoints.

## Behavior notes

- A read before `MapWolverineEndpoints()` returns still sees nothing — unchanged.
- Only Wolverine's descriptions become start-independent. Minimal-API descriptions still compose at server start, so a pre-start read on a hybrid host now freezes a document with Wolverine endpoints but no minimal-API ones, where previously it froze fully empty. Hybrid apps with boot-time ApiExplorer readers should still defer them.
- Calling `MapWolverineEndpoints()` twice now describes only the last graph. Double-mapping duplicates every route and was never a supported configuration.
- A `chain.Endpoint == null` guard skips chains added to the graph after mapping — they never reached the data source either.

## Testing

- New regression test `api_explorer_before_host_start`: maps endpoints, never starts the host, asserts descriptions are complete on the first ApiExplorer read. Fails on the old provider, passes on the new one.
- `Wolverine.Http.Tests` 818/818 and `Wolverine.Http.AspVersioning.Tests` 66/66, including `generate_openapi_without_database` (GH-2903). The `openapi` command emits the same document with no database running.
- Verified in a consuming application (Wolverine.HTTP + Marten + CritterWatch + per-surface OpenAPI documents): full suite green against packages built from this branch, generated OpenAPI spec byte-identical, and the empty-document failure no longer reproduces even with the application's interim workaround removed.

### DIFF
--- a/docs/guide/http/metadata.md
+++ b/docs/guide/http/metadata.md
@@ -299,6 +299,12 @@ asynchronous Wolverine extension) will not appear. This matches the goal of buil
 is worth knowing if you add endpoints dynamically.
 :::
 
+The same holds in-process: Wolverine's endpoint descriptions are complete on the very first
+ApiExplorer read after `MapWolverineEndpoints()` has run, even before the host starts. Only
+Wolverine's descriptions are start-independent — ASP.NET's own minimal API descriptions still
+compose at server start — so applications mixing the two should defer early ApiExplorer reads
+until after startup.
+
 ## With Microsoft.Extensions.ApiDescription.Server
 
 ::: tip
--- a/src/Http/Wolverine.Http.Tests/api_explorer_before_host_start.cs
+++ b/src/Http/Wolverine.Http.Tests/api_explorer_before_host_start.cs
@@ -1,0 +1,45 @@
+using Microsoft.AspNetCore.Builder;
+using Microsoft.AspNetCore.Mvc.ApiExplorer;
+using Microsoft.Extensions.DependencyInjection;
+using Shouldly;
+using Wolverine.Http.Tests.DifferentAssembly.Validation;
+
+namespace Wolverine.Http.Tests;
+
+// ASP.NET Core caches the first ApiExplorer read for the lifetime of the host, and a host can
+// be read before it starts (build-time OpenAPI generation, monitoring snapshots). Wolverine's
+// descriptions must be complete on that first read, however early it happens.
+public class api_explorer_before_host_start
+{
+    [Fact]
+    public async Task api_descriptions_are_complete_before_the_host_starts()
+    {
+        var builder = WebApplication.CreateBuilder(new WebApplicationOptions { EnvironmentName = "Development" });
+
+        builder.Host.UseWolverine(opts =>
+        {
+            // Pin endpoint discovery to the small, isolated "DifferentAssembly" so the test is
+            // deterministic and does not pick up the rest of the test suite's endpoints.
+            opts.ApplicationAssembly = typeof(Validated2Endpoint).Assembly;
+        });
+
+        builder.Services.AddEndpointsApiExplorer();
+        builder.Services.AddWolverineHttp();
+
+        await using var app = builder.Build();
+        app.MapWolverineEndpoints();
+
+        // Note: no app.StartAsync(). This is the first ApiExplorer read, which ASP.NET Core
+        // caches for the lifetime of the host
+        var provider = app.Services.GetRequiredService<IApiDescriptionGroupCollectionProvider>();
+
+        var relativePaths = provider.ApiDescriptionGroups.Items
+            .SelectMany(x => x.Items)
+            .Where(x => x.ActionDescriptor is WolverineActionDescriptor)
+            .Select(x => x.RelativePath)
+            .ToList();
+
+        relativePaths.ShouldContain("validate2/customer");
+        relativePaths.ShouldContain("validate2/user");
+    }
+}
--- a/src/Http/Wolverine.Http/WolverineApiDescriptionProvider.cs
+++ b/src/Http/Wolverine.Http/WolverineApiDescriptionProvider.cs
@@ -1,50 +1,57 @@
 using JasperFx.Core.Reflection;
 using Microsoft.AspNetCore.Mvc.ApiExplorer;
 using Microsoft.AspNetCore.Routing;
-using Microsoft.Extensions.Hosting;
 
 namespace Wolverine.Http;
 
 internal class WolverineApiDescriptionProvider : IApiDescriptionProvider
 {
-    private readonly EndpointDataSource _endpointDataSource;
-    private readonly IHostEnvironment _environment;
+    private readonly WolverineHttpOptions _options;
 
-    public WolverineApiDescriptionProvider(
-        EndpointDataSource endpointDataSource,
-        IHostEnvironment environment)
+    public WolverineApiDescriptionProvider(WolverineHttpOptions options)
     {
-        _endpointDataSource = endpointDataSource;
-        _environment = environment;
+        _options = options;
     }
 
     public void OnProvidersExecuting(ApiDescriptionProviderContext context)
     {
-        foreach (var endpoint in _endpointDataSource.Endpoints)
+        // Read the HttpGraph: it is complete as soon as MapWolverineEndpoints() has run,
+        // while the composite EndpointDataSource only fills at server start — and ASP.NET
+        // Core caches the first ApiExplorer read for the host lifetime, however early it is.
+        if (_options.Endpoints is not { } graph)
         {
-            if (endpoint is RouteEndpoint routeEndpoint && routeEndpoint.Metadata.GetMetadata<HttpChain>() is {} chain)
+            return;
+        }
+
+        foreach (var chain in graph.Chains)
+        {
+            // A chain added to the graph after MapWolverineEndpoints() has run never gets a
+            // built RouteEndpoint, and CreateApiDescription() requires one
+            if (chain.Endpoint == null)
             {
-                if (chain.Method.HandlerType.HasAttribute<ExcludeFromDescriptionAttribute>() ||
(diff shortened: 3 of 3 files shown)


jeremydmiller (approved):
Reviewed in depth — this is exactly right, and the key insight makes it safely behavior-preserving: even in the old code, the composite `EndpointDataSource` enumeration was only ever used to *rediscover the `HttpChain`* from endpoint metadata; the `ApiDescription` itself was always built by `chain.CreateApiDescription()` off `chain.Endpoint`. Since the data-source `RouteEndpoint` for a Wolverine endpoint is the *same instance* as `chain.Endpoint` (both from `chain.BuildEndpoint()` in `HttpGraph`), switching discovery to `WolverineHttpOptions.Endpoints` changes nothing post-start — it just makes the provider complete from the moment `MapWolverineEndpoints()` returns. Ordering, the `ExcludeFromDescription` filter, and the QUERY-verb skip all carry over verbatim, and AspVersioning's document discovery was already graph-based, so this makes the provider consistent with it.

Also verified as correct calls:
- Leaving the CLI `ComposeEndpointDataSources` path untouched — still needed for minimal-API endpoints in hybrid CLI generation.
- Documenting (not fixing) the hybrid-host limitation; that half is ASP.NET's version-keyed cache.
- The new `chain.Endpoint == null` guard, which matches prior behavior (such chains never reached the composite source either).
- The disclosed double-`MapWolverineEndpoints()` semantic (only the last graph is described) — double-mapping was never supported, acceptable.

One non-blocking suggestion: the regression test asserts a complete **pre-start** read; adding a start-then-re-read assertion in the same test would lock in the full GH-3371 loop (the cached collection staying complete across `StartAsync`). Happy to take that as a follow-up if you'd rather not touch the PR again.

Merging once CI is green. Thanks for an unusually well-researched fix — the root-cause analysis on the version-keyed cache saved us real time, and this also unblocks a monitoring scenario on our side (CritterWatch's capability snapshot is an in-box pre-start ApiExplorer reader).

Issue #3385, opened 2026-07-12
gRPC service path doesn't propagate saga-id — header-identified sagas can't be started/continued over gRPC


erikshafer:
## Context

Following #3369 / #3382 (server-side envelope propagation + tenant detection for `Wolverine.Grpc`, GH-3368), a question came up about whether **starting a saga** is similarly unsupported over the gRPC package. Tracing it, there's a real gap, but it's narrower than it first looks.

## What already works (no change needed)

A saga is just a `HandlerChain` (`SagaChain`), and the gRPC shim forwards each RPC to `IMessageBus.InvokeAsync<TResponse>` / `StreamAsync<T>`. So sagas whose identity comes from the **message body** — `[SagaIdentity]`, `OrderId`, or a conventional `Id` on the request DTO — start and continue over gRPC today with zero gRPC-specific code. `PullSagaIdFromMessageFrame` reads the id straight out of the deserialized request.

## The gap

Sagas with **no id member** on the message resolve identity from the envelope `saga-id` header via `PullSagaIdFromEnvelopeFrame` (`src/Wolverine/Persistence/Sagas/PullSagaIdFromEnvelopeFrame.cs:25,34`), which throws `IndeterminateSagaStateIdException` when `envelope.SagaId` is empty. Across a gRPC service hop that value is never populated, for **two** independent reasons:

1. **Interceptors don't carry `saga-id`.** Client stamps `correlation-id`/`tenant-id`/`id`/`parent-id`/`conversation-id` (`WolverineGrpcClientPropagationInterceptor.cs:104-113`); server reads back only `correlation-id`/`tenant-id` (`WolverineGrpcServicePropagationInterceptor.cs:109-119`).
2. **The invoke path drops it.** `Executor.InvokeAsync<T>` (`src/Wolverine/Runtime/Handlers/Executor.cs:166-181`) copies `bus.TenantId` onto the invoked envelope but has no equivalent for saga id. So even a context-level saga id wouldn't reach `PullSagaIdFromEnvelopeFrame`. (This is why the tenant fix was cheap and this one isn't — the tenant copy already existed.)

## Fix surface

| # | Location | Change |
|---|----------|--------|
| 1 | Client interceptor | Stamp `EnvelopeConstants.SagaIdKey` from `ctx.Envelope?.SagaId`. |
[...]

PR #3386, opened 2026-07-12
gRPC saga coverage: message-identified sagas work like HTTP; header-identified gap pinned (refs #3385)

erikshafer:
Refs #3385.

Adds gRPC saga tests that mirror the existing HTTP endpoint + saga coverage, proving sagas work over gRPC the same way they do over HTTP — and documenting the one case that doesn't (which HTTP shares). See the findings summary on #3385.

### What's here

**Happy path — message-identified sagas work over gRPC (green).**
`can_start_and_continue_a_message_identified_saga_over_grpc` is the gRPC parallel of `Wolverine.Http.Tests.building_a_saga_and_publishing_other_messages_from_http_endpoint`: a saga whose id rides on the request DTO is started, persisted, continued, completed, and its state deleted — all through the same `Bus.InvokeAsync` pipeline a `[WolverinePost]` endpoint uses. Persistence is asserted via the in-memory saga persistor, so **no database is required**.

**Characterization — header-identified sagas hit the same limitation HTTP has.**
`starting_a_header_identified_saga_over_grpc_fails_with_opaque_status_today` pins the current behavior: a saga whose id is resolved from the envelope `saga-id` header can't resolve it over a gRPC service hop (nothing carries `saga-id` across), surfacing as an opaque `StatusCode.Internal`. When we land the scoped diagnostic (or full propagation) from #3385, this test's assertions flip.

### Notes
- Test-only — no production change. It's the no-regret groundwork that holds whichever direction #3385 takes.
- Uses a standalone in-process host modeled on the existing `ServerPropagation` fixtures.

### DIFF
--- a/src/Wolverine.Grpc.Tests/SagaOverGrpc/CountingSaga.cs
+++ b/src/Wolverine.Grpc.Tests/SagaOverGrpc/CountingSaga.cs
@@ -1,0 +1,22 @@
+namespace Wolverine.Grpc.Tests.SagaOverGrpc;
+
+/// <summary>
+///     A <em>header-identified</em> saga: <see cref="StartCountingRequest"/> carries no id member,
+///     so Wolverine resolves the saga id from the inbound envelope's <c>saga-id</c> header. Over a
+///     gRPC service hop that header is never populated — neither the client nor server propagation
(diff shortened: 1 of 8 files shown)


jeremydmiller (approved):
Great no-regret groundwork — the in-memory persistor assertions keep this database-free, and the characterization test is exactly what we want pinned ahead of the GH-3385 decision. Merging.

Issue #3372, opened 2026-07-11
Wolverine.HTTP: [AsParameters] query binder silently ignores unparseable values instead of returning 400


outofrange-consulting:
## What happens

When an `[AsParameters]` query model has a **non-nullable enum** (or any value-typed) property with a default initializer, Wolverine's generated query-string binder silently ignores an unparseable value and leaves the property at its initializer. The request then succeeds with the *default* semantics instead of being rejected.

ASP.NET Core's own minimal-API `[AsParameters]` binder returns `400 Bad Request` for the same input, so the behaviour differs from the framework the attribute is borrowed from.

The practical consequence is that a typo'd query value is indistinguishable from an absent one, so there is no way to reject it — not even downstream. A FluentValidation `IsInEnum()` rule can't catch it either: the binder never writes an out-of-range value (such as `0`) that the validator could see, it just doesn't assign at all.

## Repro

```csharp
using Wolverine;
using Wolverine.Http;

var builder = WebApplication.CreateBuilder(args);
builder.Host.UseWolverine();
var app = builder.Build();
app.MapWolverineEndpoints();
await app.RunAsync();

public enum SortField { Name = 1, Date = 2 }

public sealed record SearchQuery
{
    [FromQuery(Name = "sort-by")]
    public SortField SortBy { get; set; } = SortField.Date;
}

public static class SearchEndpoint
{
    [WolverineGet("/search")]
    public static string Get([AsParameters] SearchQuery query) => query.SortBy.ToString();
}
```

| Request | Wolverine | ASP.NET Core minimal API |
| --- | --- | --- |
| `GET /search?sort-by=Name` | `200` — `Name` | `200` — `Name` |
| `GET /search?sort-by=bogus` | `200` — `Date` | `400` |

The same holds for `int` / `long` / `Guid` properties: `?page-size=abc` binds to the initializer rather than failing.

## Expected

An unparseable value for a bound query parameter should be reportable. Any of these would resolve it:

1. return `400` on a failed parse, matching ASP.NET Core's `[AsParameters]` binder;
2. make it opt-in (e.g. a `WolverineHttpOptions` flag) so existing lenient behaviour is preserved by default;
3. at minimum, record the failure somewhere an endpoint or middleware can observe (e.g. as a binding-problem collection on the `HttpContext`), so a policy can turn it into a `400`.

## Environment

- WolverineFx.Http 6.17.0
- .NET 10

jeremydmiller:
[...]

PR #3402, opened 2026-07-13
[AsParameters]: reject unparseable values in collection query parameters (GH-3398)

### LINKED ISSUE #3398: [AsParameters]: unparseable value in a collection query parameter silently binds null instead of returning 400 (scalar case fixed by #3372)
### Summary

#3372 (fixed in 6.17.1 / 6.17.2) made an **unparseable scalar** query string value return `400 Bad Request` instead of silently binding the property's default. The same failure mode is still present for **collection / array** query string parameters bound through `[AsParameters]`.

When one member of a multi-value query parameter cannot be parsed, the whole collection binds to `null` and the request succeeds with `200 OK`. The endpoint cannot distinguish that from "the caller did not send the parameter at all".

### Version

WolverineFx / WolverineFx.Http **6.17.2**, .NET 10.

### Repro

```csharp
public enum Colour { Red, Green, Blue }

public record SearchRequest
{
    public Colour[]? Colours { get; set; }
    public Colour? Colour { get; set; }
}

public static class SearchEndpoint
{
    [WolverineGet("/widgets")]
    public static IResult Get([AsParameters] SearchRequest request)
        => Results.Ok(new { colours = request.Colours, colour = request.Colour });
}
```

| Request | Actual (6.17.2) | Expected |
| --- | --- | --- |
| `GET /widgets?colour=Purple` | `400 Bad Request` ✅ (fixed by #3372) | `400` |
| `GET /widgets?colours=Purple` | **`200 OK`, `Colours == null`** ❌ | `400` |
| `GET /widgets?colours=Red&colours=Purple` | **`200 OK`, `Colours == null`** ❌ | `400` |

The same holds for `int[]` / `Guid[]` / `long[]` collections, not just enums — any element whose `TryParse` fails takes the entire collection to `null`.

### Why this is worse than a plain 400

On a filtering/search endpoint, the bound collection is typically used as an optional predicate:

```csharp
if (request.Colours is { Length: > 0 })
{
    query = query.Where(x => request.Colours.Contains(x.Colour));
}
```

A typo in one value therefore does not fail the request — it **silently removes the filter** and returns an *unfiltered* result set with a `200`, while the caller believes the results are filtered. That is a silent-wrong-answer failure mode rather than a loud one.

### Validation cannot compensate

FluentValidation (via `WolverineFx.Http.FluentValidation`) sees `null`, not an out-of-range value, so neither of these fires:

```csharp
RuleForEach(x => x.Colours).IsInEnum();          // no elements to iterate
RuleFor(x => x.Colours).NotEmpty().When(x => x.Colours is not null);  // it *is* null
```

There is no way, from inside the handler or a validator, to tell "unparseable value supplied" apart from "parameter omitted". The information is lost in the binder.

### Expected behaviour

A collection query string parameter whose element fails to parse should fail the request with `400 Bad Request`, consistent with the scalar behaviour introduced by the fix for #3372.

jeremydmiller:
Fixes #3398. Completes the job started in #3372 (shipped in 6.17.1/6.17.2), which explicitly left collection query binding as a follow-up.

## Why this one is worse than a plain 400

An unparseable *element* of a multi-valued query string parameter was silently dropped by the generated binder. On a filtering endpoint the bound collection is typically an optional predicate:

```csharp
if (request.Colours is { Length: > 0 })
{
    query = query.Where(x => request.Colours.Contains(x.Colour));
}
```

So `?colours=Purple` (typo) did not fail the request — it **silently removed the filter** and returned an *unfiltered* `200` while the caller believed the results were filtered. That is a silently-wrong-answer failure mode, not a loud one, and it is unrecoverable downstream: `null` is indistinguishable from "parameter omitted", so FluentValidation's `RuleForEach` has nothing to iterate and `NotEmpty().When(x => x.Colours is not null)` never fires. Applies to `int[]` / `Guid[]` / `long[]` as well as enums.

## The fix

`ParsedArrayQueryStringValue` (`T[]`) and `ParsedCollectionQueryStringValue` (`List<T>`, `IList<T>`, `IReadOnlyList<T>`, `IEnumerable<T>`) now accept the same `rejectUnparseableValue` flag `ReadHttpFrame` got in #3372, wired from `WolverineHttpOptions.RejectUnparseableQueryValues` through `HttpChain`. Both emit a shared `StrictQueryBinding.WriteRejectionBlock`: an element that is *present* but fails `TryParse` writes a 400 ProblemDetails naming the parameter (via the existing `HttpHandler.WriteQueryValueParsingProblem`) and exits the generated method.

Semantics match the scalar case exactly:

| Request | Flag off (default, unchanged) | Flag on |
| --- | --- | --- |
| `?Colours=Red&Colours=Blue` | `200` — `[Red, Blue]` | `200` — `[Red, Blue]` |
| *(omitted)* | `200` — keeps initializer | `200` — keeps initializer |
| `?Colours=Purple` | `200` — `null` | `400` ProblemDetails naming `Colours` |
| `?Colours=Red&Colours=Purple` | `200` — `[Red]` | `400` ProblemDetails naming `Colours` |

The default is **not** changed; the flag stays opt-in through 6.x and flips to strict in 7.0. String collections are untouched (nothing to parse). Form/route/header binding untouched.

## Decision: all-or-nothing binding

`?colours=Red&colours=Purple` rejects the **whole** binding rather than binding `[Red]` and dropping `Purple`. ASP.NET Core minimal APIs reject the request when any element of a bound collection parameter fails to parse (the binding failure is recorded and the pipeline returns 400) rather than yielding a partial collection, so this is also the minimal-API-compatible answer that #3372 set out to match. Independently it is the only defensible option: a partially-bound collection is exactly the silently-wrong-answer this issue is about — the caller asked to filter on two colours and would get results filtered on one, with a `200` and no signal. Silently nulling the whole collection is worse still. If the caller genuinely wants lenient element dropping, that is what the (default) flag-off behavior gives them.

## Red → green evidence

Tests added to `src/Http/Wolverine.Http.Tests/strict_query_string_binding.cs` (strict + lenient hosts already in that fixture): enum array, `int[]`, `Guid[]`, `List<int>`, a direct `int[]` method argument, the mixed valid+invalid case, valid-values-bind, omitted-parameter-keeps-initializer, and two flag-off lenience tests.

Before the fix — exactly the six strict-mode assertions fail, and the lenient/omitted expectations already pass (proving the behavior change is confined to the flag):

```
Failed  one_bad_element_rejects_the_whole_collection
Failed  malformed_enum_array_element_returns_400
Failed  malformed_int_array_element_returns_400
Failed  malformed_guid_array_element_returns_400
Failed  malformed_generic_list_element_returns_400
Failed  malformed_collection_element_on_direct_method_argument_returns_400
Failed!  - Failed: 6, Passed: 17, Total: 23
```

After:

```
Passed!  - Failed: 0, Passed: 23, Total: 23
```

Full `Wolverine.Http.Tests` suite on net9.0: **844 passed, 0 failed, 10 skipped**. `dotnet build wolverine.slnx -c Release`: 0 warnings, 0 errors.

Docs: `docs/guide/http/as-parameters.md` gains a "Collections" section under Strict Query String Binding with the behavior matrix and the all-or-nothing rationale; `docs/guide/http/querystring.md` notes collection coverage.

Thanks to @outofrange-consulting for the report (and for the scalar case in #3372).

### DIFF
--- a/docs/guide/grpc/sagas.md
+++ b/docs/guide/grpc/sagas.md
@@ -176,13 +176,24 @@ Wolverine has two ways to resolve which saga state a message belongs to:
 2. **Header-identified** — the message carries no identity member, and Wolverine falls back to the
    envelope's `saga-id` header.
 
-**Only message-identified sagas are supported over a gRPC service hop today.** The `saga-id`
-envelope header does not cross the hop — neither the client nor server propagation interceptor
-carries it, so a header-identified saga invoked through a gRPC service fails with an
-`IndeterminateSagaStateIdException` (surfaced to the caller as a `StatusCode.Internal`
-`RpcException`). This is tracked in
-[GH-3385](https://github.com/JasperFx/wolverine/issues/3385); the practical guidance is simple and
-is the better design anyway:
+**Only message-identified sagas are supported over a gRPC service hop.** The `saga-id` envelope
+header does not cross the hop — neither the client nor server propagation interceptor carries it —
+so a header-identified saga invoked through a gRPC service cannot resolve an id at all.
+
+As of 6.18.0 that fails with an explicit diagnostic rather than an opaque one: the caller gets a
+`StatusCode.InvalidArgument` `RpcException` whose detail names both the cause and the fix.
+
+> Could not determine a saga id for this request. A saga started or continued over a gRPC hop must
+> carry its identity ON THE MESSAGE BODY: the 'saga-id' envelope header is not propagated across a
+> gRPC call, so a header-identified saga cannot work over gRPC. Put the saga identity on the request
+> message itself (a property Wolverine can match to the saga id, or one marked with
+> `[SagaIdentity]`).
+
+`InvalidArgument` rather than `Internal` is deliberate ([AIP-193](https://google.aip.dev/193)): the
+request cannot succeed as sent, and no amount of retrying will change that — it is a contract
+problem, not a transient server fault.
+
+The practical guidance is simple, and is the better design anyway:
 
 ::: tip
 Put the saga identity on the request DTO. It makes the contract self-describing for non-Wolverine
--- a/docs/guide/http/as-parameters.md
+++ b/docs/guide/http/as-parameters.md
@@ -253,6 +253,28 @@ member's default / initializer in both modes. The flag applies to query string b
 `[AsParameters]` members and for endpoint method arguments bound from the query string alike.
 Route argument binding is unaffected (an unparseable route value already returns `404`).
 
+### Collections <Badge type="tip" text="6.18" />
+
+The flag covers **collection** query string parameters (`T[]`, `List<T>`, `IList<T>`,
+`IReadOnlyList<T>`, `IEnumerable<T>`) too. With the flag off, an element that fails to parse is
+silently dropped, which is especially dangerous for an optional filter: the endpoint sees `null`
+(or a partial collection) and quietly returns an *unfiltered* `200` while the caller believes the
+results were filtered.
+
+With the flag on, binding a collection is **all or nothing** — a single unparseable element
+rejects the whole request with a `400` naming the parameter, rather than silently dropping the bad
+element and keeping the good ones. For
+`[FromQuery] public Colour[]? Colours { get; set; }` on `GET /widgets`:
+
+| Request | Flag off (default) | Flag on |
+| --- | --- | --- |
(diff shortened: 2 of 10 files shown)


jeremydmiller:
Superseded by a clean branch — this one was accidentally cut on top of the GH-3385 branch in a shared working tree, so its diff carried three unrelated gRPC files. Same commit, rebased onto main. Reopening; no review time lost.

PR #3410, opened 2026-07-13
add cosmos fsharp frames

thechucklingatom:
add cosmos fsharp generation for sagas

### DIFF
--- a/src/Persistence/Wolverine.CosmosDb/Internals/CosmosDbPersistenceFrameProvider.cs
+++ b/src/Persistence/Wolverine.CosmosDb/Internals/CosmosDbPersistenceFrameProvider.cs
@@ -165,6 +165,14 @@ internal class CosmosDbUpsertFrame : AsyncFrame
             $"await {_container!.Usage}.UpsertItemAsync({_document.Usage}).ConfigureAwait(false);");
         Next?.GenerateCode(method, writer);
     }
+
+    public override void GenerateFSharpCode(GeneratedMethod method, ISourceWriter writer)
+    {
+        // UpsertItemAsync returns Task<ItemResponse<T>>, not Task; use let! _ = to discard the result.
+        writer.Write($"let! _ = {_container!.FSharpUsage}.UpsertItemAsync({_document.FSharpUsage})");
+        writer.Write("()");
+        Next?.GenerateFSharpCode(method, writer);
+    }
 }
 
 internal class CosmosDbDeleteDocumentFrame : AsyncFrame
@@ -193,6 +201,15 @@ internal class CosmosDbDeleteDocumentFrame : AsyncFrame
             $"await {_container!.Usage}.DeleteItemAsync<{_saga.VariableType.FullNameInCode()}>({_sagaId.Usage}, {typeof(PartitionKey).FullNameInCode()}.None).ConfigureAwait(false);");
         Next?.GenerateCode(method, writer);
     }
+
+    public override void GenerateFSharpCode(GeneratedMethod method, ISourceWriter writer)
+    {
+        // DeleteItemAsync returns Task<ItemResponse<T>>, not Task; use let! _ = to discard the result.
+        writer.Write(
+            $"let! _ = {_container!.FSharpUsage}.DeleteItemAsync<{_saga.VariableType.FSharpName()}>({_sagaId.FSharpUsage}, {typeof(PartitionKey).FSharpName()}.None)");
+        writer.Write("()");
+        Next?.GenerateFSharpCode(method, writer);
+    }
 }
 
 internal class CosmosDbDeleteByVariableFrame : AsyncFrame
@@ -217,5 +234,14 @@ internal class CosmosDbDeleteByVariableFrame : AsyncFrame
         writer.Write(
             $"await {_container!.Usage}.DeleteItemAsync<{_variable.VariableType.FullNameInCode()}>({_variable.Usage}.ToString(), {typeof(PartitionKey).FullNameInCode()}.None).ConfigureAwait(false);");
         Next?.GenerateCode(method, writer);
+    }
+
+    public override void GenerateFSharpCode(GeneratedMethod method, ISourceWriter writer)
+    {
+        // DeleteItemAsync returns Task<ItemResponse<T>>, not Task; use let! _ = to discard the result.
+        writer.Write(
+            $"let! _ = {_container!.FSharpUsage}.DeleteItemAsync<{_variable.VariableType.FSharpName()}>({_variable.FSharpUsage}.ToString(), {typeof(PartitionKey).FSharpName()}.None)");
+        writer.Write("()");
+        Next?.GenerateFSharpCode(method, writer);
     }
 }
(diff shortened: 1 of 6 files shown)


jeremydmiller (changes requested):
Thank you for this — and to be clear up front, **the frames themselves are good and I want to merge them.** I diffed your emitted F# against the checked-in generated C# for the equivalent Cosmos saga chain (`CosmosDbTests/Internal/Generated/WolverineHandlers/CompleteOneHandler1612253335.cs`) and it's a faithful frame-for-frame transliteration: load → try/catch-404 → null-guard → `UnknownSagaException` → SetSagaId → execute → `IsCompleted() ? Delete : Upsert` → flush. No semantic divergence. This is also the **first** `GenerateFSharpCode` saga-load support of any persistence provider — Marten, RavenDb, and EF Core all have none — so you're setting the precedent, not diverging from one.

Two things to fix first, both roughly one line, and both in the *sample and the test* rather than the frames.

## 1. The sample saga can't actually round-trip through the sample's own CosmosClient

`Domain.fs:45` declares `member val Id = "" with get, set`, and the doc comment at :30 says "the Id becomes the Cosmos document id." It won't.

`Program.fs:25` builds `CosmosClientOptions(ConnectionMode = ConnectionMode.Gateway)` with **no `SerializerOptions.PropertyNamingPolicy`**, so the default serializer emits `"Id"` — and Cosmos requires a lowercase `"id"`. The reason the C# Cosmos sagas work today is that `CosmosDbTests/AppFixture.cs:57` sets `PropertyNamingPolicy = CosmosPropertyNamingPolicy.CamelCase`. The sample doesn't.

You already knew this, which is what makes it worth catching: `Domain.fs:7` declares `Thing = { id: string; ... }` with the comment *"Cosmos identifies documents by a lowercase `id`."* The new saga contradicts the file's own established convention four lines later.

**Failure scenario:** send `StartThingSaga { Id = "abc" }` against the emulator with this host → `UpsertItemAsync` serializes `{"Id":"abc","Count":1,"Version":0}` → Cosmos 400, *"the required properties - 'id; ' - are missing"*. The saga never persists, the follow-up `ContinueThing` 404s on the point read, and you get `UnknownSagaException`. Silent saga death.

It's invisible today only because `Program.fs` never sends `StartThingSaga` and the compile gate never connects to Cosmos. Please add `SerializerOptions = CosmosSerializationOptions(PropertyNamingPolicy = CosmosPropertyNamingPolicy.CamelCase)` to the sample's client, matching `AppFixture` — I'd prefer that over annotating the property, because it also makes the sample match what real users have to do.

## 2. The test guard got weaker, and it guards the thing this PR adds

`CosmosFSharpCodegenSample.cs:66-69`. The previous code did `ChainFor(typeof(CreateThing)) ?? throw` — a hard assertion that a *specific* chain exists. The new code only asserts `chains.Length == 0` → throw.

So if saga chain discovery for F# types ever breaks (or `IncludeType<ThingSaga>()` stops resolving), `chains` comes back as just `[CreateThing]`, length 1, **the gate stays green**, `Generated.fs` silently loses both saga handlers, and `LoadDocumentFrame.GenerateFSharpCode` — the entire point of this PR — reverts to zero coverage with nobody noticing. Please assert `chains.Length.ShouldBe(3)` or add an explicit non-null assert on `ChainFor(typeof(ContinueThing))`.

## Notes, not blockers

- **CI does cover this**, for anyone wondering: the F# projects aren't in `wolverine.slnx`, but `wolverine_fsharp.slnx` is driven by `.github/workflows/fsharp.yml`, path-filtered on all three trees you touched, so `fsharp-codegen-compile-gate` fires. I regenerated `Generated.fs` and it's byte-identical to what you committed. Full `wolverine.slnx` and `wolverine_fsharp.slnx` both build clean.
- Be clear-eyed that the gate is **compile-only** — there's no start→persist→load→continue→complete round-trip for an F# Cosmos saga anywhere. That matches the existing bar for every other F# surface (EF Core, Marten, Http gates are all compile-only), so I won't block on it. But finding #1 is exactly the class of bug that walks straight through a compile gate.
- Your `[<AllowNullLiteral>]` workaround at `Domain.fs:40-42` is correct and well-commented. It's also a **framework-wide** F# constraint that your PR is the first to surface anywhere in the repo — the null guard comes from JasperFx's `IfElseNullGuardFrame` emitting `isNull x`, and F#'s `isNull` carries a `'T : null` constraint. I'm filing that upstream (emitting `isNull (box x)` would erase the constraint and make the attribute unnecessary) plus a docs issue, so you don't have to carry it.
- Dead line at `LoadDocumentFrame.cs:67` — `thingSaga <- Unchecked.defaultof<...>` in the `with` branch is already the default. Harmless, drop it if you're touching the file anyway.

Flip those two and I'll merge it into 6.18.0. Nice piece of work.

thechucklingatom:
Thank you for the quick feedback! I have pushed some new commits that address each piece of feedback.

32df64f Addressed the dead default line. I left this as `()` as I believe it needs that to compile.
f37f01c Adds the camel case policy to the cosmos client which allows it to round trip on its Id property.
5cbb2f0 Strengthens the test guard. I added a chain check for each value in the chain, which may be overkill, but will definitely show when something changes.

jeremydmiller (approved):
All three addressed, and verified rather than taken on trust — `wolverine_fsharp.slnx` builds clean (0 warnings, 0 errors) and the codegen compile gate passes.

**`f37f01c` (camelCase policy)** — exactly right, and it's the one that mattered. The sample saga can now actually round-trip through its own client instead of 400ing on a missing `id` and then dying as an `UnknownSagaException` on the follow-up message.

**`5cbb2f0` (test guard)** — stronger than I asked for. Exact chain count *and* a non-null assert per chain means saga discovery for F# types can't silently regress and quietly drop `LoadDocumentFrame` coverage. Not overkill; that gate is the only thing standing between this frame and zero coverage.

**`32df64f` (dead line)** — and **you're right that it needs the `()`**. I was wrong to call that line purely dead. F#'s `with` branch has to be a unit expression, and since `thingSaga` is already declared `Unchecked.defaultof<_>` *before* the `try`, `()` is precisely the correct body: the 404 leaves it null, and the `isNull` guard below raises `UnknownSagaException`. Good catch, and thank you for pushing back rather than just deleting it. Regenerating `Generated.fs` in the same commit was the right instinct too.

Merging. Note this lands **after** 6.18.0 was tagged, so it ships in the next release rather than this one — I'll call it out in those notes.

Three follow-ups came out of reviewing this, none of them yours to fix, all filed:

- **#3414** — Cosmos saga persistence has no optimistic concurrency. A blind `UpsertItemAsync` with no etag means two concurrent messages for one saga silently lose an update. Pre-existing on the C# path; your F# frames faithfully inherit it.
- **#3415** — `PartitionKey.None` puts every saga in the app into one logical partition (20GB / 10k RU/s ceiling). Self-consistent, so not a correctness bug, but a wall.
- **#3416** — the camelCase requirement you just hit is documented nowhere. That's on us, and it cost you a debugging cycle.
- **[jasperfx#513](https://github.com/JasperFx/jasperfx/issues/513)** — the `[<AllowNullLiteral>]` requirement your sample works around. `IfElseNullGuardFrame` should emit `isNull (box x)`, which erases the `'T : null` constraint and makes the attribute unnecessary for every F# saga, on every provider.

Genuinely nice contribution — first `GenerateFSharpCode` saga-load support of any Wolverine persistence provider. Thanks for the fast turnaround.

PR #3459, opened 2026-07-18
Proto-first gRPC client streaming via new StreamAsync<TRequest, TResponse> overload

erikshafer:
Closes the gap the gRPC work has been carrying since the streaming milestones (#2525, #2565): client streaming (`stream TRequest → TResponse`) was classified by `GrpcServiceChain` and then deliberately fail-fasted — the error text said "no adapter path **yet**", and `WolverineGrpcExceptionInterceptor` had a matching "deferred pending the matching `IMessageBus` overloads" note. This PR is that deferred work: the missing bus overload plus the codegen that consumes it. All four canonical RPC shapes are now generated on the proto-first path.

## The core primitive

`ICommandBus` gains the inverse overload of `StreamAsync<T>` — same name, per that comment's original "overloads" intent; arity separates the two directions (one type argument streams responses out, two stream requests in):

```csharp
Task<TResponse> StreamAsync<TRequest, TResponse>(
    IAsyncEnumerable<TRequest> messages, CancellationToken cancellation = default, TimeSpan? timeout = default);
// + DeliveryOptions overload
```

The pleasant surprise: almost no new machinery was needed. Handler discovery already accepts `IAsyncEnumerable<TRequest>` as a message type (first-parameter rule, no filter excludes closed generic interfaces), the chain compiles untouched, and `MessageFrame`'s cast is legal. The only thing actually missing was dispatch — `message.GetType()` on a stream is a compiler-generated iterator type, so the new method keys `Runtime.FindInvoker(typeof(IAsyncEnumerable<TRequest>))` explicitly and then reuses `Executor.InvokeAsync<T>` verbatim (envelope pooling, correlation, tracing, cascading flush all inherited). Handlers consume the live stream incrementally — nothing is materialized. Local invocation only, same contract as `StreamAsync`; a missing handler throws a `NotSupportedException` that spells out the expected handler signature instead of falling into remote routing.

I considered the alternative of draining to `TRequest[]` inside the generated wrapper and riding the existing `InvokeAsync<T>` — zero core change, and `T[]` is already a routable message type via the batching subsystem. Rejected it for semantics: the handler couldn't see item 1 until item N arrived, memory scales with stream length, and `Handle(TRequest[])` would silently overload the batching subsystem's meaning of that signature.

Heads-up on the one breaking edge: two new `ICommandBus` members, so custom `IMessageBus` implementors need a recompile-plus-implement (same shape as when `StreamAsync` landed). In-repo that was `TestMessageContext` and `SendingEnvelopeLifecycle`; called out in the CHANGELOG.

## The gRPC side

- `AssertNoUnsupportedStreamingKinds` is deleted outright — with client streaming generatable there are no unsupported kinds left. `ClientStreamingMethods` joins the classified lists, and the new `ForwardClientStreamToMessageBusFrame` emits a one-liner override:

  ```csharp
  return await _bus.StreamAsync<EchoRequest, EchoReply>(
      WolverineGrpcStreamAdapters.ReadAllAsync(requestStream, context.CancellationToken),
      context.CancellationToken);
  ```

- `WolverineGrpcStreamAdapters.ReadAllAsync` exists because the stock `Grpc.Core.AsyncStreamReaderExtensions` is declared by **both** Grpc.Core.Api and Grpc.Net.Common — generated code calling it is instantly CS0433 in any app referencing both. Wolverine's own trivial adapter sidesteps the collision.
- Before/after middleware and the `Validate` convention are **not** woven into client-streaming methods — same rationale as bidi, no single `TRequest` in scope. Tenant detection **is** covered (it only needs the trailing `ServerCallContext`, which every proto-first shape has).
- `WolverineGrpcExceptionInterceptor` gains the `ClientStreamingServerHandler` override its own doc comment was deferring — without it, a cancelled client-stream surfaced as `Unknown` instead of `Cancelled` (there's a test pinning that now). Bidi interception stays deferred; its wrapper streams responses incrementally, so trailing translation is a different problem.
- `GrpcRpcStreamKind.ClientStreaming` is appended (append-only, in case CritterWatch persists values) and the endpoint manifest surfaces the descriptors: `RequestType` is the per-item element type, response unwrapped from `Task<TResponse>`.

## Tests, sample, docs

- New `GrpcClientStreaming` end-to-end suite + `invoke_stream_async_support` core acceptance suite (empty stream, mid-drain fault, cancellation, cascading tuple, no-handler error, `DeliveryOptions`). The old fail-fast tests are inverted into construction-succeeds assertions. CoreTests 1931/1931 on net9.0 + net10.0, Wolverine.Grpc.Tests 281/281, `wolverine.slnx` Release build clean.
- `GreeterProtoFirstGrpc` gains `CollectGreetings (stream HelloRequest) → GreetingSummary` across proto/handler/client/README, verified by running server + client. While doing that I found the sample crashing at startup on 6.0 — it never picked up the `WolverineFx.RuntimeCompilation` reference during the Roslyn decoupling (the F# samples did). Fixed here for this sample; the other gRPC sample servers likely share the defect and are left for a follow-up sweep.
- Docs updated across the gRPC guide (`streaming` gets a full client-streaming section; the "unsupported" statements in `index`/`streaming`/`contracts` are gone), `message-bus.md` gets a "Streaming Requests" section beside "Streaming Responses", and the CHANGELOG gets its first gRPC entries.

## Not in scope

The code-first (protobuf-net.Grpc) generated-implementation path still skips `IAsyncEnumerable<TRequest>`-parameter methods — and its classifier would call that shape bidi regardless of return type. Documented as a limitation; fixing it is a separate, smaller change if anyone asks for it.

Smoke numbers, for flavor (Release, localhost, one connection): ~88k items/sec sustained through a 100k-item stream at a flat ~11 µs/item, and a complete 2-item client stream costs 212 µs vs 203 µs for a bare unary `SayHello` — the whole generated-wrapper + bus hop adds single-digit microseconds over the cheapest possible RPC.


erikshafer:
1. ~~I am not sure why I let the LLM start going with `Task<TResponse> InvokeStreamAsync<TRequest, TResponse>` instead of just adding the second arg such as `StreamAsync<TRequest, TResponse>`, like the `StreamAsync<T>` introduced a couple months ago. I think I'll change the name tomorrow. For now, gonna hit the hay and then give this another look Saturday or Sunday.~~ Completed. 
2. I am using `WolverineFx.Grpc` more and more heavily in CritterCab ( https://github.com/erikshafer/CritterCab ). I had a need for this feature I meant to get to earlier.  That's the out-of-repo test bed I have for a lot of the gRPC capabilities.

PR #3517, opened 2026-07-20
Update MQTT transport to MQTTnet 5

zhengchalei:
## Summary
- Update MQTT transport packages from `MQTTnet.Extensions.ManagedClient` 4.3.7 to MQTTnet 5.2.0 packages.
- Add a thin compatibility layer for the managed-client API surface Wolverine already uses, backed by MQTTnet v5's native `IMqttClient`.
- Adapt MQTT factory, server factory, payload, diagnostic endpoint, and enhanced-auth APIs for MQTTnet 5.

## Testing
- `dotnet build src/Transports/MQTT/Wolverine.MQTT/Wolverine.MQTT.csproj -v minimal`
- `dotnet build src/Transports/MQTT/Wolverine.MQTT.Tests/Wolverine.MQTT.Tests.csproj -v minimal`
- `dotnet test src/Transports/MQTT/Wolverine.MQTT.Tests/Wolverine.MQTT.Tests.csproj -f net10.0 --no-build -v minimal --filter "FullyQualifiedName!~Bug_mapper_exception_routes_to_dlq"`
  - Passed: 117, skipped: 1

## Notes
- Full `dotnet test` could not complete locally because this machine has .NET 8 and .NET 10 runtimes installed, but not .NET 9, and the SQL Server-backed dead-letter test needs a reachable local SQL Server.

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -70,7 +70,8 @@
     <PackageVersion Include="Microsoft.NET.Test.Sdk" Version="17.14.1" />
     <PackageVersion Include="Microsoft.SourceLink.GitHub" Version="8.0.0" />
     <PackageVersion Include="Microsoft.VisualStudio.Threading.Analyzers" Version="17.0.64" />
-    <PackageVersion Include="MQTTnet.Extensions.ManagedClient" Version="4.3.7.1207" />
+    <PackageVersion Include="MQTTnet" Version="5.2.0.1603" />
+    <PackageVersion Include="MQTTnet.Server" Version="5.2.0.1603" />
     <PackageVersion Include="MySqlConnector" Version="2.4.0" />
     <PackageVersion Include="NATS.Net" Version="2.8.2" />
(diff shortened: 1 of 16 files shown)


jeremydmiller:
Thank you for this @zhengchalei — the MQTTnet 5 migration here is solid work, especially the `ManagedMqttClientCompatibility` shim that re-implements the dropped managed-client API on top of v5's native `IMqttClient`.

We've decided **not** to upgrade the existing `WolverineFx.MQTT` package in place, for two reasons:

1. It would be a breaking change for everyone still on MQTTnet 4.
2. More fundamentally, a single assembly can't reference both MQTTnet 4 and 5 — both ship `MQTTnet.dll`, so there's no way to support both clients from one package.

Instead we're shipping a **parallel `WolverineFx.Mqtt5` package** so users can opt in, with the v4 package left intact. Your work here is cloned **verbatim** into that new package and you're credited as co-author on #3556. Once #3556 lands we'll close this one out.

Really appreciate the contribution 🙏

Issue #3380, opened 2026-07-12
OpenAPI: route parameters bound only by compound-handler LoadAsync/Before methods are missing from the operation (parameters: [])


jeremydmiller:
### Summary

Route parameters that are consumed **only** by a compound-handler `LoadAsync`/`Before` method are missing from the generated OpenAPI operation. Reported as a side note on #3374.

An endpoint whose route is

```
/v3/orders/{orderId:long}/order-lines/{orderLineId:long}
```

where both ids are bound by `LoadAsync` (e.g. `LoadAsync([FromRoute] long orderId, [FromRoute] long orderLineId, AppDbContext db, CancellationToken ct)`) and neither appears in the main handler method signature, produces an OpenAPI operation with

```json
"parameters": []
```

The path parameters are not declared at all, even though the route template requires them and the generated code binds (and 404s on) them.

### Root cause

`HttpChain.ApiDescription` derives the operation's parameters from the **main handler method signature** (plus the `[AsParameters]` container members), not from the route template plus the full binding chain. Middleware `MethodCall`s discovered via `ApplyImpliedMiddlewareFromHandlers` (`Load`/`LoadAsync`/`Before`/`Validate`...) contribute route/query/header bindings to codegen, but nothing walks them when the `ApiDescription` is assembled.

This is the same known weakness identified in the GH-3135 audit: OpenAPI metadata is derived from the handler signature rather than from what the chain actually binds. At minimum, every parameter in `RoutePattern.Parameters` should be declared as a required path parameter (with the type from whichever frame binds it, falling back to `string`), regardless of which method in the chain consumes it.

### Test coverage gap

Also flagged in the GH-3135 audit: there is essentially zero OpenAPI **shape**-test coverage (assertions on the rendered document's `parameters`/`requestBody` for representative endpoint shapes), which is why this class of omission keeps shipping. `AsParametersOpenApiEndpoints` + `asparameters route/type fidelity` tests started covering the `[AsParameters]` shapes; compound-handler-only bindings have no coverage at all.

### Not in scope of #3374's fix

The #3374 PR fixes the codegen duplicate-local compile failure only; this issue tracks the OpenAPI parameter fidelity separately, per the triage comment there.

jeremydmiller:
**The literal `parameters: []` symptom in this issue does not reproduce on `main`, and I want to correct that before it becomes folklore.** Fixed adjacent real defects in #3418 — details below.

## What I claimed vs. what's true

[...]

PR #3577, opened 2026-07-21
Skip the flattened container when describing chain-bound [FromQuery] parameters

### LINKED ISSUE #3575: Complex [FromQuery] type gets a duplicate whole-object OpenAPI parameter
**Describe the bug**

A `[FromQuery]` complex type is described twice: once as the flattened per-member query parameters
(correct), and again as a whole-object parameter named after the C# parameter. That entry has no wire
representation, and it pulls the container type into `components/schemas`.

**To Reproduce**

```csharp
public class OrderSearchQuery
{
    [FromQuery(Name = "filter")]     public string? Filter { get; set; }
    [FromQuery(Name = "pageNumber")] public int PageNumber { get; set; } = 1;
    public int PageSize { get; set; } = 50;
}

public static class ComplexQueryStringEndpoint
{
    [WolverineGet("/complex-query")]
    public static string Get([FromQuery] OrderSearchQuery search) => search.Filter!;
}
```

`AddOpenApi()` renders:

```json
"parameters": [
  { "name": "filter",     "in": "query", "schema": { "type": "string" } },
  { "name": "pageNumber", "in": "query", "schema": { "type": "integer", "format": "int32" } },
  { "name": "PageSize",   "in": "query", "schema": { "type": "integer", "format": "int32" } },
  { "name": "search",     "in": "query", "schema": { "$ref": "#/components/schemas/OrderSearchQuery" } }
]
```

plus `OrderSearchQuery` in `components/schemas`. `search` is the C# parameter name, not a query
string key.

**Expected behavior**

The three flattened members only, and no `OrderSearchQuery` in `components/schemas`.

**Additional context**

Two fill methods cover the same parameter:

- `fillQuerystringParameters` walks `_querystringVariables` — one entry per member. Correct.
- `fillChainBoundParameters` (#3418, for #3380) re-adds it. `allMethodCalls()` yields the endpoint
  method alongside the `Load`/`Before`/`After` calls #3380 was about, and it doesn't distinguish a
  `[FromQuery]` type that gets flattened from one bound as a single value.

`addChainBoundParameter`'s dedupe guard matches on name, and the composite (`search`) doesn't collide
with the query string keys (`filter`, `pageNumber`, `PageSize`).

There is a second manifestation when the flattening hasn't happened yet. Parameter matching runs in
the `HttpChain` constructor for the endpoint method and its compound-handler `Load`/`Before` methods,
but not for postprocessors — so a complex `[FromQuery]` on an `After` renders *only* the composite,
with no flattened members at all, and `ComplexQueryStringType` is never set for that chain.

The `[FromHeader]` branch doesn't share the gap: headers are never flattened, and
`GetOrCreateHeaderVariable` names the variable after the header key, so the dedupe guard matches.

`fillChainBoundParameters` came in with f34d7099 (#3418), first shipped in 6.18.0; 6.17.3 and earlier
are unaffected. Nothing public suppresses it — both methods are private, consult no
`WolverineHttpOptions` field, and `WolverineApiDescriptionProvider` only honors
`[ExcludeFromDescription]`, which hides the whole endpoint.

Stopgap for anyone hitting this: prune the composite in an `IApiDescriptionProvider` at
`Order > -1000` (Wolverine's own is `-2000`), matching on `HttpChain.ComplexQueryStringType`:

```csharp
foreach (var description in context.Results)
{
    if (description.ActionDescriptor is not WolverineActionDescriptor
        { Chain.ComplexQueryStringType: { } queryType }) continue;

    var composite = description.ParameterDescriptions.FirstOrDefault(p =>
        p.Source == BindingSource.Query && p.Type == queryType);

    if (composite is not null) description.ParameterDescriptions.Remove(composite);
}
```

That keys off `ComplexQueryStringType`, so it covers the endpoint-method and compound-handler shapes
but not the postprocessor one above.

It has to be at the `ApiDescription` layer — an OpenAPI operation transformer runs after the type is
registered in `components/schemas` and leaves it orphaned. Dropping `[FromQuery]` isn't a workaround:
the binding strategy requires the attribute, so the parameter falls through to
`JsonBodyParameterStrategy` and the endpoint binds a default-constructed object with no query
parameters and no request body. `[AsParameters]` avoids the composite but sets `IsFormData = true`.

esond:
Fixes #3575.

## What changed

`fillChainBoundParameters` now skips a `[FromQuery]` parameter whose type Wolverine binds by
flattening into per-member query string values. Those members are already described by
`fillQuerystringParameters`; the container never appears on the wire, so the extra whole-object
parameter was a phantom, and it pulled the container type into `components/schemas`.

`FromQueryAttributeUsage.TryMatch` already owns the simple-vs-complex distinction that decides
whether a `[FromQuery]` type gets flattened — it rejects simple types,
`DateTime`/`DateTimeOffset`/`DateOnly`/`TimeOnly`/`TimeSpan`/`Guid` and their nullable forms. That
predicate moved into `FromQueryAttributeUsage.IsComplexQueryStringType(Type)` so the binding side and
the description side read one definition. `TryMatch` behavior is unchanged.

Scoped to the `[FromQuery]` branch — `[FromHeader]` doesn't have the same gap, since headers are
never flattened and the header variable is named after the header key, so the existing dedupe guard
already matches.

## The postprocessor shape

One shape changes beyond the endpoint-method case in the issue, and it's worth being explicit about.

Compound handler `After` postprocessors are inserted into `Postprocessors` without parameter matching
(`Chain.cs:350-357` — note `OnException` and `Finally` immediately below both call
`ApplyParameterMatching`, and `After` doesn't). So a complex `[FromQuery]` on an `After` is never run
through `FromQueryAttributeUsage`, never gets a `QueryStringBindingFrame`, and is resolved by ordinary
variable resolution instead. It never reads the query string at all.

Today that shape renders *only* the composite. After this change it renders nothing for that
parameter — which is the accurate description, since no query string input is read. It's the same fix
applied to a second shape, not a regression: the composite documented a wire parameter the server
never looks at.

## Alternatives considered

- **`parameter.ParameterType == ComplexQueryStringType`.** Equivalent for the endpoint-method and
  compound-handler shapes, since both are matched in the `HttpChain` constructor and the field is set.
  Passed over because `ComplexQueryStringType` is single-valued (last write wins when a chain binds
  more than one container) and is never set for the postprocessor shape above, leaving that composite
  in place.
- **Excluding `Method` from `allMethodCalls()`.** Closest to #3380's intent, but it only fixes the
  endpoint-method case: a complex `[FromQuery]` on a compound handler's `Load` lives in `Middleware`,
  goes through the same strategy, and would still get the phantom entry.
  `complex_query_string_type_bound_only_by_a_compound_handler` pins exactly that.

## Tests

Three shapes added to the `openapi_shape_tests` harness, which renders the real
Microsoft.AspNetCore.OpenApi document:

- `/shapes/complex-query` — the flattened members only, and no `OrderSearchQuery` in
  `components/schemas`.
- `/shapes/complex-query-compound` — the same alongside a second query value bound by a compound
  handler.
- `/shapes/complex-query-on-load` — the container bound *only* by a compound handler `Load` and absent
  from the endpoint signature.

All three fail without the guard and pass with it (verified by reverting the hunk). Full
`Wolverine.Http.Tests` (873) and `Wolverine.Http.AspVersioning.Tests` (66) are green, and
`wolverine.slnx` builds clean in Release.

### DIFF
--- a/src/Http/Wolverine.Http.Tests.DifferentAssembly/OpenApiShapeEndpoints.cs
+++ b/src/Http/Wolverine.Http.Tests.DifferentAssembly/OpenApiShapeEndpoints.cs
@@ -126,7 +126,7 @@ public static class ComplexQueryStringEndpoint
         => $"{search.Filter}:{search.PageNumber}:{search.PageSize}";
 }
 
-// The flattened members and a GH-3380 query value bound only by a compound handler have to coexist.
+// The flattened members and a second query value bound by a compound handler have to coexist.
 public static class ComplexQueryStringWithCompoundHandlerEndpoint
 {
     public static string Load([FromQuery] string? audit) => audit ?? "none";
@@ -134,6 +134,20 @@ public static class ComplexQueryStringWithCompoundHandlerEndpoint
     [WolverineGet("/shapes/complex-query-compound")]
     public static string Get([FromQuery] OrderSearchQuery search, string audit)
         => $"{search.Filter}:{audit}";
+}
+
+// The container is bound ONLY by the compound handler and never appears in the endpoint signature.
+// Compound handler methods are parameter-matched with the endpoint method, so this flattens the same
+// way — and the container must not be described here either.
+public record OrderSearchResult(string Description);
+
+public static class ComplexQueryStringOnCompoundHandlerEndpoint
+{
+    public static OrderSearchResult Load([FromQuery] OrderSearchQuery search)
+        => new(search.Filter ?? "none");
+
+    [WolverineGet("/shapes/complex-query-on-load")]
+    public static string Get(OrderSearchResult result) => result.Description;
 }
 
 #endregion
--- a/src/Http/Wolverine.Http.Tests.DifferentAssembly/OpenApiShapeEndpoints.cs
+++ b/src/Http/Wolverine.Http.Tests.DifferentAssembly/OpenApiShapeEndpoints.cs
@@ -104,6 +104,40 @@ public static class UnboundRouteValueEndpoint
 
 #endregion
 
+#region [FromQuery] complex type shapes
+
+// A [FromQuery] complex type is flattened into one query parameter per member. The container itself has
+// no wire representation, so it must not also be declared as a parameter of its own.
+public class OrderSearchQuery
+{
+    [FromQuery(Name = "filter")]
+    public string? Filter { get; set; }
+
+    [FromQuery(Name = "pageNumber")]
+    public int PageNumber { get; set; } = 1;
+
+    public int PageSize { get; set; } = 50;
+}
+
+public static class ComplexQueryStringEndpoint
+{
+    [WolverineGet("/shapes/complex-query")]
+    public static string Get([FromQuery] OrderSearchQuery search)
+        => $"{search.Filter}:{search.PageNumber}:{search.PageSize}";
+}
+
+// The flattened members and a GH-3380 query value bound only by a compound handler have to coexist.
+public static class ComplexQueryStringWithCompoundHandlerEndpoint
+{
+    public static string Load([FromQuery] string? audit) => audit ?? "none";
+
+    [WolverineGet("/shapes/complex-query-compound")]
+    public static string Get([FromQuery] OrderSearchQuery search, string audit)
+        => $"{search.Filter}:{audit}";
+}
+
+#endregion
+
 #region [AsParameters] shapes
 
 public record OrderLineQuery([FromRoute] long OrderId, [FromQuery] string? Filter);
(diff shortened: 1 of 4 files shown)


esond:
Pushed a follow-up test commit after re-reading the alternatives I listed above.

The first two tests both put the container on the endpoint method, so they'd pass just as well if the
fix had only skipped `Method` in `allMethodCalls()`. That's one of the approaches I ruled out in the
description, and nothing in the suite actually held me to it. `/shapes/complex-query-on-load` puts the
container only on a compound handler `Load`, which lives in `Middleware`, so it fails under that
approach and passes under this one.

Same commit fixes a comment that credited the neighboring test's `audit` value to the chain-bound
path. `Load` gets parameter-matched in the `HttpChain` constructor, so that value actually comes from
`fillQuerystringParameters`.

Confirmed all three fail with the guard removed and pass with it back in, so none of them are green
for the wrong reason.

PR #3578, opened 2026-07-22
Stamp the assigned node onto owned event-store daemons (running_on_node, marten#5001)

jeremydmiller:
The Wolverine-managed-distribution half of [JasperFx/marten#5001](https://github.com/JasperFx/marten/issues/5001) — `running_on_node` stays NULL under managed distribution.

## What

`EventStoreAgents` carries the local `AssignedNodeNumber` and stamps it onto each daemon's `ShardStateTracker` when this node owns the daemon. `EventSubscriptionAgentFamily.BuildAgentAsync` (which has the `IWolverineRuntime`) sets it from `Options.Durability.AssignedNodeNumber` when this node is assigned an agent.

## Why

JasperFx's `ExtendedProgressionWriter` carries `ShardState.AssignedNodeNumber` into the persisted `running_on_node`, but nothing supplied that value on the runtime publish path — Wolverine owns the node assignment. A node's daemon only runs the agents assigned to *it*, so the local node number is the assigned node for every `ShardState` that daemon publishes. JasperFx **2.33.1** ([jasperfx#550](https://github.com/JasperFx/jasperfx/issues/550) / jasperfx#551) added the `ShardStateTracker.AssignedNodeNumber` stamp point this drives; the pin is bumped `2.31.0 → 2.33.1`.

This works for both the Marten and Polecat integrations (the change is in core `EventStoreAgents`, shared by both).

## Tests

`event_store_agents_observer_subscriptions`: stamps the assigned node onto the owned daemon's tracker; leaves it unset (0) when this node has no assignment. Existing subscription-lifecycle tests stay green.

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -44,9 +44,9 @@
     <PackageVersion Include="JasperFx.RuntimeCompiler" Version="5.0.0" />
     <PackageVersion Include="JasperFx.SourceGenerator" Version="2.33.1" />
     <PackageVersion Include="Lamar.Microsoft.DependencyInjection" Version="16.0.0" />
-    <PackageVersion Include="Marten" Version="9.16.1" />
+    <PackageVersion Include="Marten" Version="9.17.1" />
     <PackageVersion Include="Microsoft.Data.SqlClient" Version="6.1.3" />
-    <PackageVersion Include="Polecat" Version="[5.2.0,6.0.0)" />
+    <PackageVersion Include="Polecat" Version="[5.4.0,6.0.0)" />
     <PackageVersion Include="Microsoft.Azure.Cosmos" Version="3.46.1" />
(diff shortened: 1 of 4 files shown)


jeremydmiller:
CI red is transitive-version skew, not the diff: bumping the JasperFx pin to 2.33.1 while this repo still references the older published Marten 9.16.1 / Polecat 5.2.0 (built against pre-2.32 JasperFx) throws `TypeLoadException: PolecatProjectionDaemon.get_HighWaterLastPolledAt has no implementation` (that member was added to IProjectionDaemon in JasperFx.Events 2.32.0). IProjectionDaemon is identical between 2.32.0 and 2.33.1, and the latest published Marten 9.17.1 / Polecat 5.4.0 both target 2.32.0, so bumping those refs alongside the JasperFx pin resolves it. Marked draft pending the release-alignment decision.

PR #3585, opened 2026-07-22
Add runtime.AllLocallyOwnedDatabasesAsync(store) — first-class node-local database filter (#3584)

### LINKED ISSUE #3584: First-class node-local database filter: runtime.AllLocallyOwnedDatabasesAsync(store) — every AllDatabases() consumer re-derives ownership by hand
### Problem

Under Wolverine-managed event subscription distribution with database-affine assignment (GH-3340 / ZOD-style sharded tenancy, 512 shard databases), any code that walks `IEventStore.AllDatabases()` / `IMartenStorage.AllDatabases()` and touches each database fans out **one connection pool per database per node**. A node already holds pools for exactly the databases whose agents it owns; everything else is a fresh pool opened just to run a periodic query. At 512 databases × N pods this is the dominant connection-count driver on a shared database server (see CritterWatch#791, and jasperfx#553 for the write-side variant).

The ownership signal exists — `IAgentRuntime.AllLocallyOwnedDatabaseIds()` (GH-3340, #3345) — but every consumer currently re-derives the filter by hand:

- our application code parses `DatabaseId.Name` out of the id list and matches it against `database.Describe().DatabaseName` before reading projection progress;
- CritterWatch's dead-letter sweep (#791 fix) does a string-`Contains` between running agent URIs and `IEventDatabase.Identifier` (`EventProgressionPoller.ScopeToNodeLocalDatabases`) — which also ignores which *store* an agent belongs to;
- CritterWatch's progression/orphan/per-tenant-high-water passes are not scoped at all yet (filed separately on the CritterWatch repo).

Each re-derivation picks its own matching semantics and its own fallbacks, and none of them handle the store-scoping correctly (`AllLocallyOwnedDatabaseIds()` mixes the ids of every registered store).

### Proposal

A first-class, store-scoped helper in Wolverine core:

```csharp
IReadOnlyList<IEventDatabase> owned = await runtime.AllLocallyOwnedDatabasesAsync(store);
```

Semantics:

- zero/one database → returned as-is;
- `DurabilityMode.Solo` → all databases (the one node owns everything, and Solo runs the daemon in-process so there are no managed agent URIs to read);
- no `IEventSubscriptionAgentFamily` registered (managed distribution not in play) → all databases, preserving unscoped behavior;
- otherwise: filter the running managed-agent URIs to **this store's** `(type, name)` prefix, parse their `DatabaseId`s, and map them back to `IEventDatabase.Identifier` through the store's own usage descriptors (`TryCreateUsage` → `DatabaseDescriptor.ServerName/DatabaseName/Identifier` — the exact source `EventStoreAgents` built the agent URIs from, so the round trip cannot drift);
- a node that owns nothing gets an **empty** list (honest during startup/rebalance; readiness gates want to wait, monitoring sweeps can fall back to all themselves);
- never cached — assignments move as Wolverine rebalances.

Consumers that should adopt it once it exists: CritterWatch's `EventProgressionPoller` (all four per-database passes), `RebuildProgressPoller.LocateAsync`, and application-level readiness/uniqueness sweeps like ours.

PR incoming.

erdtsieck:
Closes #3584.

Adds `runtime.AllLocallyOwnedDatabasesAsync(store)` to Wolverine core — the canonical, store-scoped filter over `IEventStore.AllDatabases()` returning only the databases whose event-subscription agents run on THIS node under Wolverine-managed distribution.

Semantics (documented on the method):

- zero/one database → returned as-is;
- `DurabilityMode.Solo` → all databases (one node owns everything; Solo runs the daemon in-process, so there are no managed agent URIs to read ownership from);
- no `IEventSubscriptionAgentFamily` registered → all databases (ownership doesn't exist as a concept; unscoped behavior preserved);
- otherwise: running managed-agent URIs filtered to this store's `(type, name)` prefix → `DatabaseId`s → mapped back to `IEventDatabase.Identifier` through the store's usage descriptors (`TryCreateUsage`), the same source `EventStoreAgents` builds the agent URIs from, so the round trip cannot drift;
- a node that owns nothing gets an **empty** list — honest during startup or a rebalance (readiness gates want to wait; monitoring sweeps that must never go dark can fall back to all databases themselves);
- fail-open when the store yields no usage descriptor (better every database than silently none);
- results must never be cached — assignments move as Wolverine rebalances.

Seven unit tests in `CoreTests.Runtime.Agents.all_locally_owned_databases` cover each branch, including the store-scoping (agents of another store never count as ownership — `AllLocallyOwnedDatabaseIds()` mixes stores, this helper doesn't).

First consumer PRs after this lands: CritterWatch's `EventProgressionPoller` (its dead-letter sweep already hand-rolls this filter with a string-`Contains` since #791; its progression/orphan/per-tenant-high-water passes are unscoped entirely) and application-level readiness gates.

### DIFF
--- a/src/Testing/CoreTests/Runtime/Agents/all_locally_owned_databases.cs
+++ b/src/Testing/CoreTests/Runtime/Agents/all_locally_owned_databases.cs
@@ -1,0 +1,186 @@
+using JasperFx;
+using JasperFx.Descriptors;
+using JasperFx.Events;
+using JasperFx.Events.Descriptors;
+using JasperFx.Events.Projections;
+using Microsoft.Extensions.DependencyInjection;
+using NSubstitute;
+using Shouldly;
+using Wolverine.Runtime;
+using Wolverine.Runtime.Agents;
+using Xunit;
+
+namespace CoreTests.Runtime.Agents;
+
+/// <summary>
+/// Coverage for <see cref="EventStoreOwnershipExtensions.AllLocallyOwnedDatabasesAsync" /> — the
+/// first-class "which databases does THIS node own" filter over <see cref="IEventStore.AllDatabases" />.
+/// Recurring per-node work against a multi-database store (progress queries, telemetry polls, readiness
+/// gates) must scope to the databases whose event-subscription agents run locally instead of fanning a
+/// connection pool out to every shard database from every node (GH-3340, CritterWatch#791).
+/// </summary>
+public class all_locally_owned_databases
+{
+    private static readonly EventStoreIdentity TheIdentity = new("main", "marten");
+    private static readonly EventStoreIdentity OtherIdentity = new("other", "marten");
+
+    private readonly IWolverineRuntime _runtime = Substitute.For<IWolverineRuntime>();
+    private readonly IAgentRuntime _agents = Substitute.For<IAgentRuntime>();
+    private readonly IEventStore _store = Substitute.For<IEventStore>();
+    private readonly WolverineOptions _options = new();
+
+    public all_locally_owned_databases()
+    {
+        _options.Durability.Mode = DurabilityMode.Balanced;
+        _runtime.Options.Returns(_options);
+        _runtime.Agents.Returns(_agents);
+        _store.Identity.Returns(TheIdentity);
+
+        withFamilyRegistered(true);
+    }
+
+    private void withFamilyRegistered(bool registered)
+    {
+        var services = new ServiceCollection();
+        if (registered)
+        {
+            services.AddSingleton(Substitute.For<IEventSubscriptionAgentFamily>());
+        }
+
+        _runtime.Services.Returns(services.BuildServiceProvider());
+    }
+
+    private static IEventDatabase database(string identifier)
+    {
+        var db = Substitute.For<IEventDatabase>();
+        db.Identifier.Returns(identifier);
+        return db;
+    }
+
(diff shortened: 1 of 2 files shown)


jeremydmiller:
Thanks @erdtsieck — the semantics here are well thought through and this is genuinely useful. Before it lands I want to resolve how it sits next to what core already exposes.

Wolverine already has `IWolverineRuntime.Agents.AllLocallyOwnedDatabaseIds()` (`src/Wolverine/Runtime/WolverineRuntime.Agents.cs:94`): unscoped, returns `IReadOnlyList<DatabaseId>`, lives on `.Agents`. This PR adds `runtime.AllLocallyOwnedDatabasesAsync(store)`: store-scoped, returns `IReadOnlyList<IEventDatabase>`, as an extension on `runtime`. Shipping both would leave two near-homonyms sitting right next to each other — and as this PR's own tests point out, the existing one is store-blind, which is exactly the gap the new method closes.

I'd rather decide the API shape deliberately: whether the store-scoped filter should be a first-class member alongside (or replacing) `AllLocallyOwnedDatabaseIds()`, with the old one cross-referenced or deprecated, rather than adding a parallel extension. There's also a real first consumer to validate the signature against (CritterWatch's `EventProgressionPoller`, #791) before we freeze the public surface.

Going to close this for now while we settle that direction — not a rejection of the idea, which we do want. Will follow up with the resolved shape. Thanks again!

PR #3587, opened 2026-07-22
fix(http): don't describe a route-bound [FromQuery]/[FromHeader] parameter twice in OpenAPI (#3586)

### LINKED ISSUE #3586: OpenAPI: [FromQuery] on a route-template parameter emits a duplicate parameter and 500s /openapi/*.json
## Summary

When a Wolverine HTTP endpoint has a parameter that is **both** a route-template segment **and** annotated `[FromQuery]`, the generated `ApiDescription` emits that parameter **twice** — once with `Source=Path` (from the route template) and once with `Source=Query` (from `[FromQuery]`).

```csharp
public static class ThingEndpoints
{
    [WolverineGet("/things/{id}")]
    public static string Get([FromQuery] string id) => id;
}
```

Two parameters with the same name is invalid OpenAPI. In practice it also crashes .NET's built-in OpenAPI XML-documentation transformer, whose per-`<param>` lookup does:

```csharp
operation.Parameters.SingleOrDefault(p => p.Name == comment.Name)
```

throwing `InvalidOperationException: Sequence contains more than one matching element` and failing **all** of `/openapi/{document}.json` with HTTP 500 — not just the offending endpoint.

## Impact

Any project that adds XML-doc comments (`AddOpenApi()` + `IncludeXmlComments` / the .NET 9+ XML transformer) and has even one such parameter loses its entire OpenAPI document.

## Root cause

At runtime the `[FromQuery]` on a simple type here is a **no-op**. Parameter matching (`HttpGraph.ParameterMatching.cs`) tries strategies in a fixed order:

- `FromQueryAttributeUsage` declines simple types (only complex containers flatten), so it returns `false`.
- `RouteParameterStrategy` runs **before** `QueryStringParameterStrategy` and matches purely by name, so it claims `id` and binds it from the **route value**.

The generated handler confirms it — a route-value read, no query access:

```csharp
var id = (string?)httpContext.GetRouteValue("id");
if (id == null) { httpContext.Response.StatusCode = 404; return; }
var result_of_Get = ThingEndpoints.Get(id);
```

The duplicate is purely in the OpenAPI **description** path. `CreateApiDescription` correctly adds a `Path` parameter from the route template, but `fillChainBoundParameters` (introduced in #3418 / GH-3380, deriving parameters from the full binding chain) then walks the endpoint method, sees `[FromQuery]`, and adds a second `Query` parameter — describing a binding that never happens.

This regressed with GH-3380 (between 6.16 and 6.21).

## Expected

The OpenAPI description should reflect the actual binding: a single `Path` parameter `id`. The phantom `Query` entry must not be emitted.

## Note / follow-up

Separately from this description bug: `[FromQuery]` (and `[FromHeader]`) on a parameter whose name collides with a route segment is **silently ignored** at binding time — the route wins with no diagnostic. Emitting an error during codegen for this conflicting configuration, instead of silently preferring the route value, would be friendlier and is worth considering as a follow-up.

## Reproduction

Self-verifying file-based app under `repro/duplicate-fromquery-route-parameter.cs`:

```
dotnet run repro/duplicate-fromquery-route-parameter.cs
```

Against `main` the assertion fails with `2` (`sources: Path, Query`).

Blackclaws:
## What

A Wolverine HTTP parameter that is **both** a route-template segment and annotated `[FromQuery]` (or `[FromHeader]`) was emitted **twice** in the `ApiDescription` — once `Source=Path` (route template) and once `Source=Query`/`Header`. Two same-name parameters are invalid OpenAPI and crash the .NET XML-doc operation transformer (`operation.Parameters.SingleOrDefault(p => p.Name == comment.Name)` → *"Sequence contains more than one matching element"*), returning HTTP 500 for the whole `/openapi/{document}.json`.

```csharp
[WolverineGet("/things/{id}")]
public static string Get([FromQuery] string id) => id;
```

Closes #3586.

## Root cause

At runtime the `[FromQuery]` here is a no-op: `FromQueryAttributeUsage` declines simple types and `RouteParameterStrategy` runs before the query/header strategies, so the route claims `id` and binds it from the route value. The generated handler proves it — a route-value read, no query access:

```csharp
var id = (string?)httpContext.GetRouteValue("id");
```

The duplicate lived only in the OpenAPI description path: `fillChainBoundParameters` (added in #3418 / GH-3380) walked the endpoint method, saw `[FromQuery]`, and added a second parameter alongside the `Path` one the route-template loop already emits. Regressed between 6.16 and 6.21.

## The fix

Skip a route-bound parameter in `fillChainBoundParameters` (name matches a route segment + route-bindable type, mirroring `RouteParameterStrategy`/`FindRouteVariable`). **Description-only change — runtime binding is unchanged.**

The OpenAPI document now reflects what actually happens: a single `Path` parameter. To be explicit about the behavior this encodes — **the `[FromQuery]` is silently ignored, exactly as it already was at binding time.** This PR only makes the emitted OpenAPI tell that truth; it does not change how the value binds.

## Follow-up worth discussing

That silent precedence is arguably a footgun: a user who writes `[FromQuery]`/`[FromHeader]` on a parameter whose name collides with a route segment gets route binding with no diagnostic. It may be friendlier to **emit an error during codegen** for this conflicting configuration rather than silently accepting it and preferring the route value. That's a separate behavioral change from this description fix, so I've left it out here — happy to open a follow-up if maintainers agree.

## Testing

Self-verifying file-based repro added at `repro/duplicate-fromquery-route-parameter.cs`:

```
dotnet run repro/duplicate-fromquery-route-parameter.cs
```

- Repro passes (`sources: Path`, count 1); the sibling `kebab-cased-route-parameter.cs` still passes.
- Existing `openapi_shape_tests` (16) and all `open_api`/`openapi` tests (60) pass.
- `Wolverine.Http` builds with 0 warnings/errors.

### DIFF
--- a/src/Http/Wolverine.Http/HttpChain.ApiDescription.cs
+++ b/src/Http/Wolverine.Http/HttpChain.ApiDescription.cs
@@ -625,6 +625,18 @@ public partial class HttpChain
         {
             foreach (var parameter in call.Method.GetParameters())
             {
+                // A simple [FromQuery]/[FromHeader] parameter whose name matches a route-template segment is
+                // actually bound from the route value, not the query string or a header: FromQueryAttributeUsage
+                // declines simple types and RouteParameterStrategy runs before the query/header strategies, so
+                // the route claims it and the attribute is a no-op. The route-template loop already describes it
+                // as a Path parameter; describing it again here would emit a second same-name parameter, which is
+                // invalid OpenAPI and crashes downstream transformers (e.g. the XML-doc operation transformer's
+                // Parameters.SingleOrDefault). Regressed when GH-3380 began deriving parameters from the chain.
+                if (isBoundFromRouteValue(parameter))
+                {
+                    continue;
+                }
+
                 if (parameter.TryGetAttribute<FromQueryAttribute>(out var fromQuery))
                 {
                     // A complex [FromQuery] type is bound by flattening it into one query string value per
@@ -646,6 +658,22 @@ public partial class HttpChain
                 }
             }
         }
+    }
+
+    /// <summary>
+    /// Would this endpoint/chain parameter be bound from a route value rather than the query string or a
+    /// header? RouteParameterStrategy matches by the parameter's own name (case-sensitively, mirroring
+    /// <see cref="FindRouteVariable(ParameterInfo, out Variable?)"/>) and runs before the query/header
+    /// strategies, so a name collision with a route-template segment on a route-bindable type makes any
+    /// [FromQuery]/[FromHeader] attribute a no-op. Such a parameter is already described by the
+    /// route-template loop and must not be described again as chain-bound. See GH-3380.
+    /// </summary>
+    private bool isBoundFromRouteValue(ParameterInfo parameter)
+    {
+        // An explicit [FromRoute] parameter is route-bound by definition and never treated as chain-bound
+        // query/header anyway; only the implicit name-collision case needs guarding here.
+        return RoutePattern!.Parameters.Any(x => x.Name == parameter.Name)
+               && isBindableRouteType(parameter.ParameterType);
     }
 
     private static void addChainBoundParameter(ApiDescription apiDescription, string name, Type parameterType,
--- a/repro/README.md
+++ b/repro/README.md
@@ -16,3 +16,15 @@ single-file apps.
 `[FromRoute(Name = "journey-id")]` on an endpoint method parameter. Against
 `main` the first assertion fails with `0`, because the attribute's name was
 ignored and the argument was bound from the query string instead.
+
+## duplicate-fromquery-route-parameter.cs
+
+`[FromQuery]` on a parameter whose name is also a route-template segment
+(`[WolverineGet("/things/{id}")] Get([FromQuery] string id)`). Against `main`
+the assertion fails with `2`: the parameter is emitted twice in the endpoint's
+`ApiDescription` — once `Source=Path` (from the route template) and once
+`Source=Query` (from `[FromQuery]`). Two same-name parameters are invalid
+OpenAPI and crash `Microsoft.AspNetCore.OpenApi`'s XML-comment operation
+transformer (`operation.Parameters.SingleOrDefault(p => p.Name == comment.Name)`
+→ "Sequence contains more than one matching element"), returning HTTP 500 for
+all of `/openapi/{document}.json`. This regressed between 6.16 and 6.21.
--- a/repro/duplicate-fromquery-route-parameter.cs
+++ b/repro/duplicate-fromquery-route-parameter.cs
@@ -1,0 +1,76 @@
+#:sdk Microsoft.NET.Sdk.Web
+#:project ../src/Http/Wolverine.Http/Wolverine.Http.csproj
+#:project ../src/Wolverine.RuntimeCompilation/Wolverine.RuntimeCompilation.csproj
+#:property TargetFramework=net10.0
+#:property PublishAot=false
+#:property JsonSerializerIsReflectionEnabledByDefault=true
+
+using Microsoft.AspNetCore.Builder;
+using Microsoft.AspNetCore.Mvc;
+using Microsoft.AspNetCore.Mvc.ApiExplorer;
+using Microsoft.Extensions.DependencyInjection;
+using Wolverine;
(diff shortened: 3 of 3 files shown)


src/Http/Wolverine.Http/HttpChain.ApiDescription.cs
```diff
@@ -648,6 +660,22 @@ private void fillChainBoundParameters(ApiDescription apiDescription)
...
 
+    /// <summary>
+    /// Would this endpoint/chain parameter be bound from a route value rather than the query string or a
+    /// header? RouteParameterStrategy matches by the parameter's own name (case-sensitively, mirroring
+    /// <see cref="FindRouteVariable(ParameterInfo, out Variable?)"/>) and runs before the query/header
+    /// strategies, so a name collision with a route-template segment on a route-bindable type makes any
+    /// [FromQuery]/[FromHeader] attribute a no-op. Such a parameter is already described by the
+    /// route-template loop and must not be described again as chain-bound. See GH-3380.
+    /// </summary>
+    private bool isBoundFromRouteValue(ParameterInfo parameter)
```
esond, line 671: Keep the case-sensitive comparison, it's right. Worth a note on why though, because it's pinned to something that might change.

This is close to `tryDetermineRouteBinding` at line 574, which already pairs `matchesRouteName` with `isBindableRouteType`. Difference is `matchesRouteName` uses `EqualsIgnoreCase` and you compare case-sensitively. You're mirroring `FindRouteVariable(ParameterInfo)` (`HttpChain.cs:781`, `x.Name == parameter.Name`), which is the path that actually claims the parameter, so that's correct. Going case-insensitive here would suppress a query parameter the generated code really does read.

The catch is the binder looks out of step with ASP.NET. `RouteValueDictionary` and `QueryCollection` are both `OrdinalIgnoreCase`, and minimal APIs match a parameter name to a route template with `routeParams.Contains(parameter.Name, StringComparer.OrdinalIgnoreCase)` in `RequestDelegateFactory`. Wolverine isn't consistent with itself either: the explicit `[FromRoute(Name = ...)]` path (`FindRouteVariable(Type, string)`, line 826) and `matchesRouteName` both use `EqualsIgnoreCase`, only the implicit name match is case-sensitive. In practice `[WolverineGet("/things/{Id}")] Get(string id)` renders as path `Id` plus query `id` today, so the route segment goes unbound and the value comes from the query string. Minimal APIs bind that from the route.

So a line saying the case sensitivity deliberately tracks `FindRouteVariable`, and has to move with it if that ever gets aligned, would stop someone unifying this with `matchesRouteName` and quietly breaking it.

esond (commented):
Apologies if an ad-hoc review from someone not super involved with the project is unwelcome. I noticed yours today while checking on my own related PR (#3577), and since it lands in the same method I had some context worth passing along.

Fix looks right to me. Your root cause read matches what I found: `RouteParameterStrategy` sits ahead of both `FromHeaderStrategy` and `QueryStringParameterStrategy` in `HttpGraph._strategies`, so the route genuinely wins and the attribute is a no-op. I also checked that the duplicate predates #3577, so this is the GH-3380 regression you identified and nothing my change introduced.

Left one inline note on the case sensitivity in `isBoundFromRouteValue`. Short version, keep it as is, but it turned up a separate problem: `FindRouteVariable(ParameterInfo)` matches a parameter name to a route segment case-sensitively, while ASP.NET is `OrdinalIgnoreCase` everywhere comparable, including minimal APIs' own version of that comparison. Visible effect is that `[WolverineGet("/things/{Id}")] Get(string id)` binds from the query string instead of the route. Can open a separate issue for that if you want.

@jeremydmiller, broader thing since I've been in this method recently. `fillQuerystringParameters` describes from `_querystringVariables`, which is what the binder actually produced. `fillChainBoundParameters` re-derives from attributes instead, which is why it keeps describing bindings that never happen. That's three subtractive guards in one method now across two PRs: the dedupe in `addChainBoundParameter`, the GH-3575 flattening skip, and this one. Describing chain-bound parameters from the bound variables, the way the query string path already does, would make this class of bug structurally impossible. Separate from this PR, just flagging it.

On the follow-up idea, agree a codegen diagnostic beats silent route precedence, but it would break anyone currently shipping a harmless `[FromQuery]` on a route parameter, so it probably needs to be a warning or opt-in rather than an error.

Issue #3589, opened 2026-07-23
[Oracle] Main message store uses unregistered 'wolverine' URI scheme, preventing durability agent startup


pedroandrade03:
## Describe the bug

When using `WolverineFx.Oracle` 6.21.0 with `DurabilityMode.Solo`, Wolverine repeatedly fails to start the Oracle durability agent:

```text
fail: Wolverine.Runtime.Agents.NodeAgentController[0]
      Error trying to reevaluate agent assignments for 'wolverinedb' agents
      System.ArgumentOutOfRangeException: Unrecognized agent scheme 'wolverine' (Parameter 'uri')
         at Wolverine.Runtime.Agents.NodeAgentController.findAgentAsync(Uri uri)
         at Wolverine.Runtime.Agents.NodeAgentController.StartAgentAsync(Uri agentUri)
         at Wolverine.Runtime.Agents.NodeAgentController.startAllAgentsAsync()
```

The host continues running, but the Oracle durability agent does not start.

## Versions

- .NET 9.0
- WolverineFx 6.21.0
- WolverineFx.Oracle 6.21.0

## Minimal reproduction

```csharp
using JasperFx;
using Wolverine;
using Wolverine.Oracle;

var builder = Host.CreateApplicationBuilder(args);

builder.UseWolverine(options =>
{
    options.Durability.Mode = DurabilityMode.Solo;
    options.AutoBuildMessageStorageOnStartup = AutoCreate.CreateOrUpdate;

    options.PersistMessagesWithOracle(
        Environment.GetEnvironmentVariable("DB_CONNECTION")!,
        "WOLVERINE_AGENT_REPRO"
    );
});

await builder.Build().RunAsync();
```

Start the application against a reachable Oracle database.

The exception is logged on startup and again on every agent-assignment reevaluation cycle.

## Root cause

`MessageStoreCollection` is registered as an agent family using the `wolverinedb` scheme:

https://github.com/JasperFx/wolverine/blob/V6.21.0/src/Wolverine/Persistence/MessageStoreCollection.cs#L295-L300

It advertises agents by returning each message store's `Uri`.

However, `OracleMessageStore` assigns the main store this URI:

```csharp
Uri = new Uri("wolverine://messages/main");
```

This happens both in the constructor and in `PromoteToMain()`:

https://github.com/JasperFx/wolverine/blob/V6.21.0/src/Persistence/Oracle/Wolverine.Oracle/OracleMessageStore.cs#L112-L119

[...]

PR #3615, opened 2026-07-24
Fix Oracle durability operation batch execution

### LINKED ISSUE #3614: [Oracle] Durability agent cannot recover inbox/outbox because generated batch SQL is incompatible with ODP.NET
**Describe the bug**

Wolverine 6.22.0 fixes #3589: the `wolverinedb` agent URI is now recognized and the durability agent starts successfully.

However, once the agent starts, the Oracle-backed durability operations repeatedly fail with `ORA-00933` and `ORA-00936`.

The generated commands appear to:

- concatenate multiple SQL statements into a single `OracleCommand`;
- use generic `@parameter` placeholders in SQL executed by ODP.NET.

Because of this, persisted outgoing messages are not recovered after the application restarts. The message remains in `wolverine_outgoing_envelopes` with `owner_id = 0`.

This appears to be a separate issue from #3589. Version 6.22.0 fixes the agent URI problem, but then exposes this Oracle SQL incompatibility when the agent begins executing its recovery operations.

**To Reproduce**

Packages:

- `WolverineFx` 6.22.0
- `WolverineFx.Oracle` 6.22.0
- `WolverineFx.RabbitMQ` 6.22.0

Infrastructure:

- Oracle XE `gvenzl/oracle-xe:21.3.0-slim-faststart`
- RabbitMQ `rabbitmq:3.11-management-alpine`
- Application targeting `net9.0`

Relevant configuration:

```csharp
builder.UseWolverine(options =>
{
    options.Durability.Mode = DurabilityMode.Solo;
    options.AutoBuildMessageStorageOnStartup =
        AutoCreate.CreateOrUpdate;

    options.PersistMessagesWithOracle(
        oracleConnectionString,
        "WOLVERINE_AGENT_REPRO");

    var rabbitMq = options.UseRabbitMq(factory =>
    {
        factory.HostName = rabbitHost;
        factory.Port = rabbitPort;
        factory.UserName = rabbitUsername;
        factory.Password = rabbitPassword;
    });

    rabbitMq.AutoProvision();
    rabbitMq.UseSenderConnectionOnly();

    options.PublishMessage<DurableMessage>()
        .ToRabbitRoutingKey(
            "durability.repro",
            "durability.repro.v1")
        .UseDurableOutbox();
});

public sealed record DurableMessage(Guid Id);
```

Steps:

1. Start Oracle and RabbitMQ.
2. Start the application.
3. Stop RabbitMQ.
4. Publish a durable message.
5. Verify that the message exists in `wolverine_outgoing_envelopes`.
6. Stop the application.
7. Start RabbitMQ again.
8. Restart the application using the same Oracle schema.
9. Observe the durability agent logs.
10. Query `wolverine_outgoing_envelopes` again.

The durability agent starts, but its database batches fail with errors similar to:

```text
Wolverine.RDBMS.Polling.DatabaseBatchCommandException:
Database operation batch failure:

1. Wolverine.RDBMS.Durability.CheckRecoverableIncomingMessagesOperation
2. Wolverine.RDBMS.Durability.CheckRecoverableOutgoingMessagesOperation
3. Wolverine.RDBMS.Durability.MoveReplayableErrorMessagesToIncomingOperation
4. Wolverine.RDBMS.Durability.DeleteOldNodeEventRecords

Oracle.ManagedDataAccess.Client.OracleException:
ORA-00933: SQL command not properly ended
```

The generated batch contains multiple statements and `@` parameters, for example:

```sql
select received_at, count(*)
from WOLVERINE_AGENT_REPRO.wolverine_incoming_envelopes
where status = 'Incoming' and owner_id = 0
group by received_at;

select distinct destination
from WOLVERINE_AGENT_REPRO.wolverine_outgoing_envelopes
where owner_id = 0;

insert into ...
where replayable = @replayable;

delete from ...
where timestamp < @p1;
```

A separate durability operation also fails:

```text
Wolverine.RDBMS.Durability.DeleteExpiredEnvelopesOperation

delete from WOLVERINE_AGENT_REPRO.wolverine_incoming_envelopes
where status = 'Handled' and keep_until <= @p0;

Oracle.ManagedDataAccess.Client.OracleException:
ORA-00936: missing expression
```

The persisted outgoing message remains unrecovered:

```sql
select owner_id, destination, attempts
from WOLVERINE_AGENT_REPRO.wolverine_outgoing_envelopes;
```

Example result:

```text
OWNER_ID = 0
ATTEMPTS = 0
```

**Expected behavior**

After RabbitMQ becomes available and the application restarts, the durability agent should:

1. Detect the persisted outgoing message.
2. Send it to RabbitMQ.
3. Remove or mark the persisted envelope as successfully processed.

Oracle durability operations should use Oracle-compatible parameter binding and command execution.

**Potential source of the problem**

From inspecting the 6.22.0 source:

- `OracleMessageStore.ToCommandBuilder()` returns the generic `Weasel.Core.DbCommandBuilder`.
- The generic builder produces `@` parameter placeholders, while the Oracle-specific builder uses `:` parameters.
- `DatabaseOperationBatch` configures several operations into one compiled command.
- `MoveReplayableErrorMessagesToIncomingOperation` also contains a hard-coded `@replayable` parameter.

ODP.NET does not accept the generated multi-statement SQL batch as a regular `OracleCommand`.

A possible fix may require:

- using Oracle-compatible parameter markers;
- removing hard-coded `@` parameters from shared durability operations;
- executing the durability operations separately within the same transaction, or generating an Oracle-compatible block.

**Additional context**

The original `Unrecognized agent scheme 'wolverine'` exception from #3589 no longer occurs in 6.22.0. The `wolverinedb` agent now starts, so that fix is working.

The failure happens afterward, when the started agent attempts to poll and recover persisted messages from Oracle.

pedroandrade03:
## Summary

- allow message database providers to opt into provider-specific durability batch execution
- execute Oracle durability operations as individual commands in a single transaction
- translate generic `@name` bind markers to Oracle `:name` markers and normalize boolean and `Guid` parameter values
- add focused coverage for Oracle command translation

## Background

#3589 corrected the Oracle message store agent URI, which allows the durability agent to start. Once running, the agent exposed a second Oracle-specific failure: generic RDBMS durability batches concatenate multiple SQL statements into one command and use `@` bind markers. ODP.NET rejects those commands with `ORA-00933` and `ORA-00936`, so persisted inbox/outbox messages are not recovered.

## Root cause and fix

The shared `DatabaseOperationBatch` implementation assumes that a provider can execute several result sets from one multi-statement command. Oracle does not support that command shape through ODP.NET.

This change adds an internal provider hook without changing Wolverine's public API. `OracleMessageStore` uses that hook to configure each existing `IDatabaseOperation` independently, execute the statements in order within one Oracle transaction, invoke its result callback, and then keep the existing post-processing behavior.

The Oracle executor also:

- enables bind-by-name
- translates generic `@` markers to `:`
- maps booleans to `NUMBER(1)`-compatible values
- maps `Guid` values to Oracle `RAW`

Other database providers continue through the existing batching path.

Fixes #3614.

## Validation

- `dotnet test src/Persistence/Oracle/OracleTests/OracleTests.csproj --no-restore --filter "FullyQualifiedName~oracle_durability_command_translation"`  
  Passed 3 tests on `net9.0` and 3 tests on `net10.0`.
- External downstream recovery scenario against Oracle XE and RabbitMQ: persisted a durable outgoing message while RabbitMQ was unavailable, restarted the application, then verified that the durability agent recovered the outbox row and delivered the message.

This is opened as a draft to allow discussion of the provider-specific execution hook and test coverage.

### DIFF
--- a/src/Persistence/Oracle/OracleTests/oracle_durability_command_translation.cs
+++ b/src/Persistence/Oracle/OracleTests/oracle_durability_command_translation.cs
@@ -1,0 +1,46 @@
+using Oracle.ManagedDataAccess.Client;
+using Shouldly;
+using Wolverine.Oracle;
+
+namespace OracleTests;
+
+public class oracle_durability_command_translation
+{
+    [Fact]
+    public void rewrites_generic_parameter_markers_for_oracle()
+    {
+        OracleMessageStore.normalizeParameterMarkers(
+                "delete from messages where replayable = @replayable and keep_until <= @p0")
+            .ShouldBe(
+                "delete from messages where replayable = :replayable and keep_until <= :p0");
+    }
+
+    [Fact]
+    public void separates_generic_database_batches_into_oracle_statements()
+    {
+        OracleMessageStore.splitStatements(
+                "select destination from outgoing; delete from incoming;")
+            .ShouldBe(
+            [
+                "select destination from outgoing",
+                "delete from incoming"
+            ]);
+    }
+
+    [Fact]
+    public void converts_boolean_parameters_to_oracle_number()
+    {
+        using var command = new OracleCommand();
+        command.Parameters.Add(new OracleParameter
+        {
+            ParameterName = "replayable",
+            Value = true
+        });
+
+        OracleMessageStore.normalizeParameters(command);
+
+        var parameter = (OracleParameter)command.Parameters["replayable"];
+        parameter.OracleDbType.ShouldBe(OracleDbType.Int16);
+        parameter.Value.ShouldBe(1);
+    }
+}
--- a/src/Persistence/Oracle/Wolverine.Oracle/OracleMessageStore.Durability.cs
+++ b/src/Persistence/Oracle/Wolverine.Oracle/OracleMessageStore.Durability.cs
@@ -1,0 +1,133 @@
+using System.Text.RegularExpressions;
+using Oracle.ManagedDataAccess.Client;
+using Weasel.Core;
+using Wolverine.RDBMS.Polling;
+
+namespace Wolverine.Oracle;
+
+internal partial class OracleMessageStore
+{
+    async Task IDatabaseOperationBatchExecutor.ExecuteDatabaseOperationBatchAsync(
+        IDatabaseOperation[] operations,
+        CancellationToken cancellationToken)
+    {
+        await using var connection =
+            (OracleConnection)await _dataSource.OpenConnectionAsync(cancellationToken);
+        await using var transaction =
+            (OracleTransaction)await connection.BeginTransactionAsync(cancellationToken);
+
+        try
+        {
+            foreach (var operation in operations)
+            {
+                await executeOperationAsync(
+                    connection,
+                    transaction,
+                    operation,
+                    operations,
+                    cancellationToken);
+            }
+
+            await transaction.CommitAsync(cancellationToken);
+        }
+        catch
+        {
+            try
+            {
+                await transaction.RollbackAsync(cancellationToken);
+            }
+            catch
+            {
+                // Preserve the database operation failure that triggered the rollback.
+            }
+
+            throw;
+        }
+    }
+
+    private static async Task executeOperationAsync(
+        OracleConnection connection,
+        OracleTransaction transaction,
+        IDatabaseOperation operation,
+        IDatabaseOperation[] operations,
+        CancellationToken cancellationToken)
+    {
+        var builder = new DbCommandBuilder(connection);
+        operation.ConfigureCommand(builder);
+
+        await using var command = (OracleCommand)builder.Compile();
+        command.BindByName = true;
+        command.Transaction = transaction;
+
+        normalizeParameters(command);
+
+        var statements = splitStatements(command.CommandText);
+
+        if (statements.Length == 0) return;
+
+        try
+        {
+            if (operation is IDoNotReturnData)
+            {
+                foreach (var statement in statements)
+                {
+                    command.CommandText = normalizeParameterMarkers(statement);
+                    await command.ExecuteNonQueryAsync(cancellationToken);
+                }
+
+                return;
+            }
+
+            if (statements.Length != 1)
+            {
+                throw new InvalidOperationException(
+                    $"Oracle database operation '{operation}' returns data and must contain exactly one SQL statement.");
+            }
+
+            command.CommandText = normalizeParameterMarkers(statements[0]);
+            await using var reader = await command.ExecuteReaderAsync(cancellationToken);
+            var exceptions = new List<Exception>();
+            await operation.ReadResultsAsync(reader, exceptions, cancellationToken);
+            await reader.CloseAsync();
+        }
+        catch (ObjectDisposedException)
+        {
+            throw;
+        }
+        catch (Exception e)
+        {
+            throw new DatabaseBatchCommandException(command, operations, e);
+        }
+    }
+
+    internal static string normalizeParameterMarkers(string sql)
+    {
+        return Regex.Replace(sql, @"@(?=[A-Za-z_])", ":");
+    }
+
+    internal static string[] splitStatements(string sql)
+    {
+        return sql.Split(
+            ';',
+            StringSplitOptions.RemoveEmptyEntries | StringSplitOptions.TrimEntries);
+    }
+
+    internal static void normalizeParameters(OracleCommand command)
+    {
+        foreach (OracleParameter parameter in command.Parameters)
+        {
+            switch (parameter.Value)
+            {
+                case bool boolean:
+                    parameter.OracleDbType = OracleDbType.Int16;
+                    parameter.Value = boolean ? 1 : 0;
+                    break;
+
+                case Guid guid:
+                    parameter.OracleDbType = OracleDbType.Raw;
+                    parameter.Value = guid.ToByteArray();
+                    break;
+            }
+        }
+    }
+}
--- a/src/Persistence/Oracle/Wolverine.Oracle/OracleMessageStore.cs
+++ b/src/Persistence/Oracle/Wolverine.Oracle/OracleMessageStore.cs
@@ -36,7 +36,7 @@ using Table = Weasel.Oracle.Tables.Table;
 
 namespace Wolverine.Oracle;
 
-internal partial class OracleMessageStore : IMessageDatabase, IMessageInbox, IMessageOutbox, IMessageStoreAdmin, IDeadLetters, IScheduledMessages, ISagaSupport
+internal partial class OracleMessageStore : IMessageDatabase, IMessageInbox, IMessageOutbox, IMessageStoreAdmin, IDeadLetters, IScheduledMessages, ISagaSupport, IDatabaseOperationBatchExecutor
 {
     private readonly OracleDataSource _dataSource;
     private readonly DatabaseSettings _settings;
--- a/src/Persistence/Wolverine.RDBMS/Polling/DatabaseOperationBatch.cs
+++ b/src/Persistence/Wolverine.RDBMS/Polling/DatabaseOperationBatch.cs
@@ -47,6 +47,21 @@ internal class DatabaseOperationBatch : IAgentCommand
     {
         if (_operations.Length == 0) return AgentCommands.Empty;
 
+        if (_database is IDatabaseOperationBatchExecutor executor)
+        {
+            try
+            {
+                await executor.ExecuteDatabaseOperationBatchAsync(_operations, cancellationToken);
+            }
+            catch (ObjectDisposedException)
+            {
+                // The system is shutting down, let this go.
+                return AgentCommands.Empty;
+            }
+
+            return postProcessingCommands();
+        }
+
         var builder = _database.ToCommandBuilder();
         foreach (var operation in _operations) operation.ConfigureCommand(builder);
 
@@ -79,12 +94,7 @@ internal class DatabaseOperationBatch : IAgentCommand
 
         try
         {
-            var commands = new AgentCommands();
-            foreach (var operation in _operations)
-            {
-                commands.AddRange(operation.PostProcessingCommands());
-            }
-            return commands;
+            return postProcessingCommands();
         }
         finally
         {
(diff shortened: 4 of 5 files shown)


jeremydmiller:
@pedroandrade03 This isn't going to make the release today, but will by EOD Monday

pedroandrade03:
Thanks

jeremydmiller:
## Review notes

Read this one closely since it touches the shared RDBMS durability path. The root cause is right — `DatabaseOperationBatch` concatenates every operation's SQL into one command with `@` markers and reads multiple result sets back, and ODP.NET supports neither — and the shape of the fix (per-operation command, one Oracle transaction, opt-in hook so no other provider changes behavior) is the right instinct. The `internal` interface also sidesteps the default-interface-member trap we've hit before with mocks, and it compiles cross-assembly because `Wolverine.RDBMS/AssemblyAttributes.cs:12` already has `InternalsVisibleTo("Wolverine.Oracle")`.

A few things to work through.

### 1. Test coverage is the main gap

The reported failure in #3614 was `ORA-00933`/`ORA-00936` at runtime, and the three new tests only exercise the three string helpers in isolation. Nothing in the suite would catch a regression in the actual batch execution path — the thing that was broken.

The good news is this is cheap: `CIOracle` already stands up `gvenzl/oracle-free` (`build/CITargets.cs:321-331`) and runs the whole `OracleTests` project against it in CI. A single integration test that persists an incoming/outgoing envelope, drives the real operation set through `DatabaseOperationBatch.ExecuteAsync` (or `DatabaseBatcher`), and asserts recovery would have caught the original bug and would guard the statement-splitting and marker-rewriting logic going forward. That's what I'd most like to see added before merge.

### 2. Parity divergences from the generic path

Worth deciding on each deliberately rather than by omission:

- **`IExceptionTransform`** — `ApplyCallbacksAsync` gives operations a chance to transform driver exceptions (`DatabaseOperationBatch.cs:118,143`). The Oracle path calls `operation.ReadResultsAsync` directly and skips that entirely. No `IDatabaseOperation` implements `IExceptionTransform` today, so nothing is lost right now, but it's a silent divergence that will quietly not work when someone adds one.
- **`ObjectDisposedException`** — the generic path swallows it and still returns `postProcessingCommands()`; the new branch returns `AgentCommands.Empty`. Arguably the new behavior is more correct on shutdown, but the two paths should agree.
- The dropped `exceptions` list in `executeOperationAsync` mirrors the generic path, which also collects and discards it. Parity preserved — no action, just noting I checked.

### 3. The string-level SQL surgery is the fragile part

`splitStatements` splits on **every** `;` and `normalizeParameterMarkers` rewrites **every** `@` followed by a letter or underscore — including occurrences inside string literals. I walked all of `Wolverine.RDBMS/Durability/*` plus `Transport/PollDatabaseControlQueue` and `Transport/DeleteExpiredMessages`, and nothing currently trips either one (the only inline literal is `'{EnvelopeStatus.Incoming}'` in `MoveReplayableErrorMessagesToIncomingOperation`). But nothing *prevents* it, and the failure mode would be a mangled statement at runtime, Oracle-only.

Minimum: state the assumption in a comment on both helpers. Better long-term option worth considering — the operations hardcode the marker themselves (`MoveReplayableErrorMessagesToIncomingOperation.cs:28,32` writes `@replayable` into the SQL text), so having the shared operations emit a provider-supplied prefix would remove the need for any rewrite. That's a bigger change than this PR should carry, but it's the version that doesn't need a regex.

Related: the "returns data and must contain exactly one SQL statement" `InvalidOperationException` is currently unreachable — the only data-returning operations (`CheckRecoverableIncoming/OutgoingMessagesOperation`, `DeleteExpiredDeadLetterMessagesOperation`, `PollDatabaseControlQueue`) are all single-statement. Keeping the guard is fine, but it's a runtime-only trap for whoever writes the next multi-statement data-returning operation; a comment naming the constraint next to the throw would help them.

### 4. Naming and placement

- Per CLAUDE.md, casing follows accessibility: **internal and public members are PascalCase**, private/protected are camelCase. So `normalizeParameterMarkers`, `splitStatements`, and `normalizeParameters` are `internal static` and should be `NormalizeParameterMarkers` / `SplitStatements` / `NormalizeParameters`. `executeOperationAsync` (private static) is correct as-is.
- These three are pure SQL/parameter translation with no message-store state. `Wolverine.Oracle/Util/` (alongside `OracleCommandExtensions`) is a more natural home than `internal static` members hanging off `OracleMessageStore`, and it would let the transport/listener code reuse them.
- `normalizeParameters`'s Guid → `OracleDbType.Raw` + `ToByteArray()` conversion already exists in two other places (`Util/OracleCommandExtensions.cs:27,53` and `Sagas/OracleSagaSchema.cs:180`). Worth consolidating rather than adding a third copy.

### 5. One question

`normalizeParameters` handles `bool` and `Guid`. `DeleteExpiredDeadLetterMessagesOperation` and `DeleteExpiredEnvelopesOperation` pass a `DateTimeOffset` through `builder.AppendParameter`, which then relies on ODP.NET's implicit mapping — while the rest of the store is explicit about `OracleDbType.TimeStampTZ` (e.g. `OracleMessageStore.Incoming.cs:30`). Did the downstream recovery scenario you ran exercise an expiry/`keep_until` comparison? If not, worth confirming that mapping before trusting it, since a silent timezone/precision mismatch there would show up as expired rows never being cleaned rather than as an error.

### CI

Green across the full matrix on this branch, `CIOracle` included. (Minor: the description says it's opened as a draft, but the PR isn't marked draft.)

jeremydmiller:
Follow-up to my review notes above, consolidating what needs to happen here so this doesn't have to be reconstructed later.

**Where this stands.** The diagnosis is correct and the shape of the fix is right — per-operation command, one Oracle transaction, opt-in hook so no other provider changes behavior. CI is green including `CIOracle`. What's holding it is not the design, it's that the change puts string-level SQL rewriting on the shared durability path with no test exercising that path.

**The one thing I'd most like before merge:** an integration test in `OracleTests` that drives the real durability operations through `DatabaseOperationBatch.ExecuteAsync` (or `DatabaseBatcher`) against Oracle and asserts recovery. `CIOracle` already stands up `gvenzl/oracle-free` and runs the whole project against it (`build/CITargets.cs:321-331`), so this costs a test file, not infrastructure. The three tests here cover `NormalizeParameterMarkers` / `SplitStatements` / `NormalizeParameters` in isolation; none of them would catch a regression in the thing that was actually broken in #3614.

**The rest, in priority order:**

1. `splitStatements` splits on every `;` and `normalizeParameterMarkers` rewrites every `@` followed by a letter or underscore — including inside string literals. I walked all of `Wolverine.RDBMS/Durability/*` plus `Transport/PollDatabaseControlQueue` and `Transport/DeleteExpiredMessages` and nothing currently trips either, but nothing prevents it either, and the failure would be a mangled statement at runtime on Oracle only. At minimum state the assumption in a comment on both helpers.
2. Parity gaps with the generic path: `IExceptionTransform` handling (`DatabaseOperationBatch.cs:118,143`) isn't replicated — no live loss today since no `IDatabaseOperation` implements it, but it will quietly not work when one does. And `ObjectDisposedException` returns `AgentCommands.Empty` here where the generic path still returns `postProcessingCommands()`; pick one deliberately.
3. Casing: per CLAUDE.md, casing follows accessibility — internal and public members are PascalCase, private and protected are camelCase. `normalizeParameterMarkers`, `splitStatements`, and `normalizeParameters` are `internal static` and should be `NormalizeParameterMarkers` / `SplitStatements` / `NormalizeParameters`. `executeOperationAsync` is private static and is correct as written.
4. Placement: these three are pure SQL/parameter translation with no message-store state. `Wolverine.Oracle/Util/` (next to `OracleCommandExtensions`) is a better home than internal statics on `OracleMessageStore`, and the Guid → `OracleDbType.Raw` + `ToByteArray()` conversion in `normalizeParameters` already exists twice (`Util/OracleCommandExtensions.cs:27,53` and `Sagas/OracleSagaSchema.cs:180`) — worth consolidating rather than adding a third copy.
5. The "returns data and must contain exactly one SQL statement" throw is currently unreachable (the only data-returning operations are all single-statement). Fine to keep, but a comment naming the constraint would help whoever writes the next multi-statement data-returning operation, since it's a runtime-only trap on Oracle.
6. Still open from before: `DeleteExpiredDeadLetterMessagesOperation` and `DeleteExpiredEnvelopesOperation` pass a `DateTimeOffset` through `builder.AppendParameter` and rely on ODP.NET's implicit mapping, while the rest of the store is explicit about `OracleDbType.TimeStampTZ` (e.g. `OracleMessageStore.Incoming.cs:30`). Did your downstream recovery run exercise an expiry / `keep_until` comparison? A mismatch there would show up as expired rows never being cleaned rather than as an error, so it's worth confirming rather than assuming.

**Worth considering, separately from this PR.** The rewrite exists because the shared operations hardcode the marker into the SQL text (`MoveReplayableErrorMessagesToIncomingOperation.cs:28,32` writes `@replayable`). Having those operations emit a provider-supplied prefix would remove the need for any regex at all. That's a bigger change than this PR should carry, and it's a maintainer call whether it's the direction — but it's the version that doesn't need string surgery.

@pedroandrade03 — thanks for the thorough write-up on the original PR, and for tracking this down past #3589. The `IDatabaseOperationBatchExecutor` hook was a good instinct: keeping it an internal interface rather than a default interface member also sidesteps a mocking trap we've hit before in this codebase.

jeremydmiller:
Thanks for the thorough root-cause work here, @pedroandrade03 — the diagnosis in this PR is what made the fix possible, and your test cases are carried forward in #3659 with you as co-author.

We went a different route on the implementation. Rather than give Oracle a provider-specific `IDatabaseOperationBatchExecutor` and repair the already-compiled SQL, #3659 teaches the *shared* batching mechanics about statement boundaries and lets each provider decide what a boundary means. That's paired with JasperFx/weasel#390 (released as Weasel 9.19.0), which adds `Weasel.Oracle.OracleDbCommandBuilder`.

Two of the three things this PR fixed turned out to already exist in Weasel and just weren't wired up: the `@` markers came from `OracleMessageStore.ToCommandBuilder()` handing back `Weasel.Core.DbCommandBuilder`, whose constructor hardcodes `'@'`, and the Guid→RAW conversion was already in `Weasel.Oracle.CommandBuilder`. The reason neither was reachable is that `Weasel.Oracle.CommandBuilder` is a *sibling* of `DbCommandBuilder` rather than a subclass. The third thing — that ODP.NET genuinely cannot execute several statements from one command — is real and unavoidable, and your PR was right about that.

Working through it surfaced two cases the `;`-splitting couldn't have covered:

- Four operations write more than one statement each (both `ReleaseOrphanedMessages*`, `MoveReplayableErrorMessagesToIncoming`, and `PersistNodeRecord`'s insert-per-event), so splitting per *operation* isn't enough.
- `:replayable` is bound by two different statements, and `AddNamedParameter` finds-or-adds, so it exists exactly once — any index-based slicing binds it to one command and the other fails at execution.

Your PR also indirectly turned up a latent Weasel bug: the Guid→RAW conversion was a `new` member rather than an `override`, so every typed `AppendParameter` overload routed straight past it. Fixed in weasel#390 too.

Closing this in favour of #3659. Really appreciate the contribution.

Issue #2897, opened 2026-05-25
NullReferenceException on Envelope construction when using durable policies in test/dev environment (Wolverine 6 startup-order regression)


gergelyurbancsik:
### Wolverine version
6.x (regression from Wolverine 5)

### Description

When `UseDurableOutboxOnAllSendingEndpoints()`, `UseDurableLocalQueues()`, or `UseDurableInboxOnAllListeners()` are configured, Wolverine sets `MessageRoute.Sender` to a `LocalDurableQueue` instance at route-compilation time. 

In test environments (e.g. using Alba), route compilation happens before `LocalDurableQueue` is fully initialized. As a result, `MessageRoute.Sender` is `null` at the time `Envelope` is constructed, causing a `NullReferenceException` on `agent.Endpoint.DefaultSerializer`.

This did not happen in Wolverine 5 — the startup order changed in 6.0.1.

### Stack trace

```
System.NullReferenceException: Object reference not set to an instance of an object.
   at Wolverine.Runtime.Routing.MessageRoute.CreateForSending(...)
   at Wolverine.Envelope..ctor(...)
```

### Reproduction

```csharp
builder.UseWolverine(opts =>
{
    opts.Policies.UseDurableLocalQueues();
    opts.Policies.UseDurableOutboxOnAllSendingEndpoints();
    opts.Policies.UseDurableInboxOnAllListeners();
});
```

Run any integration test using Alba's `AlbaHost`. The host fails to start with NRE.

### Workaround

Set the environment to `"Development"` in the test host, and guard the durable policies behind an `!IsDevelopment` check:

```csharp
// In test setup:
builder.UseEnvironment("Development");

// In Wolverine configuration:
if (!options.IsDevelopment)
{
    opts.Policies.UseDurableLocalQueues();
    opts.Policies.UseDurableOutboxOnAllSendingEndpoints();
    opts.Policies.UseDurableInboxOnAllListeners();
}
```

### Expected behavior

Durable policies should either initialize safely before route compilation, or gracefully handle the case where `LocalDurableQueue` is not yet ready.

jeremydmiller:
And I can't reproduce this w/ the latest in the TC code tonight

PR #3617, opened 2026-07-24
Resource discovery no longer eager-connects broker transports

### LINKED ISSUE #3616: `resources setup`: broker transports eager-connect during resource *discovery*, defeating the database-creating resources that are deliberately ordered first
This has come from our explorations of provisioning and handling migrations for databases in CI pipelines, that lead to a dive into the source code, and some additional investigations.

### Context

`WolverineSystemPart.FindResources()` calls `await transport.InitializeAsync(_runtime)` for every transport that can build a stateful resource, this happens during resource **discovery**, before any resource's `Setup()` runs. 

`BrokerTransport.InitializeAsync` is sealed and ends in `startupAsync` → `ConnectAsync`, wrapped in a 20×5-second retry loop.

For the database-backed transports (`PostgresqlTransport`, `SqlServerTransport`, Sqlite, Oracle, MySql), `ConnectAsync` ends with a connectivity probe (`Storage.Admin.CheckConnectivityAsync()` — "de facto a little environment test"). 

When the target database is meant to be created by the resource-setup pass itself, discovery burns ~100 seconds retrying against a database that cannot yet exist and then throws out of `FindResources` — which, unlike per-resource `Setup()` failures, is not caught and aggregated by `ResourceExecutor`, so the whole command dies before any resource runs:

```
Wolverine.Transports.BrokerInitializationException: Unable to initialize the Broker postgresql in time
  ← Npgsql.PostgresException 3D000: database "…" does not exist
  at BrokerTransport.InitializeAsync → WolverineSystemPart.FindResources → ResourceExecutor.FindResources
```

The ordering machinery for exactly this scenario already exists and is defeated by the eager connect:

- `FindResources` adds `IResourceCreator` services **first**, with the comment "These have to run first. Right now, the only options are for building multi-tenanted databases with EF Core".
- `ResourceExecutor.ExecuteSetupAsync` runs every `IResourceCreator.EnsureCreatedAsync` before any `Setup()`.

So Wolverine's own EF Core multi-tenancy (`TenantedDbContextInitializer.EnsureCreatedAsync`) combined with a Postgres or SqlServer queue transport pointing at one of the to-be-created databases hits the same wall: discovery throws before the initializer ever runs.

### Why the discovery-time connect appears to be unnecessary

Two observations from the V6.22.0 source:

1. `BrokerResource` re-calls `_transport.ConnectAsync(_runtime)` at the start of **every** operation (`Check`/`Setup`/`Teardown`/`ClearState`/`DetermineStatus`). For the DB transports that call also re-binds the runtime state (`Store`/`Databases`) the endpoints need. The connection made during discovery is redundant for the resource's own function.
2. The only parts of `InitializeAsync` the resource needs at discovery time are connection-free in every transport: compiling explicitly-configured endpoints and `tryBuildSystemEndpoints` (DLQ topology, response/retry queues). The per-endpoint `InitializeAsync` provisioning inside `startupAsync` only acts under `AutoProvision` and delegates to the same `SetupAsync` that `BrokerResource.Setup` already calls unconditionally.

### Suggestion

Split the endpoint-shaping half out of `BrokerTransport.InitializeAsync` (e.g. an `InitializeEndpointsAsync` default interface member on `ITransport` that falls back to `InitializeAsync`) and have `WolverineSystemPart.FindResources` call that instead. Host startup (`WolverineRuntime.HostService.startMessagingTransportsAsync`) keeps the full `InitializeAsync` behaviour, unchanged.

Effects, per transport family:

- **DB transports (Postgres, SqlServer, Sqlite, Oracle, MySql + interop variants):** `resources setup` can provision onto a database an `IResourceCreator` creates in the same pass. The connectivity probe still runs — inside each resource operation, where it belongs, after the creators have run.
- **Broker transports that hard-fail discovery today (RabbitMQ, NATS, Redis, GCP Pub/Sub):** `resources list`/`describe`/`setup` discovery no longer requires live broker connectivity; a down broker fails in the resource operation with a per-resource aggregated error instead of killing discovery after a 100-second retry stall.
- **Soft-connect transports (Azure Service Bus, SQS, SNS, Kafka):** no observable change; their clients are lazy.
- **Unaffected:** MQTT (both), Pulsar, SignalR, and the `DatabaseControlTransport` family — they don't build stateful resources, so they never enter this loop. Runtime-only side effects currently living in `ConnectAsync` paths (SQS orphaned-queue cleanup, keep-alive) stay on the host-startup path.

A PR with this change follows, including a test in `PostgresqlTests` that registers a database-creating `IResourceCreator`, points `UsePostgresqlPersistenceAndTransport` at a database that does not exist, and asserts `IHost.SetupResources()` provisions database + queue tables + envelope storage end-to-end (it fails on V6.22.0 with the `BrokerInitializationException` above, in ~100 seconds; passes with the change in ~0.5s).

Context: found while provisioning per-PR preview databases; same issue-first pattern as #3573.

uniquelau:
Small PR that Closes #3616.

## Problem

`WolverineSystemPart.FindResources()` calls `await transport.InitializeAsync(_runtime)` for every transport that can build a stateful resource — during resource **discovery**, before any resource's `Setup()` runs. `BrokerTransport.InitializeAsync` ends in `startupAsync` → `ConnectAsync` inside a 20×5s retry loop, and for the database-backed transports `ConnectAsync` ends with a connectivity probe.

When the target database is meant to be created by the resource-setup pass itself — the `IResourceCreator` services that `FindResources` deliberately adds *first* ("These have to run first… building multi-tenanted databases") — discovery burns ~100 seconds retrying against a database that cannot yet exist, then throws `BrokerInitializationException` out of `ResourceExecutor.FindResources`, where nothing catches it. Wolverine's own EF Core multi-tenancy (`TenantedDbContextInitializer.EnsureCreatedAsync`) plus a Postgres/SqlServer queue transport pointing at a to-be-created database hits the same wall.

## Change

Discovery now shapes endpoint topology without connecting:

- `ITransport` gains a default interface member `InitializeEndpointsAsync(runtime) => InitializeAsync(runtime)` (non-breaking; same pattern as the existing `Describe()`/`DescribeEndpoint()` defaults).
- `BrokerTransport` implements it as the connection-free half of `InitializeAsync` — compile `explicitEndpoints()` + `tryBuildSystemEndpoints` — and `InitializeAsync` now starts by calling it, so host startup (`WolverineRuntime.HostService.startMessagingTransportsAsync`) is behaviourally unchanged.
- `WolverineSystemPart.FindResources` calls `InitializeEndpointsAsync` instead of `InitializeAsync`.

`BrokerResource` already re-runs `_transport.ConnectAsync` as the first line of every operation (`Check`/`Setup`/`Teardown`/`ClearState`/`DetermineStatus`), which for the DB transports also re-binds `Store`/`Databases` — so each resource connects at operation time, after the `IResourceCreator` pass has created the target.

## Proof test

`PostgresqlTests/Transport/resource_setup_against_a_missing_database.cs`: `UsePostgresqlPersistenceAndTransport` pointed at a database that does not exist, plus a test-local `IResourceCreator` that creates it. Asserts `IHost.SetupResources()` provisions database → transport queue tables → envelope storage, then the host boots.

- On clean `V6.22.0`: fails in 1 m 35 s with `BrokerInitializationException` at `BrokerTransport.InitializeAsync → WolverineSystemPart.FindResources → ResourceExecutor.FindResources`.
- With this change: passes in ~0.5 s.

## Transport audit (why this is safe)

Every `BrokerTransport`-derived transport was audited for resource ops depending on state that only per-endpoint `InitializeAsync(logger)` or `Endpoint.Compile` establishes (the two things discovery no longer runs for non-explicit endpoints):

- **No gaps found.** All resource ops self-resolve their identifiers (e.g. `AmazonSqsQueue.CheckAsync` calls `GetQueueUrlAsync(QueueName)` directly; RabbitMQ ops run inside `WithAdminChannelAsync`; the five DB-transport queues resolve through `Parent.Store` set by the op-level `ConnectAsync`; ASB ops run inside `WithManagementClientAsync`; NATS/Redis/Kafka/GCP ops use transport-level clients from `ConnectAsync`).
- `Endpoint.Compile` guards with `_hasCompiled` — idempotent, and no resource op reads compile-derived state (serializers/Mode are consumed on the sender/listener paths only).
- Only `BrokerTransport` returns `true` from `TryBuildStatefulResource` (sealed); SignalR and the four `DatabaseControlTransport` variants return `false`, MQTT/Pulsar inherit `false` — so the default fallback is never exercised on the resources path and non-broker transports are untouched.
- `describe`/`StartLightweightAsync` stubs external transports and never connected; unchanged.

**Known behavioural nuance (disclosed, not fixed here):** GCP Pub/Sub `CompetingConsumers` listeners get their per-node subscription suffix applied inside `SetupAsync`. Pre-change, a discovery-time `AutoProvision` init applied it before any op; now a *standalone* `resources check`/`statistics` with no prior `setup` in the same process probes the un-suffixed name. Normal `resources setup` and host-startup flows are unaffected.

## Gates (exit-code-gated, net10.0)

- Proof test red on V6.22.0 → green here
- CoreTests 2032/2032; PostgresqlTests full suite 458/458 (1 pre-existing perf skip); SqliteTests 157/157 (1 pre-existing skip)
- Full `wolverine.slnx` Release build: 0 warnings / 0 errors
- Not run locally: SqlServer/RabbitMQ/cloud-broker suites (no containers); net9.0 test leg (runtime not installed on this machine; aborts identically on clean V6.22.0)

### DIFF
--- a/src/Persistence/PostgresqlTests/Transport/resource_setup_against_a_missing_database.cs
+++ b/src/Persistence/PostgresqlTests/Transport/resource_setup_against_a_missing_database.cs
@@ -1,0 +1,139 @@
+using IntegrationTests;
+using JasperFx.Resources;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Hosting;
+using Npgsql;
+using Shouldly;
+using Spectre.Console;
+using Spectre.Console.Rendering;
+using Weasel.Postgresql;
+using Wolverine;
+using Wolverine.Postgresql;
+using Wolverine.RDBMS;
+
+namespace PostgresqlTests.Transport;
+
+public class resource_setup_against_a_missing_database : PostgresqlContext, IAsyncLifetime
+{
+    private const string DatabaseName = "wolverine_fresh_provisioning";
+    private string _connectionString = null!;
+
+    public async Task InitializeAsync()
+    {
+        _connectionString = new NpgsqlConnectionStringBuilder(Servers.PostgresConnectionString)
+        {
+            Database = DatabaseName
+        }.ConnectionString;
+
+        await using var conn = new NpgsqlConnection(Servers.PostgresConnectionString);
+        await conn.OpenAsync();
+        await using var drop = new NpgsqlCommand($"DROP DATABASE IF EXISTS {DatabaseName} WITH (FORCE)", conn);
+        await drop.ExecuteNonQueryAsync();
+        await conn.CloseAsync();
+    }
+
+    public Task DisposeAsync()
+    {
+        return Task.CompletedTask;
+    }
+
+    [Fact]
+    public async Task setup_resources_provisions_the_transport_onto_a_database_created_by_a_resource_creator()
+    {
+        using var host = Host.CreateDefaultBuilder()
+            .UseWolverine(opts =>
+            {
+                opts.Durability.Mode = DurabilityMode.Solo;
+                opts.UsePostgresqlPersistenceAndTransport(_connectionString, schema: "fresh",
+                    transportSchema: "fresh_queues");
+                opts.ListenToPostgresqlQueue("incoming");
+
+                // Stand-in for any database-building registration (EF Core's tenanted DbContext
+                // initializer, or an application-owned creator). IResourceCreator services are
+                // executed before any IStatefulResource.Setup(), so this is the documented way to
+                // make the target database exist before the transport or message store touch it.
+                opts.Services.AddSingleton<IResourceCreator>(new DatabaseCreator(DatabaseName));
+            })
+            .Build();
+
+        // The whole point: resource discovery + setup must succeed while the target database
+        // does not exist yet. Before the discovery seam, FindResources() threw
+        // BrokerInitializationException out of PostgresqlTransport's eager connectivity probe
+        // before DatabaseCreator ever ran.
+        await host.SetupResources();
+
+        (await tableExistsAsync("fresh_queues", "wolverine_queue_incoming")).ShouldBeTrue();
+        (await tableExistsAsync("fresh", DatabaseConstants.IncomingTable)).ShouldBeTrue();
+
+        await host.StartAsync();
+        await host.StopAsync();
+    }
+
+    private async Task<bool> tableExistsAsync(string schema, string table)
+    {
+        await using var conn = new NpgsqlConnection(_connectionString);
+        await conn.OpenAsync();
+
+        var tables = await conn.ExistingTablesAsync(schemas: [schema]);
+        await conn.CloseAsync();
+
+        return tables.Any(x => x.Name == table);
+    }
+
+    private class DatabaseCreator : IResourceCreator
+    {
+        private readonly string _databaseName;
+
+        public DatabaseCreator(string databaseName)
+        {
+            _databaseName = databaseName;
+        }
+
+        public async Task EnsureCreatedAsync(CancellationToken cancellationToken)
+        {
+            await using var conn = new NpgsqlConnection(Servers.PostgresConnectionString);
+            await conn.OpenAsync(cancellationToken);
+
+            await using var exists =
+                new NpgsqlCommand("select 1 from pg_database where datname = @name", conn);
+            exists.Parameters.AddWithValue("name", _databaseName);
+            if (await exists.ExecuteScalarAsync(cancellationToken) is null)
+            {
+                await using var create = new NpgsqlCommand($"CREATE DATABASE {_databaseName}", conn);
+                await create.ExecuteNonQueryAsync(cancellationToken);
+            }
+
+            await conn.CloseAsync();
+        }
+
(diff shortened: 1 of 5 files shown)


uniquelau:
CI on my fork caught a gap the Postgres proof test couldn't reach: with conventional routing, `FindResources`' routing pass hits `MessageRoutingConvention.DiscoverSenders`, which forces `GetOrBuildSendingAgent` — previously safe only because discovery had just connected the broker. `49202cb` guards it with the existing `WithinDescription` machinery (description-mode routes already tolerate a null Sender and are never cached, GH-2897): under description the convention yields the endpoint without forcing its agent. The full RabbitMQ conventional-routing suite passes against a live broker with the change; CoreTests and the Postgres/Sqlite suites stay green. The other fork-CI failure (RavenDb) was a runner disk-full flake, unrelated.

PR #3626, opened 2026-07-25
Fix double execution of compiled queries returned as Wolverine.Http resources (GH-3625)

### LINKED ISSUE #3625: Marten compiled query returned as a Wolverine.Http resource is executed twice per request
When a Wolverine.Http endpoint returns a Marten `ICompiledQuery`/`ICompiledListQuery` as its resource (the `UseMartenCompiledQueryResultPolicy()` pattern), the query is executed **twice** per request:

1. `QuerySpecificationPolicy` (Wolverine.Marten) sees the compiled-query variable created by the endpoint method and injects a `FetchSpecificationFrame`, generating a `QueryAsync(...)` call whose result is never used.
2. `CompiledQueryWriterPolicy` (Wolverine.Http.Marten) then streams the same query with `Marten.AspNetCore.QueryableExtensions.WriteArray(...)` / `WriteOne(...)` (or its own `QueryAsync` for primitive results).

Correctness is unaffected — every request just pays for the query twice.

Observed generated code (Wolverine 6.22.0, endpoint `public static ActiveJourneysQuery Handle() => new();` where `ActiveJourneysQuery : ICompiledListQuery<JourneyDescriptor>`):

```csharp
var activeJourneysQuery = ...Handler.Handle();

// materialized AND discarded
var activeJourneysQuery_result = await ((Marten.IQuerySession)documentSession)
    .QueryAsync<JourneyDescriptor, IEnumerable<JourneyDescriptor>>(activeJourneysQuery, httpContext.RequestAborted).ConfigureAwait(false);

// second execution
await Marten.AspNetCore.QueryableExtensions.WriteArray<JourneyDescriptor, IEnumerable<JourneyDescriptor>>(
    ((Marten.IQuerySession)documentSession), activeJourneysQuery, httpContext, "application/json", 200).ConfigureAwait(false);
```

The same double execution reproduces with the endpoints already in `WolverineWebApi` (`GET /invoices/approved`, `GET /invoices/compiled/{id}`, `GET /invoices/compiled/count`).

Root cause: `QuerySpecificationPolicy` (the Load-returned-specification ergonomic) has no carve-out for a specification variable that is the HTTP chain's resource — the thing another policy is about to execute itself.

PR with a fix incoming.

uniquelau:
Closes #3625.

## Problem

When a Wolverine.Http endpoint returns a Marten `ICompiledQuery` / `ICompiledListQuery` as its resource (the `UseMartenCompiledQueryResultPolicy()` pattern), the query is executed **twice** per request. `QuerySpecificationPolicy` sees the compiled-query variable created by the endpoint method and injects a `FetchSpecificationFrame` (the Load-returned-specification ergonomic), even though `CompiledQueryWriterPolicy` is about to execute the same query itself.

Generated code before (for the existing `GET /invoices/approved` sample endpoint):

```csharp
var approvedInvoicedCompiledQuery = WolverineWebApi.Marten.InvoicesEndpoint.GetApproved();

// materialized AND discarded
var approvedInvoicedCompiledQuery_result = await ((Marten.IQuerySession)documentSession).QueryAsync<WolverineWebApi.Marten.Invoice, System.Collections.Generic.IEnumerable<WolverineWebApi.Marten.Invoice>>(approvedInvoicedCompiledQuery, httpContext.RequestAborted).ConfigureAwait(false);

// second execution
await Marten.AspNetCore.QueryableExtensions.WriteArray<WolverineWebApi.Marten.Invoice, System.Collections.Generic.IEnumerable<WolverineWebApi.Marten.Invoice>>(((Marten.IQuerySession)documentSession), approvedInvoicedCompiledQuery, httpContext, "application/json", 200).ConfigureAwait(false);
```

After:

```csharp
var approvedInvoicedCompiledQuery = WolverineWebApi.Marten.InvoicesEndpoint.GetApproved();
await Marten.AspNetCore.QueryableExtensions.WriteArray<WolverineWebApi.Marten.Invoice, System.Collections.Generic.IEnumerable<WolverineWebApi.Marten.Invoice>>(((Marten.IQuerySession)documentSession), approvedInvoicedCompiledQuery, httpContext, "application/json", 200).ConfigureAwait(false);
```

The primitive-result path (`GET /invoices/compiled/count`) had the same double execution — a discarded fetch ahead of `CompiledQueryWriterPolicy`'s own `QueryAsync`; it now runs exactly once.

## Fix

`QuerySpecificationPolicy` now skips a specification variable that some other frame in the method already consumes **directly** as a pre-bound `MethodCall` argument — which is exactly how `CompiledQueryWriterPolicy` binds the chain's resource into its `WriteArray` / `WriteOne` / `QueryAsync` calls. Arguments are only pre-bound at chain-building time; ordinary data dependencies resolve later during `FindVariables`, so a Load-returned spec whose *result* feeds the handler never matches and keeps its `FetchSpecificationFrame`. The check is structural, so `Wolverine.Marten` needs no reference to `Wolverine.Http.Marten`.

## Tests

New codegen-assertion tests in `Wolverine.Http.Tests/Marten/compiled_query_writer.cs` (alongside the existing behavioral tests, which still pass):

- `compiled_list_query_resource_is_streamed_without_a_redundant_fetch` — `GET /invoices/approved` generates `WriteArray` and no `QueryAsync`
- `compiled_single_query_resource_is_streamed_without_a_redundant_fetch` — `GET /invoices/compiled/{id}` generates `WriteOne` and no `QueryAsync`
- `compiled_primitive_query_resource_is_executed_exactly_once` — `GET /invoices/compiled/count` generates exactly one `.QueryAsync<` call site
- `load_returned_compiled_query_dependency_keeps_its_fetch` — guards the legit case: a new `WolverineWebApi` endpoint whose `Load()` returns a compiled query as a data dependency (not the resource) still gets its fetch, and the fetch result feeds the endpoint at runtime

All three redundant-fetch tests fail on `main` and pass with the fix. Also verified green: the full `Wolverine.Http.Tests` suite (906 passed) and `MartenTests` `fetch_specifications_tests` / batching tests (message-handler Load-returned specs, tuple batching, `[FromQuerySpecification]`).

### DIFF
--- a/src/Http/Wolverine.Http.Tests/Marten/compiled_query_writer.cs
+++ b/src/Http/Wolverine.Http.Tests/Marten/compiled_query_writer.cs
@@ -10,6 +10,90 @@ public class compiled_query_writer : IntegrationContext
     {
     }
 
+    // Warm the endpoint through the real request pipeline so the chain's handler is
+    // compiled by the same path production uses, then hand back the generated source.
+    // Calling ICodeFile.InitializeSynchronously here instead would pre-empt the
+    // route's own lazy buildHandler() and leave the shared AppFixture chain unable
+    // to serve requests in later tests.
+    private async Task<string> sourceCodeFor(string url, int expectedStatusCode = 200)
+    {
+        await Host.Scenario(x =>
+        {
+            x.Get.Url(url);
+            x.StatusCodeShouldBe(expectedStatusCode);
+        });
+
+        var chain = HttpChains.ChainFor("GET", url);
+        chain.ShouldNotBeNull();
+
+        var source = chain.SourceCode;
+        source.ShouldNotBeNull("Failed to generate the source code");
+        return source;
+    }
+
+    // GH-3625: QuerySpecificationPolicy saw the compiled query variable created by
+    // the endpoint method and injected a redundant QueryAsync() fetch whose result
+    // was discarded, so every request executed the query twice — once for the
+    // discarded fetch, once for CompiledQueryWriterPolicy's streaming call.
+
+    [Fact]
+    public async Task compiled_list_query_resource_is_streamed_without_a_redundant_fetch()
+    {
+        var source = await sourceCodeFor("/invoices/approved");
+
+        source.ShouldContain("WriteArray");
+        source.ShouldNotContain("QueryAsync",
+            customMessage: $"The chain's resource is streamed by WriteArray(); a QueryAsync() call means the query also ran a second, discarded time. Source code follows:\n{source}");
+    }
+
+    [Fact]
+    public async Task compiled_single_query_resource_is_streamed_without_a_redundant_fetch()
+    {
+        var chain = HttpChains.ChainFor("GET", "/invoices/compiled/{id}");
+        chain.ShouldNotBeNull();
+
+        // 404 is the WriteOne path for an unknown id — still compiles and runs the chain
+        await Host.Scenario(x =>
+        {
+            x.Get.Url($"/invoices/compiled/{Guid.NewGuid()}");
+            x.StatusCodeShouldBe(404);
+        });
+
+        var source = chain.SourceCode;
+        source.ShouldNotBeNull("Failed to generate the source code");
+
+        source.ShouldContain("WriteOne");
+        source.ShouldNotContain("QueryAsync",
+            customMessage: $"The chain's resource is streamed by WriteOne(); a QueryAsync() call means the query also ran a second, discarded time. Source code follows:\n{source}");
+    }
+
+    [Fact]
+    public async Task compiled_primitive_query_resource_is_executed_exactly_once()
+    {
+        var source = await sourceCodeFor("/invoices/compiled/count");
+
+        // Match the call site rather than the bare name, which also appears in the
+        // result variable's name
+        var count = source.Split(".QueryAsync<").Length - 1;
+        count.ShouldBe(1,
+            $"Expected exactly one QueryAsync() call in the generated source, found {count}. Source code follows:\n{source}");
+    }
+
+    [Fact]
+    public async Task load_returned_compiled_query_dependency_keeps_its_fetch()
+    {
+        var source = await sourceCodeFor("/invoices/approved/count");
+
+        // Here the compiled query is a data dependency produced by Load(), not the
+        // chain's resource, so the FetchSpecificationFrame must still execute it.
+        source.ShouldContain("QueryAsync",
+            customMessage: $"The Load-returned compiled query must still be fetched for the endpoint method. Source code follows:\n{source}");
+
+        // And the fetch result actually feeds the endpoint method at runtime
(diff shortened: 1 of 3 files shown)


jeremydmiller:
## Review notes

Traced this one through both policies and the discriminator is sound. `CompiledQueryWriterPolicy` is an `IResourceWriterPolicy`, so it runs at chain-build time and stamps the spec variable directly into `MethodCall.Arguments` (`Wolverine.Http.Marten/CompiledQueryWriterPolicy.cs:70` for `MartenQueryMethodCall`, `:83` for `MartenWriteJsonToStreamMethodCall`), whereas a `Load()`-returned spec that feeds `Handle`/the endpoint method has its *materialized result* resolved later during `FindVariables`. So "is this spec pre-bound as someone else's argument?" really does separate the resource case from the data-dependency case, and it does it without `Wolverine.Marten` taking a dependency on `Wolverine.Http.Marten`. Good call on the structural check.

The snapshot-vs-mutation detail is also right: `IsConsumedDirectly` reads the pre-loop `frames` snapshot, so the fetch frames this policy inserts for a tuple-returning `Load()` can't accidentally suppress each other.

### Things worth noting

**The coupling is real but implicit.** The guard holds only as long as (a) the resource writer's postprocessor frames are already in `method.Frames` when pre-compilation policies run, and (b) the consumer is a `MethodCall`. A custom `IResourceWriterPolicy` that consumes the spec through some other `Frame` type would still get the double execution. That's an acceptable limitation for the bug at hand — and to your credit the three new codegen tests would fail loudly if the ordering ever inverted, so the invariant isn't unguarded. Just flagging it so it's a known constraint rather than a surprise later.

**Sibling policy has the same shape.** `EFCoreQuerySpecificationPolicy.Apply` (`src/Persistence/Wolverine.EntityFrameworkCore/Codegen/EFCoreQuerySpecificationPolicy.cs:26-50`) is a near-identical copy of the Marten policy and inserts its fetch frame unconditionally. There's no EF Core resource-writer policy that consumes a spec today, so there's no live bug — but if one ever lands, it needs this same guard. Either a comment there pointing at GH-3625, or hoisting the check somewhere shared, would keep the two from drifting.

**`load_returned_compiled_query_dependency_keeps_its_fetch` is the test that matters most** — the guard's whole risk is over-suppression, and that test plus the runtime assertion on the response body covers it. Adding `ApprovedInvoiceCountEndpoint` to `WolverineWebApi` as a permanent fixture is the right call for that.

**Naming**: `IsConsumedDirectly` is `private static` and CLAUDE.md wants camelCase there, but the file already has `IsSpecification` in PascalCase, and "match the surrounding file" wins. Leave it.

### CI

Full matrix green on this branch, including `CIMarten` (15m) and the HTTP suite, so the new `WolverineWebApi` endpoint isn't disturbing any of the route/OpenAPI assertions. The before/after generated-source comparison in the description matches what I'd expect from the frame layout.

PR #3629, opened 2026-07-25
fix(http): parse enum array query and form values case-insensitively

outofrange-consulting:
An enum element bound into an array parsed case-sensitively, while the scalar binder and the List<TEnum> binder both passed ignoreCase to Enum.TryParse. A camelCased value -- the default System.Text.Json spelling clients see in responses -- was therefore dropped from the array while binding fine on a scalar parameter of the same enum type.

With WolverineHttpOptions.RejectUnparseableQueryValues enabled the dropped element turns the request into a 400; before that option existed it silently produced an unfiltered result set, which is the more dangerous failure.

Two emission sites were affected, both on the array path:

  - ParsedArrayQueryStringValue for TEnum[] bound from the query string
  - ParsedArrayFormValue for [FromForm] TEnum[]

Unparseable values are still rejected -- only the case comparison changes, so a genuinely bad value keeps its 400 under strict binding.

The form collection tests are skipped because Alba cannot post form collections, so the form binder is covered end to end with a raw client, matching the approach already used in from_form_file_binding.

### DIFF
--- a/src/Http/Wolverine.Http.Tests/strict_query_string_binding.cs
+++ b/src/Http/Wolverine.Http.Tests/strict_query_string_binding.cs
@@ -345,6 +345,22 @@ public class strict_query_string_binding : IClassFixture<StrictQueryBindingFixtu
         response.Pages.ShouldBe([3]);
     }
 
+    // An enum array element used to parse case-sensitively while the scalar binder passed
+    // ignoreCase, so a camelCased value -- the default System.Text.Json spelling clients see in
+    // responses -- fell through to the rejection block and turned a valid request into a 400.
+    [Fact]
+    public async Task differently_cased_enum_array_elements_bind_in_strict_mode()
+    {
+        var result = await _fixture.StrictHost.Scenario(x =>
+        {
+            x.Get.Url("/strict/collections?Colours=red&Colours=GREEN&Colours=bLuE");
+            x.StatusCodeShouldBe(200);
+        });
+
+        var response = await result.ReadAsJsonAsync<StrictCollectionQueryModel>();
+        response.Colours.ShouldBe([StrictColour.Red, StrictColour.Green, StrictColour.Blue]);
+    }
+
     [Fact]
     public async Task omitted_collections_keep_initializers_in_strict_mode()
     {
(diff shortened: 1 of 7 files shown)


jeremydmiller:
## Review notes

Verified the analysis and it holds up. I swept every enum-parsing emission site in `Wolverine.Http`:

| Site | Path | `ignoreCase` |
|---|---|---|
| `IReadHttpFrame.cs:91` | scalar route/query value | already `true` |
| `ReadClaimFrame.cs:62` | claim value | already `true` |
| `QueryStringHandling.cs:95` | `List<TEnum>` / collection from query | already `true` |
| `FormHandling.cs:48` | `List<TEnum>` / collection from form | already `true` |
| `ParsedArrayQueryStringValue.cs:62` | `TEnum[]` from query | **was missing** → fixed here |
| `FormHandling.cs:149` | `[FromForm] TEnum[]` | **was missing** → fixed here |

So these two really were the last case-sensitive holdouts, and after this the array path matches the scalar and collection paths exactly. Nice, surgical diff.

I also confirmed the claim about strict binding rather than taking it on faith: `ParsedArrayQueryStringValue` does emit `StrictQueryBinding.WriteRejectionBlock` when `WolverineHttpOptions.RejectUnparseableQueryValues` is on, so before this change a camelCased enum element genuinely produced a 400 there — and the silent element-drop when the option is off is the nastier of the two failures, as the description says. (`FormHandling`'s array path has no rejection block, since strict binding is query-only. Correct, nothing to do.)

### One ask

The three new tests cover the lenient default only. The strict-mode case is the one that turns a valid request into a 400, and there's an existing fixture for it — `src/Http/Wolverine.Http.Tests/strict_query_string_binding.cs`. A single test there asserting that a camelCased `TEnum[]` element now binds (200) instead of tripping the rejection block would nail down the more visible half of this bug and would fail loudly if someone ever drops the `true` back out.

### CI

`CICircuitBreaking` is red on `CircuitBreakingTests.RabbitMq.durable_and_not_parallel.the_circuit_breaker_should_trip_and_restart`. That's unrelated to query-string binding — the same job fails on #3639 too, and that suite is chronically load-sensitive in CI. Needs a rerun, not a code change.

Everything else is green, and the `WolverineWebApi` endpoint additions (`/querystring/enumarray`, `/form/array/enum`) are consistent with how the neighbouring array cases are covered.

PR #3634, opened 2026-07-25
feat(pubsub): add ListenToPubsubSubscriptionOnNamedBroker

### LINKED ISSUE #3631: GCP PubSub integration Missing ListenToPubsubSubscription overload supporting specifying non-default broker
## Summary

Named-broker support for `Wolverine.Pubsub` (added in #3318, from #3306) added `OnNamedBroker` overloads for topic-based publishing and listening, but not for attaching to an existing subscription. There's no `ListenToPubsubSubscriptionOnNamedBroker`, so `ListenToPubsubSubscription` only works against the default/unnamed transport.

## Current behavior

`ListenToPubsubSubscription` (`src/Transports/GCP/Wolverine.Pubsub/PubsubTransportExtensions.cs`) is hard-wired to the default transport:

```csharp
public static PubsubTopicListenerConfiguration ListenToPubsubSubscription(
    this WolverineOptions endpoints, string subscriptionName, Action<PubsubEndpoint>? configure = null)
{
    var transport = endpoints.PubsubTransport();   // no BrokerName param
    var topic = transport.Topics[subscriptionName];

    topic.IsListener = true;
    topic.IsExistingSubscription = true;

    configure?.Invoke(topic);

    return new PubsubTopicListenerConfiguration(topic);
}
```

`endpoints.PubsubTransport()` resolves via the internal `PubsubTransport(this WolverineOptions endpoints, BrokerName? name = null)` overload, which calls `TransportCollection.GetOrCreate<PubsubTransport>(name)` with `name = null`, always the default/unnamed transport instance. There's no equivalent that takes a `BrokerName` and threads it into `Topics[subscriptionName]` with `IsExistingSubscription = true`.

By comparison, the topic-based methods already have named-broker parity:

- `AddNamedPubsubBroker(BrokerName name, string projectId, ...)`
- `ListenToPubsubTopicOnNamedBroker(BrokerName name, string topicName, ...)`
- `ToPubsubTopicOnNamedBroker(BrokerName name, string topicName, ...)`

There's also no public workaround. `AddNamedPubsubBroker` returns a `PubsubConfiguration`, which extends `BrokerExpression<...>` and exposes the underlying `PubsubTransport` via a `Transport` property, but that property is `protected internal` (`src/Wolverine/Transports/BrokerExpression.cs`), so it isn't reachable from application code. There's no way to get at a named transport's `Topics` cache from outside the assembly to build the equivalent of `ListenToPubsubSubscription` by hand.

Pub/Sub subscriptions can be attached to without any provisioning rights on the topic, that's the whole point of `ListenToPubsubSubscription` existing separately from `ListenToPubsubTopic`. It lets a consumer attach to an out-of-band-provisioned subscription with `AutoProvision()` left off, which matters when the consuming service has no rights to create infrastructure in the publisher's GCP project. This is quite a common pattern in GCP: each service deployed in its own GCP project, using conventional routing for the default broker, and then using a named broker to receive domain events from a pre-created Pub/Sub subscription exposed from another service's GCP project, with limited permissions.

## Proposed fix

Add `ListenToPubsubSubscriptionOnNamedBroker` to `PubsubTransportExtensions.cs`, mirroring `ListenToPubsubSubscription`'s semantics with the `BrokerName` plumbing already used by `ListenToPubsubTopicOnNamedBroker`:

```csharp
public static PubsubTopicListenerConfiguration ListenToPubsubSubscriptionOnNamedBroker(
    this WolverineOptions endpoints,
    BrokerName name,
    string subscriptionName,
    Action<PubsubEndpoint>? configure = null
)
{
    var transport = endpoints.PubsubTransport(name);
    var topic = transport.Topics[subscriptionName];

    topic.IsListener = true;
    topic.IsExistingSubscription = true;

    configure?.Invoke(topic);

    return new PubsubTopicListenerConfiguration(topic);
}
```

Placement: directly after `ListenToPubsubSubscription`, keeping the file's existing `X` / `XOnNamedBroker` pairing.

`PubsubEndpoint`, `PubsubListener`, `SetupAsync` etc. already branch on `IsExistingSubscription` per-endpoint and don't care which transport instance the endpoint came from, so no other production code should need to change.

### Test coverage

A unit test alongside `Wolverine.Pubsub.Tests/NamedBrokerComplianceTests.cs`, asserting the new method wires `IsExistingSubscription` on the named transport rather than the default one:

```csharp
[Fact]
public void named_broker_can_listen_to_an_existing_subscription()
{
    var options = new WolverineOptions();

    options.AddNamedPubsubBroker(new BrokerName("americas"), "some-project");
    options.ListenToPubsubSubscriptionOnNamedBroker(new BrokerName("americas"), "existing-sub");

    var named = options.Transports.OfType<PubsubTransport>().Single(x => x.Protocol == "americas");
    var endpoint = named.Topics["existing-sub"];

    endpoint.IsListener.ShouldBeTrue();
    endpoint.IsExistingSubscription.ShouldBeTrue();
    endpoint.Uri.Scheme.ShouldBe("americas");
}
```

An integration test that pre-creates a subscription against the emulator and attaches via the new method with `AutoProvision()` off would also help, to confirm messages flow without Wolverine attempting to provision anything.

### Docs

The "Multiple / Named Brokers" section of `docs/guide/messaging/transports/gcp-pubsub/index.md` currently only shows the topic-based sample. It could use a short paragraph plus snippet for the existing-subscription variant, e.g.:

```csharp
// Attach to an existing, already-provisioned subscription on the named broker, no
// AutoProvision() needed if this app has no provisioning rights on that project.
opts.ListenToPubsubSubscriptionOnNamedBroker(new BrokerName("americas"), "existing-subscription-name");
```

bittercoder:
## Summary

Closes #3631.

Named-broker support for `Wolverine.Pubsub` (#3318) added `OnNamedBroker` overloads for topic-based publishing and listening, but not for attaching to an existing subscription. `ListenToPubsubSubscription` only worked against the default/unnamed transport, with no public way to reach a named transport's `Topics` cache from application code.

- Adds `ListenToPubsubSubscriptionOnNamedBroker(BrokerName, string, Action<PubsubEndpoint>?)` to `PubsubTransportExtensions`, mirroring `ListenToPubsubSubscription`'s semantics with the same `BrokerName` plumbing used by `ListenToPubsubTopicOnNamedBroker`.
- No other production code changes needed — `PubsubEndpoint`/`PubsubListener`/`SetupAsync` already branch on `IsExistingSubscription` per-endpoint regardless of which transport instance the endpoint came from.

## Test plan

- [x] New unit test `named_broker_can_listen_to_an_existing_subscription` in `NamedBrokerComplianceTests.cs`, asserting the new method wires `IsExistingSubscription`/`IsListener` on the named transport.
- [x] `dotnet test` on `Wolverine.Pubsub.Tests` (net9.0 + net10.0) — all passing.
- [x] Added a short "Multiple / Named Brokers" doc snippet (`sample_listen_to_pubsub_subscription_on_named_broker`) showing the new overload.

### DIFF
--- a/docs/guide/messaging/transports/gcp-pubsub/index.md
+++ b/docs/guide/messaging/transports/gcp-pubsub/index.md
@@ -94,6 +94,24 @@ var host = await Host.CreateDefaultBuilder()
 
 Note that the `Uri` scheme within Wolverine for any endpoint on a *named* GCP Pub/Sub broker is the broker name you supply, not `pubsub`. So in the example above you would see `Uri` values like `americas://americas-project-id/colors`.
 
+Attaching to an existing, already-provisioned subscription (`ListenToPubsubSubscription`) also has a named-broker counterpart, `ListenToPubsubSubscriptionOnNamedBroker`:
+
+<!-- snippet: sample_listen_to_pubsub_subscription_on_named_broker -->
+<a id='snippet-sample_listen_to_pubsub_subscription_on_named_broker'></a>
+```cs
+var host = await Host.CreateDefaultBuilder()
+    .UseWolverine(opts =>
+    {
+        opts.AddNamedPubsubBroker(new BrokerName("americas"), "americas-project-id");
+
+        // Attach to an existing, already-provisioned subscription on the named broker.
+        // No AutoProvision() needed if this app has no provisioning rights on that project.
+        opts.ListenToPubsubSubscriptionOnNamedBroker(new BrokerName("americas"), "existing-subscription-name");
+    }).StartAsync();
+```
+<sup><a href='https://github.com/JasperFx/wolverine/blob/main/src/Transports/GCP/Wolverine.Pubsub.Tests/DocumentationSamples.cs#L91-L102' title='Snippet source file'>snippet source</a> | <a href='#snippet-sample_listen_to_pubsub_subscription_on_named_broker' title='Start of snippet'>anchor</a></sup>
+<!-- endSnippet -->
+
 ## Multi-Tenancy with a Broker Per Tenant
 
 Named brokers (above) are a *static* topology: you pin specific endpoints to a specific broker at configuration time. **Broker-per-tenant** is different — it is *runtime* routing. You declare one shared topic topology, and each tenant is served by its **own dedicated GCP project**. Which project a message goes to (and which project an inbound message came from) is decided at runtime by the message's [tenant id](/guide/handlers/multi-tenancy), typically set through `DeliveryOptions.TenantId`.
--- a/src/Transports/GCP/Wolverine.Pubsub.Tests/DocumentationSamples.cs
+++ b/src/Transports/GCP/Wolverine.Pubsub.Tests/DocumentationSamples.cs
@@ -86,6 +86,22 @@ public class DocumentationSamples
         #endregion
     }
 
+    public async Task listen_to_existing_subscription_on_named_broker()
+    {
+        #region sample_listen_to_pubsub_subscription_on_named_broker
+        var host = await Host.CreateDefaultBuilder()
+            .UseWolverine(opts =>
+            {
+                opts.AddNamedPubsubBroker(new BrokerName("americas"), "americas-project-id");
+
+                // Attach to an existing, already-provisioned subscription on the named broker.
+                // No AutoProvision() needed if this app has no provisioning rights on that project.
+                opts.ListenToPubsubSubscriptionOnNamedBroker(new BrokerName("americas"), "existing-subscription-name");
+            }).StartAsync();
+
+        #endregion
+    }
+
     public async Task broker_per_tenant()
     {
         #region sample_pubsub_broker_per_tenant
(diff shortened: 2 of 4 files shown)


jeremydmiller:
## Review notes

Checked this against the non-named original and it's a faithful mirror. `ListenToPubsubSubscription` (`PubsubTransportExtensions.cs:171-185`) indexes `transport.Topics[subscriptionName]` **without** `MaybeCorrectName` — deliberate, since an existing subscription mustn't get the `IdentifierPrefix` applied — and sets only `IsListener` / `IsExistingSubscription`, not `EndpointName`. The new overload does exactly the same, differing only in resolving the transport via `endpoints.PubsubTransport(name)`.

Also confirmed the claim that no other production change is needed: every `IsExistingSubscription` branch is per-endpoint (`PubsubEndpoint.cs:98,258,319,334`), so provisioning and listener startup behave identically regardless of which transport instance owns the endpoint.

`CIPubsub` is green, and the doc snippet is wired through `DocumentationSamples` the way the rest of that page is.

Two small notes, neither blocking:

- The two overloads are now a near-exact copy of each other; a private `listenToSubscription(PubsubTransport, string, Action<PubsubEndpoint>?)` shared by both would keep them from drifting. Low stakes given how short they are — your call.
- #3635 also appends to the end of `PubsubTransportExtensions.cs` and to the same doc page. Different hunks, so either merge order should be conflict-free, but whichever lands second is worth a rebuild before merging.

PR #3635, opened 2026-07-25
feat(pubsub): named-broker support for sharded/partitioned topics

### LINKED ISSUE #3632: GCP PubSub integration: Sharded/partitioned topics have no named-broker equivalent
## Summary

Named-broker support for `Wolverine.Pubsub` (added in #3318, from #3306) added `OnNamedBroker` overloads for plain topic publishing/listening (`ListenToPubsubTopicOnNamedBroker`, `ToPubsubTopicOnNamedBroker`), but sharded/partitioned topics, `PublishToShardedPubsubTopics` / `UseShardedPubsubTopics`, don't have an equivalent.

The gap isn't just a missing extension method: the internal topology class backing sharded topics hard-codes the default transport, so a named-broker version needs an internal change, not only new public methods.

## Current behavior

`PublishToShardedPubsubTopics` / `UseShardedPubsubTopics` (`src/Transports/GCP/Wolverine.Pubsub/PubsubTransportExtensions.cs`) are backed by `Internal/PartitionedMessageTopologyWithTopics.cs`, which hard-codes the default transport in all three of its overridden hooks:

```csharp
protected override Endpoint buildEndpoint(WolverineOptions options, string name)
{
    var transport = options.PubsubTransport();               // default only
    return transport.Topics[transport.MaybeCorrectName(name)];
}

protected override PubsubTopicListenerConfiguration buildListener(WolverineOptions options, string name)
    => options.ListenToPubsubTopic(name);                    // default only

protected override PubsubTopicSubscriberConfiguration buildSubscriber(IPublishToExpression expression, string name)
    => expression.ToPubsubTopic(name);                       // default only
```

Each of these resolves the transport via the parameterless `options.PubsubTransport()` / non-named `ListenToPubsubTopic` / `ToPubsubTopic`, all of which resolve the default/unnamed `PubsubTransport`. There's no `PublishToShardedPubsubTopicsOnNamedBroker` / `UseShardedPubsubTopicsOnNamedBroker`, and adding thin wrapper methods alone wouldn't be enough, since `PartitionedMessageTopologyWithTopics` does the actual endpoint/listener/subscriber construction internally and would still need to know which broker to target.

## Proposed fix

Thread a `BrokerName?` through `PartitionedMessageTopologyWithTopics`'s constructor and into its three overridden hooks, so it can resolve either the default or a named transport:

```csharp
protected override Endpoint buildEndpoint(WolverineOptions options, string name)
{
    var transport = options.PubsubTransport(brokerName);      // brokerName may be null (default)
    return transport.Topics[transport.MaybeCorrectName(name)];
}

protected override PubsubTopicListenerConfiguration buildListener(WolverineOptions options, string name)
    => brokerName is null
        ? options.ListenToPubsubTopic(name)
        : options.ListenToPubsubTopicOnNamedBroker(brokerName, name);

protected override PubsubTopicSubscriberConfiguration buildSubscriber(IPublishToExpression expression, string name)
    => brokerName is null
        ? expression.ToPubsubTopic(name)
        : expression.ToPubsubTopicOnNamedBroker(brokerName, name);
```

Then add the two public entry points that pass the broker name through:

- `PublishToShardedPubsubTopicsOnNamedBroker(BrokerName name, ...)`
- `UseShardedPubsubTopicsOnNamedBroker(BrokerName name, ...)`

matching the existing `PublishToShardedPubsubTopics` / `UseShardedPubsubTopics` signatures plus a `BrokerName` parameter, in the same style as `ListenToPubsubTopicOnNamedBroker` / `ToPubsubTopicOnNamedBroker`.

bittercoder:
## Summary

Closes #3632.

`PublishToShardedPubsubTopics` / `UseShardedPubsubTopics` had no named-broker equivalent, because `Internal/PartitionedMessageTopologyWithTopics.cs` always resolved the default/unnamed `PubsubTransport` in its `buildEndpoint`/`buildListener`/`buildSubscriber` overrides.

- Threaded an optional `BrokerName?` through the transport-agnostic `PartitionedMessageTopology` base class (`src/Wolverine/Runtime/Partitioning/PartitionedMessageTopology.cs`), set **before** the endpoint-building loop runs in the base constructor. This matters because `buildEndpoint` is invoked from the base constructor, which runs before any derived class's own constructor body (and thus before a field set there would be assigned) — so the broker name has to be known to the base class itself, not just captured in a `PartitionedMessageTopologyWithTopics` field.
- `PartitionedMessageTopologyWithTopics` now resolves `options.PubsubTransport(BrokerName)` / `ListenToPubsubTopicOnNamedBroker` / `ToPubsubTopicOnNamedBroker` when a broker name is present, falling back to the existing default-transport behavior otherwise.
- Added `PublishToShardedPubsubTopicsOnNamedBroker` / `UseShardedPubsubTopicsOnNamedBroker`, mirroring the existing non-named methods plus a `BrokerName` parameter.
- The base class change is additive and optional (defaults to `null`), so RabbitMQ/SQS/Kafka/Pulsar/Redis/Azure Service Bus sharded topologies are unaffected.

## Test plan

- [x] `dotnet build wolverine.slnx -c Release` — full solution builds clean
- [x] New unit tests in `src/Transports/GCP/Wolverine.Pubsub.Tests/ShardedTopicsNamedBrokerTests.cs` covering: default (unnamed) sharded topology still targets the default transport, a named-broker sharded topology targets the named transport's URIs, and both `PublishToShardedPubsubTopicsOnNamedBroker`/`UseShardedPubsubTopicsOnNamedBroker` register topics on the named transport
- [x] Existing `NamedBrokerComplianceTests` still pass
- [x] Terse doc addition cross-referencing named brokers from the existing Global Partitioning section

### DIFF
--- a/docs/guide/messaging/transports/gcp-pubsub/index.md
+++ b/docs/guide/messaging/transports/gcp-pubsub/index.md
@@ -226,6 +226,12 @@ using var host = await Host.CreateDefaultBuilder()
 
 This creates Pub/Sub topics named `orders1` through `orders4` with companion local queues `global-orders1` through `global-orders4`. Messages are routed to the correct shard based on their group id, and Wolverine handles the coordination between nodes automatically.
 
+Both `PublishToShardedPubsubTopics()` and `UseShardedPubsubTopics()` have `...OnNamedBroker()` equivalents to shard across a [named broker](#multiple-named-brokers) instead of the default one:
+
+```csharp
+topology.UseShardedPubsubTopicsOnNamedBroker(new BrokerName("americas"), "orders", 4);
+```
+
 ## URI reference
 
 The `GcpPubsubEndpointUri` helper class builds canonical endpoint URIs:
--- a/src/Transports/GCP/Wolverine.Pubsub.Tests/ShardedTopicsNamedBrokerTests.cs
+++ b/src/Transports/GCP/Wolverine.Pubsub.Tests/ShardedTopicsNamedBrokerTests.cs
@@ -1,0 +1,83 @@
+using Shouldly;
+using Wolverine.Configuration;
+using Wolverine.Pubsub.Internal;
+using Wolverine.Runtime.Partitioning;
+using Xunit;
+
+namespace Wolverine.Pubsub.Tests;
+
+// GH-3632: PublishToShardedPubsubTopics / UseShardedPubsubTopics had no named-broker equivalent, because
+// PartitionedMessageTopologyWithTopics always resolved the default/unnamed PubsubTransport.
+public class ShardedTopicsNamedBrokerTests
+{
+    private static readonly BrokerName TheName = new("americas");
+
+    [Fact]
+    public void sharded_topology_without_a_broker_name_uses_the_default_transport()
+    {
+        var options = new WolverineOptions();
+        options.UsePubsub("wolverine");
+
+        var topology = new PartitionedMessageTopologyWithTopics(options, PartitionSlots.Five, "orders", 4);
+
+        topology.Slots.Count.ShouldBe(4);
+        topology.Slots.ShouldAllBe(x => x.Uri.Scheme == PubsubTransport.ProtocolName);
+    }
+
+    [Fact]
+    public void sharded_topology_with_a_broker_name_targets_the_named_transport()
+    {
+        var options = new WolverineOptions();
+        options.UsePubsub("wolverine");
+        options.AddNamedPubsubBroker(TheName, "wolverine2");
+
+        var topology = new PartitionedMessageTopologyWithTopics(options, PartitionSlots.Five, "orders", 4, TheName);
+
+        topology.Slots.Count.ShouldBe(4);
+        topology.Slots.ShouldAllBe(x => x.Uri.Scheme == TheName.Name);
+        topology.Slots.Select(x => x.Uri).ShouldBe(new[]
+        {
+            new Uri("americas://wolverine2/orders1"),
+            new Uri("americas://wolverine2/orders2"),
+            new Uri("americas://wolverine2/orders3"),
+            new Uri("americas://wolverine2/orders4")
+        });
+    }
+
+    [Fact]
+    public void publish_to_sharded_pubsub_topics_on_named_broker_targets_the_named_transport()
+    {
+        var options = new WolverineOptions();
+        options.UsePubsub("wolverine");
+        options.AddNamedPubsubBroker(TheName, "wolverine2");
+
+        options.MessagePartitioning.ByMessage<OrderPlaced>(x => x.OrderId.ToString());
+        options.MessagePartitioning.PublishToShardedPubsubTopicsOnNamedBroker(TheName, "orders", 4,
+            topology => topology.Message<OrderPlaced>());
+
+        var named = options.Transports.OfType<PubsubTransport>().Single(x => x.Protocol == TheName.Name);
+        named.Topics.Select(x => x.EndpointName).OrderBy(x => x)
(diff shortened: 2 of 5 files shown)


jeremydmiller:
## Review notes

The constructor-ordering argument in the description is exactly right, and it's the non-obvious part of this change, so worth confirming out loud: `PartitionedMessageTopology`'s base constructor calls `buildEndpoint` inside its own loop (`PartitionedMessageTopology.cs:73-84`), which runs before any derived constructor body. A field assigned in `PartitionedMessageTopologyWithTopics`' constructor would still be null at that point, so threading the broker name into the base constructor really is the only way this works. Good diagnosis.

Also worth noting the nice side effect: `Uri = new Uri($"shard://{transportScheme}/{baseName}")` picks up the named broker's scheme, so a default and a named sharded topology sharing a `baseName` get distinct topology URIs rather than colliding. Tests confirm it (`americas://wolverine2/orders1`…).

`CIPubsub` green.

### Notes

**Naming — `protected readonly BrokerName? BrokerName`.** Two problems with this one. Per CLAUDE.md, private/protected *fields* keep the leading-underscore camelCase form (`_brokerName`), so this should be `_brokerName` regardless. And separately, a field whose name is identical to its type shadows the type name in any derived class: `new BrokerName("x")` still resolves (type context), but `BrokerName.Anything` static access inside a subclass would bind to the field and fail to compile. `_brokerName` avoids both.

**Public constructor signature change.** Adding an optional parameter to the existing public constructors on `PartitionedMessageTopology`, `PartitionedMessageTopology<TListener,TSubscriber>`, and `PartitionedMessageTopologyWithTopics` is source-compatible but *binary*-breaking — anything precompiled against the 4-arg form gets a `MissingMethodException`. On the 6.0 line that may well be fine, but adding an overload instead would sidestep the question entirely. Jeremy's call on whether it matters here.

**Follow-up opportunity.** The base-class plumbing is transport-agnostic, and there are eight sharded topologies deriving from it (RabbitMQ, SQS, Azure Service Bus, Kafka, NATS, Pulsar, Redis, Pub/Sub). Each one is now a ~50-line addition away from the same `...OnNamedBroker` support. Not this PR's job — but if the intent is parity across transports (as with the rest of the named-broker work), this is the cheap moment to file it.

**Nit.** `ShardedTopicsNamedBrokerTests.cs` declares `public record OrderPlaced(Guid OrderId);` at namespace scope in `Wolverine.Pubsub.Tests`. It compiles today, but a generically-named top-level record in a shared test namespace is a future collision; something like `ShardedOrderPlaced` would be safer.

**Merge ordering.** #3634 also appends to `PubsubTransportExtensions.cs` and edits the same doc page. Different hunks, so no textual conflict expected either way — just rebuild whichever lands second.

jeremydmiller:
Superseded by #3653 — closing this one, but the work is yours and it's carried over intact as your commit, authorship preserved.

Two things changed on top of it, both from the review above:

1. **Constructor overloads instead of the optional parameter.** Adding an optional `BrokerName?` to the three existing public constructors is source-compatible but binary-breaking — anything precompiled against the four-argument signature would fail with `MissingMethodException`. The broker name now comes in through a separate overload and the original signatures are untouched.

2. **`BrokerName` field renamed to `_brokerName`**, per the private/protected field convention and to remove the type-name shadowing.

Nothing else moved. The base-constructor ordering insight — that `buildEndpoint` runs from the base constructor, ahead of any derived constructor body, so the broker name has to reach the base class itself — was the crux of this and it carried over exactly as you wrote it. All four of your tests still pass.

Thanks for this one, and for #3634 (merged) and the conventional-routing fix (now #3652).

PR #3639, opened 2026-07-26
Fix conventional routing ignoring named brokers

### LINKED ISSUE #3633: Conventional routing ignores named brokers, across every transport
## Summary

`UseConventionalRouting` configuration on every transport (Pub/Sub, SQS, Azure Service Bus, RabbitMQ) is built on the shared `MessageRoutingConvention<TTransport, TListener, TSubscriber, TSelf>` base class in `Wolverine/Transports/MessageRoutingConvention.cs`. That base class resolves the transport to route against with:

```csharp
var transport = runtime.Options.Transports.GetOrCreate<TTransport>();
```

This call takes no `BrokerName`, so it always resolves the default/unnamed transport instance. If conventional routing is configured against a named broker, it ends up routing against the default broker instead, with no error or warning.

## Current behavior

On transports that support named brokers, the broker-scoped configuration object returned by e.g. `AddNamed<X>Broker` exposes instance methods that mutate that named broker's transport directly, so chaining configuration off it correctly scopes to the named broker. `UseConventionalRouting` is called the same way, as an instance method in that chain, but it doesn't mutate the transport directly, it builds a routing convention object and registers it via `Options.RouteWith(...)`. That convention is applied later by the framework, via the base class above, which has no way to know which broker the original call site intended.

So `opts.AddNamed<X>Broker(name, ...).UseConventionalRouting()` doesn't throw and doesn't route to the named broker, it silently discovers listeners/senders against the default transport instead.

The call site looks scoped to the named broker (it's chained off the named-broker configuration object), but the actual routing happens against the default one. That's a difference in behaviour from every other named-broker-aware call on the same object, and it's easy to miss since nothing signals it at configuration time or runtime.

## Proposed fix

Each transport's `UseConventionalRouting` is a method on a `BrokerExpression<TTransport, ...>` subclass (`PubsubConfiguration`, `AmazonSqsTransportConfiguration`, `AzureServiceBusConfiguration`, `RabbitMqTransportExpression`), and `BrokerExpression` already holds the exact transport instance the call is scoped to, default or named, in its `Transport` property. The routing convention it builds just never receives that instance.

I think if we gave `MessageRoutingConvention<TTransport, ...>` a settable transport binding, and used it instead of `GetOrCreate<TTransport>()` when present, that would fix this:

```csharp
// Wolverine/Transports/MessageRoutingConvention.cs
public TTransport? BoundTransport { get; set; }

private TTransport resolveTransport(IWolverineRuntime runtime)
    => BoundTransport ?? runtime.Options.Transports.GetOrCreate<TTransport>();
```

Then swap all three lookup sites (`DiscoverListeners`, `Describe`, `tryRegisterSenderConfiguration`) to call `resolveTransport(runtime)` instead of `runtime.Options.Transports.GetOrCreate<TTransport>()` directly.

Each transport's `UseConventionalRouting` then sets the binding from the `Transport` it already has, for example in `PubsubConfiguration`:

```csharp
public PubsubConfiguration UseConventionalRouting(
    Action<PubsubMessageRoutingConvention>? configure = null)
{
    var routing = new PubsubMessageRoutingConvention
    {
        BoundTransport = Transport
    };

    configure?.Invoke(routing);

    Options.RouteWith(routing);

    return this;
}
```

`Transport` here is whatever transport instance the surrounding call chain already resolved, the default one for `opts.UsePubsubTransport(...).UseConventionalRouting()`, or the named one for `opts.AddNamedPubsubBroker(name, projectId).UseConventionalRouting()`. The same change applies to the equivalent `UseConventionalRouting` overloads on `AmazonSqsTransportConfiguration`, `AzureServiceBusConfiguration`, and `RabbitMqTransportExpression`, since they're all built the same way against `BrokerExpression`.

### Test coverage

A test per transport that configures conventional routing against a named broker and asserts the resulting listeners/senders were created on that named transport, not the default one.

bittercoder:
## Summary

Fixes #3633.

`MessageRoutingConvention<TTransport, TListener, TSubscriber, TSelf>` resolved the transport to route against with `runtime.Options.Transports.GetOrCreate<TTransport>()`, which always returns the default/unnamed transport instance. `UseConventionalRouting()` is called as an instance method chained off a broker-scoped `BrokerExpression` (e.g. from `AddNamedRabbitMqBroker`, `AddNamedAmazonSqsBroker`, `AddNamedAzureServiceBusBroker`), but the routing convention it builds never received that broker binding — so conventional routing configured against a named broker silently discovered listeners/senders against the default broker instead, with no error or warning.

- Added a settable `BoundTransport` property on `MessageRoutingConvention<...>` and a `resolveTransport(runtime)` helper (`BoundTransport ?? runtime.Options.Transports.GetOrCreate<TTransport>()`), used at all three lookup sites (`DiscoverListeners`, `Describe`, `tryRegisterSenderConfiguration`).
- Each transport's `UseConventionalRouting(...)` overloads (RabbitMQ, SQS, Azure Service Bus, Pub/Sub) now set `BoundTransport = Transport` on the convention they build, using the transport instance the surrounding `BrokerExpression` was already scoped to.
- No doc changes — this restores behavior that already looked like it should work.

## Test plan

- [x] New regression test per transport (RabbitMQ, SQS) that configures both a default broker and a named broker, applies `UseConventionalRouting()` to the named broker, and asserts the discovered listener URI uses the named broker's scheme, not the default transport's. Verified each test fails against unmodified code (listener lands on the default broker) and passes with the fix.
- [x] Added the equivalent test for Azure Service Bus (against the tenant/secondary emulator so it's isolated from the default broker's test data) — could not execute it in this environment because the Azure SQL Edge image used by the local Service Bus emulator doesn't run under this sandbox's arm64 Docker, but it's mechanically identical to the verified RabbitMQ/SQS fix and follows the existing `end_to_end_with_named_broker` test pattern for that transport.
- [x] `dotnet build wolverine.slnx -c Release` — the only failures are pre-existing `NU1903` (Microsoft.OpenApi advisory) errors in unrelated Samples projects, unaffected by this change.
- [x] Existing `ConventionalRouting` test suites for RabbitMQ and SQS pass (one RabbitMQ multi-node tracking test and most of the SQS suite have pre-existing flakiness/broker-timeout failures against this sandbox's LocalStack that reproduce identically on unmodified `main`, unrelated to this change).

### DIFF
--- a/src/Transports/AWS/Wolverine.AmazonSqs.Tests/ConventionalRouting/Bug_3633_conventional_routing_respects_named_broker.cs
+++ b/src/Transports/AWS/Wolverine.AmazonSqs.Tests/ConventionalRouting/Bug_3633_conventional_routing_respects_named_broker.cs
@@ -1,0 +1,46 @@
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Hosting;
+using Shouldly;
+using Wolverine.Runtime;
+using Xunit;
+
+namespace Wolverine.AmazonSqs.Tests.ConventionalRouting;
+
+// https://github.com/JasperFx/wolverine/issues/3633
+public class Bug_3633_conventional_routing_respects_named_broker : IDisposable
+{
+    private static readonly BrokerName theBrokerName = new("other");
+    private readonly IHost _host;
+
+    public Bug_3633_conventional_routing_respects_named_broker()
+    {
+        _host = Host.CreateDefaultBuilder()
+            .UseWolverine(opts =>
+            {
+                // A default, unnamed broker is also registered so that conventional
+                // routing has somewhere wrong to go if the named broker is ignored
+                opts.UseAmazonSqsTransportLocally();
+
+                opts.UseAmazonSqsTransportLocallyAsNamedBroker(theBrokerName)
+                    .UseConventionalRouting(x => x.IncludeTypes(t => t == typeof(RoutedMessage)))
+                    .AutoProvision()
+                    .AutoPurgeOnStartup();
+            }).Start();
+    }
+
+    public void Dispose()
+    {
+        _host.Dispose();
+    }
+
+    [Fact]
+    public void discovers_listener_against_the_named_broker_not_the_default()
+    {
+        var runtime = _host.Services.GetRequiredService<IWolverineRuntime>();
+
+        var uris = runtime.Endpoints.ActiveListeners().Select(x => x.Uri).ToArray();
+
+        uris.ShouldContain(new Uri("other://routed"));
+        uris.ShouldNotContain(new Uri("sqs://routed"));
+    }
+}
--- a/src/Transports/AWS/Wolverine.AmazonSqs/Internal/AmazonSqsTransportConfiguration.cs
+++ b/src/Transports/AWS/Wolverine.AmazonSqs/Internal/AmazonSqsTransportConfiguration.cs
@@ -67,7 +67,10 @@ public class AmazonSqsTransportConfiguration : BrokerExpression<AmazonSqsTranspo
     public AmazonSqsTransportConfiguration UseConventionalRouting(
         Action<AmazonSqsMessageRoutingConvention>? configure = null)
     {
-        var routing = new AmazonSqsMessageRoutingConvention();
+        var routing = new AmazonSqsMessageRoutingConvention
+        {
+            BoundTransport = Transport
+        };
         configure?.Invoke(routing);
 
         Options.RouteWith(routing);
@@ -86,7 +89,10 @@ public class AmazonSqsTransportConfiguration : BrokerExpression<AmazonSqsTranspo
     public AmazonSqsTransportConfiguration UseConventionalRouting(NamingSource namingSource,
         Action<AmazonSqsMessageRoutingConvention>? configure = null)
     {
-        var routing = new AmazonSqsMessageRoutingConvention();
+        var routing = new AmazonSqsMessageRoutingConvention
+        {
+            BoundTransport = Transport
+        };
         routing.UseNaming(namingSource);
         configure?.Invoke(routing);
 
--- a/src/Transports/Azure/Wolverine.AzureServiceBus.Tests/ConventionalRouting/Bug_3633_conventional_routing_respects_named_broker.cs
+++ b/src/Transports/Azure/Wolverine.AzureServiceBus.Tests/ConventionalRouting/Bug_3633_conventional_routing_respects_named_broker.cs
@@ -1,0 +1,50 @@
+using IntegrationTests;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Hosting;
+using Shouldly;
+using Wolverine.Runtime;
+using Xunit;
+
+namespace Wolverine.AzureServiceBus.Tests.ConventionalRouting;
+
+// https://github.com/JasperFx/wolverine/issues/3633
+public class Bug_3633_conventional_routing_respects_named_broker : IAsyncLifetime
+{
+    private static readonly BrokerName theBrokerName = new("other");
+    private IHost _host = null!;
+
+    public async Task InitializeAsync()
+    {
+        await AzureServiceBusTesting.DeleteAllEmulatorObjectsAsync(Servers.AzureServiceBusTenantManagementConnectionString);
+
+        _host = await Host.CreateDefaultBuilder()
+            .UseWolverine(opts =>
+            {
+                // A default, unnamed broker is also registered so that conventional
+                // routing has somewhere wrong to go if the named broker is ignored
+                opts.UseAzureServiceBusTesting();
+
+                opts.AddNamedAzureServiceBusBroker(theBrokerName, Servers.AzureServiceBusTenantConnectionString)
+                    .UseConventionalRouting(x => x.IncludeTypes(t => t == typeof(RoutedMessage)))
+                    .AutoProvision()
+                    .AutoPurgeOnStartup();
+            }).StartAsync();
+    }
+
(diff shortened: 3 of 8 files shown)


jeremydmiller:
## Review notes

Read the whole diff against `main` and traced the routing paths. The diagnosis is right and the fix is at the right level: `MessageRoutingConvention<>` had three transport lookups (`DiscoverListeners`, `Describe`, and the `tryRegisterSenderConfiguration` body reached from both `PreregisterSenders` and the lazy `DiscoverSenders`), and all three are now routed through `resolveTransport`. I checked that nothing else in the convention pipeline re-resolves a transport — the abstract `FindOrCreateListenerForIdentifier` / `FindOrCreateSubscriber` / `ApplyListenerRoutingDefaults` hooks all take the transport as a parameter, so the four subclasses inherit the fix for free. Good.

Two things to fix before this goes in.

### 1. The Azure Service Bus topic/subscription convention is still unbound

`AzureServiceBusConfiguration` has four convention-building methods, not two. The PR patches `UseConventionalRouting` (`AzureServiceBusConfiguration.cs:170,190`) but not `UseTopicAndSubscriptionConventionalRouting` (`AzureServiceBusConfiguration.cs:132,152`), which builds an `AzureServiceBusTopicBroadcastingRoutingConvention` — also a `MessageRoutingConvention<AzureServiceBusTransport,…>`. Since `AddNamedAzureServiceBusBroker` returns the same `AzureServiceBusConfiguration`, GH-3633 reproduces verbatim through that path: topic/subscription conventional routing on a named broker still lands on the default broker. Same two-line addition as the other four sites — `Transport` is already in scope there (it's used at `:122`).

Those two are the only remaining sites; I grepped every public `*ConventionalRouting(` entry point across `src/Transports` and RabbitMQ / SQS / Pub/Sub each have exactly the two the PR already covers.

### 2. The new ASB test fails in CI, and it isn't flakiness

`CIAzureServiceBus` on the previous run (job 89767655693) failed three times on `Bug_3633_conventional_routing_respects_named_broker` with `Connection refused (localhost:5301)`, burning ~8 minutes on the first attempt before the flaky-retry harness took two more passes at it.

Cause: `AddNamedAzureServiceBusBroker(theBrokerName, Servers.AzureServiceBusConnectionString)` gives the named transport the AMQP endpoint (`sb://localhost:5673`) but no `ManagementConnectionString`, so the admin client used by `AutoProvision`/`AutoPurgeOnStartup` derives the HTTP port from the AMQP port and lands on 5301 — which in our compose is `asb-emulator-2`'s management port, and `CIAzureServiceBus` only starts `asb-emulator` (`build/CITargets.cs:640`). This is exactly why the existing `end_to_end_with_named_broker.correct_scheme_on_reply_uri` does:

```csharp
var namedTransport = opts.Transports.GetOrCreate<AzureServiceBusTransport>(theName);
namedTransport.ManagementConnectionString = Servers.AzureServiceBusManagementConnectionString;
```

Note also that the PR description says this test targets "the tenant/secondary emulator", but it passes `Servers.AzureServiceBusConnectionString` (the *primary* emulator), so the intended isolation isn't actually there — and pointing it at `Servers.AzureServiceBusTenantConnectionString` instead wouldn't help, because CI doesn't start that emulator either.

Either mirror the existing pattern (set `ManagementConnectionString` explicitly), or just drop the ASB test. The fix lives in shared core code and the RabbitMQ + SQS tests already cover it; a broker-dependent test that costs 8 minutes to fail isn't earning its keep here. If you do keep an ASB test, keep the `UseTopicAndSubscriptionConventionalRouting` variant in mind — that's the case that's actually still broken.

### Smaller notes

- **`BoundTransport` setter**: all four call sites use an object initializer, so `public TTransport? BoundTransport { get; init; }` would work unchanged and would stop anyone from re-pointing a live convention at a different transport after configuration. A public `set` on a public abstract base also makes it possible to bind a transport that was never registered in `WolverineOptions.Transports`, which would silently produce endpoints that never get initialized.
- **Docs**: the PR argues no doc change is needed since this restores expected behavior. Reasonable, but a sentence in the named-broker sections ("conventional routing configured off a named broker applies to that broker") is cheap and answers the question the issue was really asking.
- **Other CI reds are noise**: the latest run's `CIAzureServiceBus` and `CICircuitBreaking` are *cancelled* at ~20m (job timeout), and the `CICircuitBreaking` failure is `RabbitMq.durable_and_not_parallel.the_circuit_breaker_should_trip_and_restart`, which also fails on the unrelated #3629. Not this diff.

jeremydmiller:
Superseded by #3652 — closing this one, but the fix is yours and it's carried over unchanged as your commit, authorship preserved. The diagnosis was the valuable part here: a silent misroute with no error and no warning, and `BoundTransport` on the shared base is the right level to fix it. `BoundTransport` stays a public settable property.

Three things added on top, all from the review above:

1. **`UseTopicAndSubscriptionConventionalRouting()` now sets `BoundTransport` too.** `AzureServiceBusConfiguration` has four convention-building methods, not two — the topic/subscription pair builds an `AzureServiceBusTopicBroadcastingRoutingConvention` off the same base, and `AddNamedAzureServiceBusBroker()` returns that same configuration object, so GH-3633 reproduced verbatim through that path.

2. **The Azure Service Bus regression test is dropped.** It couldn't pass in CI for the reason in the review — the named broker gets no management connection string, the admin client derives port 5301, and `CIAzureServiceBus` only starts `asb-emulator`. RabbitMQ and SQS already cover the shared-core fix.

3. **Docs for the trap that broke that test**, since it's a genuine footgun on its own: named broker registration has no emulator-flavored overload, and nothing fails until something actually calls the management API, so it surfaces as `Connection refused` against a port nobody configured.

Thanks for this one, and for #3634 (merged) and the sharded-topics work (now #3653).

PR #3653, opened 2026-07-26
feat(pubsub): named-broker support for sharded/partitioned topics (#3632)

### LINKED ISSUE #3632: GCP PubSub integration: Sharded/partitioned topics have no named-broker equivalent

jeremydmiller:
Supersedes #3635. Closes #3632.

The feature is @bittercoder's work, carried over unchanged and preserved as his commit. This PR adds two adjustments that came out of review.

## The feature

`PublishToShardedPubsubTopics` / `UseShardedPubsubTopics` had no named-broker equivalent, because `PartitionedMessageTopologyWithTopics` always resolved the default/unnamed `PubsubTransport` in its `buildEndpoint` / `buildListener` / `buildSubscriber` overrides.

The broker name is threaded through the transport-agnostic `PartitionedMessageTopology` base class and assigned *before* the endpoint-building loop runs in the base constructor. That ordering is the non-obvious part and it's the only thing that works here: `buildEndpoint` is invoked from the base constructor, which runs ahead of any derived class's own constructor body, so a field assigned in `PartitionedMessageTopologyWithTopics` would still be null at that point. Nice catch by @bittercoder.

Adds `PublishToShardedPubsubTopicsOnNamedBroker` / `UseShardedPubsubTopicsOnNamedBroker` mirroring the existing non-named methods. Side effect worth noting: `Uri` becomes `shard://{brokerName}/{baseName}`, so a default and a named sharded topology sharing a `baseName` get distinct topology URIs rather than colliding.

## What's different from #3635

**1. Constructor overloads instead of an optional parameter.** #3635 added an optional `BrokerName?` parameter to three existing *public* constructors (`PartitionedMessageTopology`, `PartitionedMessageTopology<TListener,TSubscriber>`, `PartitionedMessageTopologyWithTopics`). That is source-compatible but binary-breaking — anything precompiled against the four-argument signature would fail with `MissingMethodException`. The broker name now arrives through a separate overload and the original signatures are untouched.

**2. `BrokerName` field renamed to `_brokerName`.** The field was named identically to its own type. Beyond the private/protected field convention, the shadowing is a live hazard: in a derived class the field wins simple-name resolution, so a static member access like `BrokerName.Something` wouldn't compile.

## Notes

This is Pub/Sub-only, matching the issue. The base-class plumbing is transport-agnostic and there are eight sharded topologies deriving from it (RabbitMQ, SQS, Azure Service Bus, Kafka, NATS, Pulsar, Redis, Pub/Sub), so the same support is now a small addition away for each of the others if we want parity later.

## Verification

- All 4 tests in `ShardedTopicsNamedBrokerTests` pass after the refactor.
- `dotnet build wolverine.slnx -c Release` — clean, which also confirms the seven other transports' sharded topologies still bind to the unchanged four-argument constructor.

### DIFF
--- a/docs/guide/messaging/transports/gcp-pubsub/index.md
+++ b/docs/guide/messaging/transports/gcp-pubsub/index.md
@@ -244,6 +244,12 @@ using var host = await Host.CreateDefaultBuilder()
 
 This creates Pub/Sub topics named `orders1` through `orders4` with companion local queues `global-orders1` through `global-orders4`. Messages are routed to the correct shard based on their group id, and Wolverine handles the coordination between nodes automatically.
 
+Both `PublishToShardedPubsubTopics()` and `UseShardedPubsubTopics()` have `...OnNamedBroker()` equivalents to shard across a [named broker](#multiple-named-brokers) instead of the default one:
+
+```csharp
+topology.UseShardedPubsubTopicsOnNamedBroker(new BrokerName("americas"), "orders", 4);
+```
+
 ## URI reference
 
 The `GcpPubsubEndpointUri` helper class builds canonical endpoint URIs:
--- a/src/Transports/GCP/Wolverine.Pubsub.Tests/ShardedTopicsNamedBrokerTests.cs
+++ b/src/Transports/GCP/Wolverine.Pubsub.Tests/ShardedTopicsNamedBrokerTests.cs
@@ -1,0 +1,83 @@
+using Shouldly;
+using Wolverine.Configuration;
+using Wolverine.Pubsub.Internal;
+using Wolverine.Runtime.Partitioning;
+using Xunit;
+
+namespace Wolverine.Pubsub.Tests;
+
+// GH-3632: PublishToShardedPubsubTopics / UseShardedPubsubTopics had no named-broker equivalent, because
+// PartitionedMessageTopologyWithTopics always resolved the default/unnamed PubsubTransport.
+public class ShardedTopicsNamedBrokerTests
+{
+    private static readonly BrokerName TheName = new("americas");
+
+    [Fact]
+    public void sharded_topology_without_a_broker_name_uses_the_default_transport()
+    {
+        var options = new WolverineOptions();
+        options.UsePubsub("wolverine");
+
+        var topology = new PartitionedMessageTopologyWithTopics(options, PartitionSlots.Five, "orders", 4);
+
+        topology.Slots.Count.ShouldBe(4);
+        topology.Slots.ShouldAllBe(x => x.Uri.Scheme == PubsubTransport.ProtocolName);
+    }
+
+    [Fact]
+    public void sharded_topology_with_a_broker_name_targets_the_named_transport()
+    {
+        var options = new WolverineOptions();
+        options.UsePubsub("wolverine");
+        options.AddNamedPubsubBroker(TheName, "wolverine2");
+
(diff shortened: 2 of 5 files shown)


bittercoder:
Thanks heaps for refactoring/fixing this up @jeremydmiller really appreciate it :)

Issue #3492, opened 2026-07-18
RabbitMQ performance deep dive: inline-by-default consumption, durable inbox batching, ack/mapper hot paths


jeremydmiller:
> Follow-up to #3490 (Kafka performance deep dive): the same investigation for RabbitMQ. Shared metrics semantics, resolution-chain costs, and back-pressure mechanics are documented in #3490 and not restated here. Plan authored 2026-07-18; reference line numbers are against `main` @ 6.20.0 and will drift.

Goal: measure Wolverine-over-RabbitMQ overhead vs raw RabbitMQ.Client 7.x across endpoint modes,
produce throughput-tuning guidance, and land the confirmed optimizations. Companion/umbrella:
`KAFKA-PERF-DEEP-DIVE-PLAN.md` (wolverine#3490) — the metrics semantics (§0 there:
execution-time vs effective-time definitions, sub-1ms sample drop, middleware-blocking
inclusion), the resolution-chain/mapper base costs, the BatchedSender debounce, and the
back-pressure agent mechanics are shared and not restated in full here.

All file:line cites verified against `main` @ 6.20.0. Client lib: RabbitMQ.Client **7.1.2**
(fully async).

---

## 0. Transport-specific facts that frame everything

- **Default endpoint mode is `Inline`** (`RabbitMqEndpoint.cs:21`, re-asserted
  `RabbitMqQueue.cs:43`) — unlike the core default of BufferedInMemory. Combined with
  `ConsumerDispatchConcurrency` default **1** (`WolverineRabbitMqChannelOptions.cs:28`), an
  out-of-the-box RabbitMQ listener consumes **one message at a time, end-to-end through the
  handler**, per endpoint. `MaxDegreeOfParallelism` is irrelevant for inline; the only scaling
  knobs are `ListenerCount` (N channels+consumers, `ListeningAgent.cs:365-375`) and the
  transport-wide dispatch concurrency.
- **RabbitMQ never hands the receiver `Envelope[]`** — the consumer's only call is the
  single-envelope `ReceivedAsync` (`WorkerQueueMessageConsumer.cs:100`). The batched
  multi-VALUES durable-inbox insert (`DurableReceiver.cs:608-668`) is therefore **unreachable**;
  durable RabbitMQ pays one `StoreIncomingAsync` INSERT per message (`DurableReceiver.cs:494`)
  even though prefetch delivers natural batches. Same structural gap as Kafka (#3490 T4); SQS
  and ASB do *not* have it.
[...]

PR #3672, opened 2026-07-28
perf(rabbitmq): per-endpoint consumer dispatch concurrency (GH-3492)

jeremydmiller:
Wave 2 of the RabbitMQ performance deep dive (GH-3492). Wave 1 was #3512.

**Scope note:** this PR originally also flipped the acknowledgement semantics. CI caught a real
regression there, so that half has been split onto its own branch — see the bottom of this
description. What remains here is RO6 plus a measured negative result.

## RO6 — `ConsumerDispatchConcurrency` is now per endpoint

The RabbitMQ client's consumer dispatch concurrency was only settable transport-wide, through
`ConfigureChannelCreation`. That made the single most effective lever for RabbitMQ's `Inline`
default an all-or-nothing process setting.

```cs
opts.ListenToRabbitQueue("orders").ConsumerDispatchConcurrency(20);
```

Measured on the in-repo rig (`KafkaPerfRig`, rabbit harness), Inline listener, 1Kb payloads, 5ms
simulated handler, 2,000 msg/s offered load, 10s warmup / 30s measured:

| `ConsumerDispatchConcurrency` | Throughput | Consumed in window | Transit p50 | Total p50 | Handler p50 |
|---|---|---|---|---|---|
| 1 (client default) | 163.7/s | 8,602 | — (nothing from the measured window was consumed before the run ended) | — | — |
| 5 | 828/s | 43,456 | 22,871.9 ms | 22,878.4 ms | 6.311 ms |
| 20 | **1,999.1/s** | 79,980 | **1.486 ms** (p95 2.54, p99 3.22) | **7.275 ms** | 5.573 ms |

The throughput multiples (5.1x, 12.2x) understate it. At 1 and 5 the listener never catches up and
the backlog grows without bound — at 5 the delivery p50 is 22.9 **seconds**. At 20 it matches the
full offered load and delivery p50 collapses to 1.5 ms. Handler time is flat across all three cells,
so this is scheduling, not handler drift. Unlike `ListenerCount`, it costs no extra channels or
connections.

## RO2 ack coalescing — built, measured, rejected

The plan proposed coalescing acks into cumulative ones behind a batching window. I built it
(BatchingChannel + contiguous watermark + cumulative ack, 7 unit tests green) and threw it away on
the numbers, `r-max-inline` (uncapped publisher, 0ms handler):

| Build | Runs |
|---|---|
| `origin/main` baseline | 43,899.8 / 41,024.9 / 44,201.3 msg/s |
| **RO2 ack coalescing** | **39,501 / 37,715 / 37,818.9 msg/s** |
| native `RabbitMQ.Client` twin | 50,965.1 msg/s |

**~10% regression, three runs each, no overlap between the distributions.** The premise was wrong:
`basic.ack` is a fire-and-forget AMQP frame, not an RPC, so coalescing saves zero round trips while
the extra channel hops cost real throughput. The native twin acks per message too and still runs at
51k/s, which is the direct evidence that acks were never the inline gap — Wolverine inline sits at
~84% of the native twin, and whatever the remaining 16% is, it is not acknowledgement.

Recorded in `RABBITMQ-PERF-DEEP-DIVE-PLAN.md` §6 so nobody re-litigates it.

## Split out: the ack semantics change

`BasicAckAsync(tag, multiple: true)` also acks every lower delivery tag on the channel, so
out-of-order completion under a concurrent listener acks messages still in the handler pipeline.
That is a genuine at-least-once hole and I originally fixed it here.

CI proved it is not a one-liner. **The cumulative sweep is load-bearing** — it is currently the only
thing settling deliveries that some paths never acknowledge at all. Measured after a full
`Wolverine.RabbitMQ.Tests` run against a freshly deleted queue:

| Build | `quorum1` after the full suite |
|---|---|
| `origin/main` | 0 messages |
| `multiple: false` | **1 message ready (leaked)** |

That leaked message then fails `quorum_queue_compliance.will_requeue_and_increment_attempts` on the
next run against the same broker, which is exactly what CI hit — twice, on fresh containers, 2-of-2
retry attempts — while #3673 and #3674, branched off the same base with zero RabbitMQ changes, kept
CIRabbitMQ green. Locally the same test passes against a clean queue and fails against a dirty one
on **both** branches, which is why my local baseline missed it.

The fix belongs with the leaking settle path, not with keeping the sweep (prime suspect:
`RabbitMqInteropFriendlyCallback.MoveToErrorsAsync` posts a copy to the dead-letter queue and never
acks or nacks the original, unlike `RabbitMqChannelCallback.moveToErrorQueueAsync` which nacks).
Branch: `fix/3492-rabbitmq-ack-semantics`.

## Tests

New unit coverage for the per-endpoint knob (`consumer_dispatch_concurrency_3492`). Full
`Wolverine.RabbitMQ.Tests` results are on the CI run for this branch.

GH-3492 stays open for the remaining items: the ack semantics fix, RO5 (back-pressure churn), O1b
(batched mark-handled), RO4 (publisher-confirm windowing).

### DIFF
--- a/RABBITMQ-PERF-DEEP-DIVE-PLAN.md
+++ b/RABBITMQ-PERF-DEEP-DIVE-PLAN.md
@@ -163,13 +163,35 @@ transfer); the GH-3490 debounce never applied to Rabbit's sender path (default i
 buffered both unaffected — checked both directions); R5 buffered-below-inline max anomaly
 recorded but unfixed (RO5 follow-up).
 
+### Measured 2026-07-27 (wave 2; local rig, same box, `perf/3492-rabbitmq-wave2` vs `origin/main` @073312262)
+
+| Cell | BEFORE | AFTER | Verdict |
+|---|---|---|---|
+| r-thru-inline @2000/s, 5ms handler, default | 163.7/s | — | R1 reconfirmed on this box |
+| ...same cell + `ConsumerDispatchConcurrency(5)` | 163.7/s | **828/s (5.1x)** | **RO6 headline** |
+| ...same cell + `ConsumerDispatchConcurrency(20)` | 163.7/s | **1,999/s (12.2x)** | RO6 — keeps up with the full offered load |
+| r-max-inline (uncapped, 0ms handler) | 41.0k/41.0-44.2k/s | 42.7k/41.3k/s | ack `multiple:false` fix is throughput-neutral |
+| r-max-native (anchor) | 51.0k/s | — | Wolverine inline ≈ 84% of the native twin at max throughput |
+
+**RO2 ack-watermark coalescing: REJECTED on measurement.** A prototype that batched settlements
+behind a 5ms/prefetch-sized `BatchingChannel` and issued one cumulative `basic.ack` per batch
+measured **37.7-39.5k/s vs a 41.0-44.2k/s baseline — a ~10% regression** across three runs each.
+The premise was wrong: `basic.ack` is a fire-and-forget AMQP frame, not an RPC, so coalescing
+saves no round trips while the extra channel hops cost real throughput. The native twin acks per
+message too, which is why acks were never the inline gap. What survives from RO2 is the
+correctness half — `BasicAckAsync(tag, multiple: true)` on every completion also acked every
+lower delivery tag on the channel, so out-of-order completion under a concurrent listener acked
+messages that were still in the handler pipeline. Now `multiple: false`.
+
 Same rules as #3490 §9: rows only from archived rig runs, exact E-cell + config recorded,
 before/after from the same rig version, one-liner phrased for release notes. Log negative
 results below the table.
 
 | Optimization | PR | Scenario (RE-cell) | Metric | Before | After | Release-note one-liner |
 |---|---|---|---|---|---|---|
-| _(RO1 batched inbox, RO2 ack coalescing, RO3 mapper fixes, ... — rows added as measured)_ | | | | | | |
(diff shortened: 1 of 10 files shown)


jeremydmiller:
## Full measured detail

Adding the latency percentiles behind the throughput table in the description — they are the
stronger half of the story. Rig: `src/Testing/KafkaPerfRig`, rabbit harness, single-node docker
broker, 10s warmup / 30s measured / 10s drain per cell.

### RO6 — Inline listener, 5ms handler, 2,000 msg/s offered

| `ConsumerDispatchConcurrency` | Throughput | Consumed in window | Transit p50 | Total p50 | Handler p50 |
|---|---|---|---|---|---|
| 1 (default) | 163.7/s | 8,602 | — (nothing from the measured window was consumed before the run ended) | — | — |
| 5 | 828/s | 43,456 | 22,871.9 ms | 22,878.4 ms | 6.311 ms |
| 20 | 1,999.1/s | 79,980 | **1.486 ms** (p95 2.54, p99 3.22) | **7.275 ms** | 5.573 ms |

At 1 and 5 the listener never catches up and the backlog grows without bound — at 5 the delivery
p50 is 22.9 **seconds**. At 20 it matches the full offered load and delivery p50 is 1.5 ms. Handler
time is flat across all three cells, so this is scheduling, not handler drift.

### Ack change — `r-max-inline`, uncapped publisher, 0ms handler

| Build | Runs |
|---|---|
| `origin/main` | 43,899.8 / 41,024.9 / 44,201.3 msg/s |
| RO2 ack coalescing prototype (rejected) | 39,501 / 37,715 / 37,818.9 msg/s |
| **This branch (`multiple: false`)** | **42,723.8 / 41,295 msg/s** |
| native `RabbitMQ.Client` twin | 50,965.1 msg/s |

The shipped change is inside baseline run-to-run variance. The rejected prototype is not — three
runs, no overlap with baseline.

Issue #3493, opened 2026-07-18
Amazon SQS performance deep dive: per-message deletes, batch-send failure handling, FIFO sequencing


jeremydmiller:
> Follow-up to #3490 (Kafka performance deep dive): the same investigation for Amazon SQS. Shared metrics semantics, resolution-chain costs, and back-pressure mechanics are documented in #3490 and not restated here. Plan authored 2026-07-18; reference line numbers are against `main` @ 6.20.0 and will drift.

Goal: measure Wolverine-over-SQS overhead vs the raw AWS SDK across endpoint modes, fix the
bug-shaped findings uncovered during research, produce throughput-tuning guidance, and land
confirmed optimizations. Companion/umbrella: `KAFKA-PERF-DEEP-DIVE-PLAN.md` (wolverine#3490) —
shared metrics semantics, resolution-chain costs, BatchedSender mechanics, and back-pressure
behavior live there and are not restated.

All file:line cites verified against `main` @ 6.20.0.

---

## 0. Transport-specific facts that frame everything

- **SQS is the good citizen on receive**: one `ReceiveMessageAsync` returns up to
  `MaxNumberOfMessages` = **10** (`AmazonSqsQueue.cs:107`) with long polling ON by default
  (`WaitTimeSeconds` = 5, `:100`), and the listener hands the receiver a real **`Envelope[]`**
  (`SqsListener.cs:112-115`) → durable mode DOES use the batched multi-VALUES inbox insert
  (`DurableReceiver.cs:608-668`). The Kafka/RabbitMQ per-message-insert gap does not exist here.
- **…but per-message on delete**: completion = one `DeleteMessageAsync` HTTP round trip per
  message (`SqsListener.cs:191-194`); `DeleteMessageBatch` is **never used** (zero hits
  repo-wide). 10 delete RTTs per 1 receive RTT, flowing through a sequential RetryBlock. This
  is SQS's analogue of Kafka's per-message flush — the presumptive #1 ceiling.
- **Send batching is real but throttled by defaults**: `SqsSenderProtocol` chunks by 10 and uses
  `SendMessageBatchAsync` (`SqsSenderProtocol.cs:38-46`); `MessageBatchSize` is overridden to 10
  (`AmazonSqsQueue.cs:44`) but `MessageBatchMaxDegreeOfParallelism` default **1**
  (`Endpoint.cs:281`) caps sustained sends at ~10 msgs per SQS RTT per endpoint, behind the
[...]

PR #3673, opened 2026-07-28
perf(sqs): batch message deletions, and chunk outgoing batches on the request size limit (GH-3493)

jeremydmiller:
The Amazon SQS performance deep dive (GH-3493). SO2 (batch-send partial failures) already landed
in #3503; this is SO1 and SO3.

## SO1 — `DeleteMessageBatch` instead of a delete per message

SQS is a good citizen on receive — one `ReceiveMessage` returns up to 10 messages and durable
endpoints persist the whole array with one batched inbox insert — but completion was one
`DeleteMessage` HTTP round trip per message, flowing through a concurrency-1 retry block. Ten
delete round trips for every one receive round trip. `DeleteMessageBatch` was never used anywhere
in the repo.

Completed messages now accumulate for up to 50ms and are deleted with a single
`DeleteMessageBatch` of up to 10 entries:

- **10x fewer round trips** on the receive side, and since SQS bills per API call, **10x lower
  SQS spend** for message completion.
- `DeleteMessageBatch` is not transactional, so any per-entry failure falls back to a retried
  single delete. A delete that never lands only means the message reappears at its visibility
  timeout, which the durable inbox deduplicates — the same failure mode as before.
- The window is a *maximum batch age*, not a quiet period, so a single message never waits longer
  than 50ms. Batches are flushed on `StopAsync` and on dispose, so a paused listener does not
  leave settled messages to reappear.

Configurable per endpoint, with 1 restoring the old behavior exactly:

```cs
opts.ListenToSqsQueue("orders").DeleteMessageBatchSize(10, 50.Milliseconds());
```

## SO3 — outgoing batches respect the request payload limit

`SendMessageBatch` was chunked on the 10-entry limit only. SQS also caps the **entire request**
at 256KB, not just each message, so ten individually legal 30KB messages bounced the whole
request. Chunking now cuts on both limits. An oversized single message still gets its own request
and fails there, exactly as before.

## Tests

- 11 new unit tests covering chunking (count limit, payload limit, oversized singletons, no
  dropped or duplicated messages) and the delete-batch configuration guard rails.
- Full `Wolverine.AmazonSqs.Tests` against LocalStack, net9.0: **246 passed, 6 skipped, 1 failed**.
  The one failure, `end_to_end_with_conventional_routing_with_prefix.send_from_one_node_to_another_all_with_conventional_routing`,
  fails identically on a clean `origin/main` worktree against the same LocalStack container.

## Not done here

The plan's §8 is explicit that LocalStack numbers are not publishable — the latency/throughput
cells still need a real-SQS account run. The ledger rows added to `SQS-PERF-DEEP-DIVE-PLAN.md`
quote API-call counts, which are exact, rather than timings, which are not. SO5 (visibility
renewal), SO6 (high-throughput FIFO helper) and the §4 sequencing decision table remain open on
the issue.

### DIFF
--- a/SQS-PERF-DEEP-DIVE-PLAN.md
+++ b/SQS-PERF-DEEP-DIVE-PLAN.md
@@ -143,7 +143,14 @@ latency. Log negative results below the table.
 
 | Optimization | PR | Scenario (QE-cell) | Metric | Before | After | Release-note one-liner |
 |---|---|---|---|---|---|---|
-| _(SO1 delete batching, SO2 batch-failure handling, SO4 send parallelism, ... — rows added as measured)_ | | | | | | |
+| SO2 batch-send failure handling | #3503 | — | correctness | silent loss | routed to retry | Per-entry SendMessageBatch failures are no longer silently marked successful |
+| SO1 delete batching | this wave | QE2 | API calls / 10 msgs | 10 DeleteMessage | 1 DeleteMessageBatch | SQS listeners delete completed messages in batches: 10x fewer round trips and 10x lower SQS API spend on the receive side |
+| SO3 size-aware chunking | this wave | — | correctness | whole request bounces | chunked under 256KB | Outgoing SQS batches now respect the 256KB per-request limit, not just the 10-message limit |
+
+**Not measured on real SQS.** SO1's ratio is structural (10 deletes become 1 request) and is
+covered by unit tests plus the LocalStack suites, but the latency/throughput numbers this plan
+asks for still need a real-SQS run per §8 — LocalStack RTTs are not publishable. The ledger rows
+above quote API-call counts, which are exact, rather than timings, which are not.
 
 ## 7. Sequencing & exit criteria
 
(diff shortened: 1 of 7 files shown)


jeremydmiller:
## Exact counts (and an explicit note on what was not measured)

**No timing numbers here, on purpose.** GH-3493 §2/§8 rule LocalStack out as a source of publishable
latency or throughput figures, and the real-SQS run has not happened. These are exact request counts.

### API calls to settle received messages

| Messages settled | Before | After |
|---|---|---|
| 10 (one full receive) | 10 `DeleteMessage` | **1** `DeleteMessageBatch` |
| 1,000 | 1,000 | **100** |

`ceil(N / 10)` requests instead of `N` — 10x fewer round trips and, since SQS bills per API call,
10x lower spend on completion. `DeleteMessageBatch` previously had zero call sites repo-wide, so an
SQS listener paid 10 delete round trips for every 1 receive round trip.

### Outgoing batch chunking, ten 60KB messages

| | Before | After |
|---|---|---|
| Requests | 1 request of ~833KB | 5 requests of 2 messages |
| Outcome | rejected — all 10 fail together | all 10 delivered |

The estimator budgets 240KB per request and models base64 inflation at 4/3 plus 1KB per entry, so it
errs toward more, smaller requests rather than risking a bounce.

### Suite

`Wolverine.AmazonSqs.Tests` vs LocalStack, net9.0: 246 passed / 6 skipped / 1 failed in 7m51s. The
one failure reproduces identically on a clean `origin/main` worktree against the same container.

PR #3688, opened 2026-07-28
GH-3683: make conjoined tenant partition bucketing actually work

### LINKED ISSUE #3683: Conjoined tenant partition bucketing does not actually share a partition on either engine
Found while finishing #3496 (PR #3682). Tenant "bucketing" — registering several small tenants
against one partition suffix so they share a physical partition — is documented on the EF Core
multi-tenancy page and exposed as `ConjoinedTenancyOptions.PartitionPerTenant(p => p.AllowPartitionSharing = true)`,
but it does not work on either engine. It had **no test coverage**, which is why this went unnoticed.

Both defects were reproduced against real PostgreSQL and SQL Server with a throwaway compliance
fixture (host registered with `AllowPartitionSharing = true`, two tenants `smalla`/`smallb` sharing
suffix `shared_bucket`).

## 1. PostgreSQL: the second tenant in a bucket gets no partition slot

`ManagedListPartitions.AddPartitionToAllTables` emits one `ListPartition` per tenant:

```csharp
var partition = new ListPartition(pair.Value /* suffix */, pair.Key.FormatSqlValue() /* one value */);
```

which writes `CREATE TABLE IF NOT EXISTS {parent}_{suffix} PARTITION OF {parent} FOR VALUES IN (<one value>)`.

For the second tenant in the bucket the partition table already exists, so `IF NOT EXISTS` makes the
statement a no-op — the partition keeps accepting only the **first** tenant's id. The second tenant's
first write fails with `23514 no partition of relation "..." found for row`.

Reproduces both when the two tenants are registered in one `AddTenantsAsync` batch and across separate
`AddTenantAsync(id, suffix)` calls.

## 2. SQL Server: bucketing silently breaks across separate calls

Within a single `AddTenantsAsync` batch, bucketing works — the members are grouped by suffix and passed
to the explicit-ordinal `AddPartitionsToAllTables` overload with a shared ordinal.

Across calls it silently does not. `ManagedTenantPartitions`' registry persists only `tenant_id -> ordinal`
with no column for the suffix, so a later call has no way to discover the bucket:

```csharp
var existing = members.Where(m => _partitions.Ordinals.ContainsKey(m)) // empty for a brand new tenant
    .Select(m => _partitions.Ordinals[m]).Distinct().ToArray();
var ordinal = existing.Length == 1 ? existing[0] : (max + 1);          // -> a fresh ordinal
```

Verified: after
```csharp
await partitions.AddTenantAsync("smalla", "shared_bucket");
await partitions.AddTenantAsync("smallb", "shared_bucket");
```
`SELECT COUNT(DISTINCT ordinal) FROM ..._tenant_partitions WHERE tenant_id IN ('smalla','smallb')`
returns **2**, expected 1. Each tenant quietly got its own partition, so the SQL Server
15,000-partition ceiling that bucketing exists to dodge is not actually mitigated.

This is exactly the sequential shape the docs show (`docs/guide/durability/efcore/multi-tenancy.md`).

## Why this is not fixable inside Wolverine

* **PostgreSQL** needs Weasel's `ListPartition` to carry every value belonging to a suffix in one
  partition (`FOR VALUES IN ('smalla','smallb')`), or an attach/detach path to widen an existing
  partition. Wolverine only hands Weasel a `tenantId -> suffix` dictionary.
* **SQL Server** needs the ordinal registry to persist a bucket key alongside the ordinal so a later
  call can resolve it. Registering the suffix itself as a pseudo-tenant row was considered and rejected:
  Weasel releases an ordinal only when its *last* registered tenant is dropped, so the pseudo-row would
  pin the partition forever and defeat `TenantDropBehavior.DeleteData`.

Upstream companion issue: JasperFx/weasel#391.

## Acceptance criteria

- Two tenants registered against one suffix share a single physical partition on both engines, whether
  registered together or in separate calls, and both can read and write.
- Dropping one bucket member leaves the others working; the partition is released only with the last member.
- Compliance coverage for bucketing on both engines in `ConjoinedPartitioningCompliance`.
- Until this is fixed, either the docs carry a loud caveat or `AllowPartitionSharing` fails fast.

## Interim mitigation shipped in #3682

PostgreSQL was not enforcing `AllowPartitionSharing` at all (SQL Server always threw), so with default
settings a user could walk into defect 1 unaware. #3682 applies the guard on both engines; on PostgreSQL
it spans calls, since the tenant -> suffix map *is* persisted there.

Related: #3496, #3465, #3463.

### LINKED ISSUE #3686: Dropping one tenant from a shared partition bucket destroys its co-tenants' data
Found while implementing #3683 / JasperFx/weasel#391, and reproduced against real PostgreSQL. Filing
separately from #3683 because that issue is a *feature does not work* gap, while this is silent data loss
for tenants the caller never named.

## What happens

With tenant bucketing (several small tenants sharing one partition suffix), dropping **one** member of a
bucket destroys the rows of **every other tenant in that bucket**.

`ManagedListPartitions.DropPartitionFromAllTablesForValue` resolves the value to its suffix and then
delegates to the by-**suffix** drop:

```csharp
var suffix = _partitions[value];
await DropPartitionFromAllTables(database, logger, [suffix], token);
```

`DropPartitionFromAllTables` then

1. `delete from <registry> where partition_suffix = ANY(:suffix)` — removing the registry rows for *every*
   tenant in the bucket, not just the one requested, and
2. `ALTER TABLE ... DETACH PARTITION` + `drop table if exists <partition> cascade` — dropping the shared
   physical partition, taking the co-tenants' rows with it.

In Wolverine this is reachable through `IConjoinedTenantPartitions<T>.DropTenantAsync(tenantId, deleteData: true)`.

## Reproduction

Two tenants `smalla` / `smallb` registered against suffix `shared_bucket`, one row each. Dropping `smalla`
leaves `smallb` with no partition and no rows. As an xUnit regression test against real PostgreSQL:

```
Weasel.Postgresql.Tests...managed_list_partitions
    .dropping_one_bucket_member_preserves_the_other_members_data
```

which fails on Weasel 9.19.1 and passes with the weasel#391 fix.

## Blast radius

`ManagedListPartitions` is shared, so this is not Wolverine-specific — Marten and Polecat consumers using
managed list partitions with a shared suffix are exposed to the same behavior.

Mitigating factor: bucketing never actually worked end to end (#3683), so a bucket with more than one live
member was hard to arrive at in the first place, and Wolverine 6.23.1+ fails fast on a shared suffix unless
`AllowPartitionSharing` is enabled.

## Fix

Landed with the weasel#391 work: the by-value drop now narrows the bucket instead — deleting only the
departing tenant's rows and re-binding the partition to the remaining members — and releases the physical
partition only when its last member is dropped.

Related: #3683, #3496, JasperFx/weasel#391.

jeremydmiller:
Closes #3683. Also closes #3686 (the co-tenant data loss found alongside it).

> [!IMPORTANT]
> **Draft — blocked on Weasel 9.20.0 reaching nuget.org.** The root cause is in Weasel, fixed by
> JasperFx/weasel#392. CI will fail restore until that version is published. Everything here was
> verified locally against a packed 9.20.0 build.

## The problem

Tenant *bucketing* — registering several small tenants against one partition suffix so they share a physical
partition — is documented on the EF Core multi-tenancy page and exposed through
`PartitionPerTenant(p => p.AllowPartitionSharing = true)`. It did not work on either engine. It had **no test
coverage**, which is why it went unnoticed; the doc sample demonstrating it is compile-only and never executed.

Both defects were reproduced against real PostgreSQL and SQL Server:

* **PostgreSQL** emitted one single-value `ListPartition` per tenant, so the second member of a bucket was
  swallowed by `CREATE TABLE IF NOT EXISTS` and its first write failed with `23514 no partition of relation
  found for row`.
* **SQL Server** worked inside a single batch but silently not across calls — the natural tenant-onboarding
  shape, and the one the docs showed. The registry stored only `tenant_id -> ordinal`, so a brand-new tenant
  could not discover which ordinal its bucket already owned and quietly got its own partition. The
  15,000-partition ceiling that bucketing exists to dodge was not mitigated at all.

Neither was fixable inside Wolverine, hence JasperFx/weasel#392.

### Plus a data-loss defect found along the way (#3686)

Dropping **one** member of a bucket destroyed **every co-tenant's rows**: the by-value drop resolved the tenant
to its suffix and then dropped by *suffix*, deleting the whole bucket's registry rows and `DROP`ping the shared
partition table. Also fixed in weasel#392, and covered here.

## Changes

`PostgresqlTenantPartitioning` needs **no change** — it hands Weasel the tenant → suffix map, and the fixed
Weasel does the right thing.

`SqlServerTenantPartitioning` was resolving the bucket's ordinal itself out of the tenant → ordinal map, which
is blind to buckets: for a brand-new tenant the lookup matched nothing and it allocated a fresh ordinal. It now
delegates to Weasel's bucket-aware overload, and its `AllowPartitionSharing` guard spans calls via the
persisted bucket map — matching what PostgreSQL already did.

**Compliance coverage on both engines** for #3683's acceptance criteria: members registered together and
separately land in one physical partition, and dropping one member leaves the others reading, writing, and
holding their data.

**Docs.** The bucketing sample showed the sequential pattern *without* enabling `AllowPartitionSharing`, so as
written it would have thrown. The config sample now enables it, and a new **Tenant Bucketing** section covers
when to reach for it, the drop semantics, the isolation trade-off, and the 9.20.0 requirement.

## Verification

Real databases throughout.

| Suite | Result |
|---|---|
| `EfCoreTests.MultiTenancy` | 193/193 |
| `SqlServerTests` | 388/390 (2 skipped, pre-existing) |
| `PostgresqlTests` | 466/467 (1 skipped, pre-existing) |
| `wolverine.slnx` Release build | clean, 0 warnings |

Upstream weasel#392 carries 5 new PostgreSQL tests (**all 5 red on 9.19.1**) and 5 new SQL Server tests, with
both full Weasel suites green.

## Merge order

1. JasperFx/weasel#392
2. Publish Weasel 9.20.0
3. Mark this ready — the `Directory.Packages.props` bump to 9.20.0 is already in this branch

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -127,13 +127,13 @@
     <PackageVersion Include="System.Diagnostics.DiagnosticSource" Version="9.0.5" />
     <PackageVersion Include="System.Net.NameResolution" Version="4.3.0" />
     <PackageVersion Include="System.Threading.Tasks.Dataflow" Version="9.0.5" />
-    <PackageVersion Include="Weasel.Core" Version="9.19.0" />
-    <PackageVersion Include="Weasel.EntityFrameworkCore" Version="9.19.0" />
-    <PackageVersion Include="Weasel.MySql" Version="9.19.0" />
-    <PackageVersion Include="Weasel.Oracle" Version="9.19.0" />
-    <PackageVersion Include="Weasel.Postgresql" Version="9.19.0" />
-    <PackageVersion Include="Weasel.SqlServer" Version="9.19.0" />
-    <PackageVersion Include="Weasel.Sqlite" Version="9.19.0" />
+    <PackageVersion Include="Weasel.Core" Version="9.20.0" />
+    <PackageVersion Include="Weasel.EntityFrameworkCore" Version="9.20.0" />
+    <PackageVersion Include="Weasel.MySql" Version="9.20.0" />
+    <PackageVersion Include="Weasel.Oracle" Version="9.20.0" />
+    <PackageVersion Include="Weasel.Postgresql" Version="9.20.0" />
+    <PackageVersion Include="Weasel.SqlServer" Version="9.20.0" />
+    <PackageVersion Include="Weasel.Sqlite" Version="9.20.0" />
     <PackageVersion Include="xunit" Version="2.9.3" />
     <PackageVersion Include="xunit.assemblyfixture" Version="2.2.0" />
     <PackageVersion Include="xunit.runner.visualstudio" Version="2.8.2" />
--- a/docs/guide/durability/efcore/multi-tenancy.md
+++ b/docs/guide/durability/efcore/multi-tenancy.md
@@ -572,7 +572,12 @@ opts.Services.AddDbContextWithWolverineManagedConjoinedTenancy<ConjoinedTenancy.
 
     // Weasel-managed physical partitioning: one partition (or shared
     // bucket) per tenant on every non-saga ITenanted entity table
-    tenancy => tenancy.PartitionPerTenant());
+    tenancy => tenancy.PartitionPerTenant(partitioning =>
+    {
+        // Opt in before registering two tenants against one suffix.
+        // Without this a shared suffix is rejected outright
+        partitioning.AllowPartitionSharing = true;
+    }));
 ```
 <sup><a href='https://github.com/JasperFx/wolverine/blob/main/src/Persistence/EfCoreTests.MultiTenancy/MultiTenancyDocumentationSamples.cs#L257-L265' title='Snippet source file'>snippet source</a> | <a href='#snippet-sample_conjoined_tenancy_with_partitioning' title='Start of snippet'>anchor</a></sup>
 <!-- endSnippet -->
@@ -582,7 +587,7 @@ With partitioning enabled:
 * On PostgreSQL, every non-saga `ITenanted` entity table becomes `PARTITION BY LIST (tenant_id)` with one partition per tenant, managed through a `wolverine_tenant_partitions` control table in the durability schema
 * On SQL Server (which can only range-partition over a compact value), entities gain an `int tenant_ordinal` column stamped automatically by Wolverine, and tables are `RANGE RIGHT` partitioned over the ordinal with a registry table mapping tenant ids to ordinals
 * The composite `(tenant, id)` primary key exists **only in the database** — your EF model keeps its own single key, so `FindAsync()`, `Attach()`, and saga loads keep exactly the same call shapes
-* Multiple small tenants can share one physical partition ("bucketing") by registering them with the same partition suffix — the answer to SQL Server's partition count ceiling and to "small tenants don't deserve their own partition"
+* Multiple small tenants can share one physical partition ("bucketing") by registering them with the same partition suffix — the answer to SQL Server's partition count ceiling and to "small tenants don't deserve their own partition". See [Tenant Bucketing](#tenant-bucketing) below
 * Partitioned conjoined contexts require `UseEntityFrameworkCoreWolverineManagedMigrations()` — EF migrations cannot express the partition DDL
 
 Manage tenants through `IConjoinedTenantPartitions<TDbContext>`:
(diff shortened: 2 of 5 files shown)


jeremydmiller:
Unblocked — Weasel 9.20.0 is published (JasperFx/weasel#392 merged; publish workflow completed 16:36 UTC).

Verified against the **published** package, with no local feed overlay in play:

- `Directory.Packages.props` resolves `Weasel.Core/9.20.0` and `Weasel.SqlServer/9.20.0` straight from nuget.org
- Full `wolverine.slnx` Release build clean — 0 warnings, 0 errors

Marking ready for review.

Issue #3752, opened 2026-07-31
CI wall clock is set by CIMarten and CIAzureServiceBus; compliance classes dominate transport suites


jeremydmiller:
## Where the CI wall clock actually goes

The `tests.yml` matrix runs ~30 jobs concurrently, so the workflow's wall clock is the **slowest single job**, not the total. From run [30634981923](https://github.com/JasperFx/wolverine/actions/runs/30634981923):

| Job | Duration |
|---|---|
| CIMarten | **18m** |
| CIAzureServiceBus | **18m** |
| CISqlServer | 14m |
| CIPubsub | 12m |
| CIPersistence | 12m |
| CIKafka | 12m |
| CIRabbitMQ | 11m |
| CIEfCore | 11m |
| CIRedis | 9m |
| CIRavenDb | 9m |
| CICosmosDb | 6m |
| CIPolecat / CICircuitBreaking / CIAWSSqs | 5m |
| CIPulsar | 3m |

**Two jobs set the number.** Any work on a job below ~14m changes the workflow's wall clock by zero. That is worth stating explicitly because it is easy to optimise the suite you happen to be looking at.

## What a measurement of one suite showed

`Wolverine.Redis.Tests` (144 tests) was measured in detail as a pilot — running it across worker *processes*, partitioned by test class:

| Workers | Wall | Speed-up | Result |
|---|---|---|---|
| 1 | 451.8s | 1.00x | 144/144 |
| 2 | 243.1s | 1.86x | 144/144 |
| 4 | 205.8s | 2.20x | 144/144 |
| 8 | 199.0s | 2.27x | 144/144 |

Green at every size, with retries disabled so nothing could mask instability. But it **plateaus hard at ~199s**, and the reason is the useful part:

| Class | Total | Tests |
|---|---|---|
| `BufferedSendingAndReceivingCompliance` | **188.1s** | 22 |
| `InlineSendingAndReceivingCompliance` | **178.6s** | 22 |
| everything else (29 classes) | 233s | 100 |

**Two compliance classes hold 61% of the suite's total test time** (366.7s of 599.4s). Since a test class cannot be split across processes without breaking its isolation contract, the largest class is a hard floor: `599.4 / 188.1 = 3.19x` is the ceiling, and no amount of parallelism goes below ~188s.

[...]

PR #3758, opened 2026-08-01
Bobcat-supervised CI test runs: parallel workers, per-lane databases, honest retries

jeremydmiller:
Replaces the Nuke build's `dotnet test` + TRX-parse-and-retry harness with test runs driven by [Bobcat's](https://github.com/JasperFx/bobcat) `Supervisor` (`Bobcat.Supervisor` 0.6.1 on nuget.org), and takes the first measured bite out of the CI wall clock: **CIMarten, co-longest of the 28-job matrix (#3752), went 18m → 9.5m — 544 + 12 tests passed with zero retries.**

## Measured results, CI hardware (baseline = run 30634981923)

| Job | Before | After | | Job | Before | After |
|---|---|---|---|---|---|---|
| **CIMarten** | **18m** | **9.5m** | | CIPolecatWorkflow | 4m | 5.3m |
| **CIAzureServiceBus** | **18m** | 16.9m | | CIPolecatSagas | 4m | 4.5m |
| CISqlServer | 14m | 15.9m | | CIAWSSqsCompliance | 4m | 4.5m |
| CIPubsub | 12m | 12.8m | | CIPulsar | 3m | 3.4m |
| CIPersistence | 12m | 10.9m | | CIMQTT5 | 3m | 2.5m |
| CIKafka | 12m | 12.1m | | CIAWSSns | 3m | 3.7m |
| CIRabbitMQ | 11m | 12.5m | | CISqlite | 2m | 2.2m |
| CIEfCore | 11m | 11.9m | | CIOracle | 2m | 2.4m |
| CIRedis | 9m | 9.5m | | CINATS | 2m | 2.6m |
| CIRavenDb | 9m | 7.9m | | CIMySql | 2m | 2.8m |
| CICosmosDb | 6m | 5.9m | | CIMQTT | 2m | 2.7m |
| CIPolecat | 5m | 4.9m | | CIGrpc | 2m | 2m |
| CICircuitBreaking | 5m | 5.2m | | CIOtel | ≤1m | 1.6m |
| CIAWSSqs | 5m | 5.8m | | CIAotSmoke | ≤1m | 0.9m |

Reading it honestly:

- **CIMarten is the only job that was parallelized, and it halved.** Everything else runs at 1 worker; their ±1m deltas are runner variance plus the supervisor's small discovery overhead — parity, as intended.
- **Workflow wall clock: 18m → 16.9m.** The matrix takes as long as its slowest job, and **CIAzureServiceBus is now the sole pole** — still sequential against one emulator, but its per-test retries already stopped re-running the whole suite. Next in line after ASB: CISqlServer (15.9m), then Kafka/Pubsub/RabbitMQ (~12m each, broker-bound, needing the same per-lane isolation story).
- **Final check state: 31 of 32 green.** The one red is CIKafka's `DeadLetterQueueTests.dead_letter_message_has_exception_headers`, which was failing on `main` before this branch (runs 30666416761, 30634981923) and failed identically, sequentially, in every round here — pre-existing, now reported with its name and exception in the job summary.
- **The ASB flaky ledger now has an address:** all 10 of its pass-on-retries in the final round sit in `TopicAndSubscriptionWithCustomRuleSendingAndReceivingCompliance`. That class is where the `Flaky`-tag removal work should start.

### 1. `Servers.cs`: every connection string overridable by environment variable

`WOLVERINE_POSTGRES`, `WOLVERINE_SQLSERVER`, `WOLVERINE_MYSQL`, `WOLVERINE_ORACLE`, and the four Azure Service Bus endpoints, with the existing docker-compose values as fallbacks — a developer or CI job that sets nothing behaves exactly as before. This is what makes per-worker isolation possible: schema names are hard-coded throughout the suites, so two processes sharing one database collide however the tests are partitioned. Isolation has to be the database.

### 2. Every `*Tests` project is a Microsoft.Testing.Platform host

One property in `Directory.Build.props` (`UseMicrosoftTestingPlatformRunner`). It changes only the built executable's entry point — `dotnet test`, VSTest filters, TRX and coverlet all keep working. Set repo-wide rather than per-job so a build can never silently produce a non-supervisable executable.

### 3. The supervised runner (`build/SupervisedTests.cs`)

`RunTestProject` keeps its call shape; the internals now launch the suite executable as supervised worker processes over the MTP wire protocol:

- **Class-level partitioning, never per-test** — splitting a class across processes breaks the isolation contract xUnit writes against static state.
- **`Category!=Flaky` and the AWS/Polecat shard filters became `TestFilter` predicates.** A shard filter that matches nothing now *fails the run* — a renamed namespace must not read as a green job.
- **Retry parity, reported honestly.** Up to three attempts (the old harness's ceiling), each retry in a fresh quiet process, under a real `RetryBudget`. A pass-on-retry is `[FLAKY]` in the log — never folded into a clean pass — and a crashed worker is `[INDETERMINATE]` with its exit code and stderr.
- **Knobs:** per-target `workers` (clamped to `ProcessorCount/2` unless `--test-workers` overrides) and `postgresDatabasePerLane` (provisions `wolverine_w0..N-1` through `WOLVERINE_POSTGRES`), plus `--disable-test-retry` for measurement.

The TRX parsing, per-test re-invocation, and filter-string escaping in `TestAllPersistence.cs` are deleted.

### 4. CIMarten at 4 workers — and the ten tests that exposed

The first parallel runs surfaced two latent bug classes, both fixed here:

- **Nine tests asserted the literal default database name** (durability-agent / event-subscription URIs, raw connection strings). They now build expectations from `Servers.PostgresDatabaseName`.
- **One theory computed its data from the environment**, giving the same test a different identity in the discovery worker than in the executing lane — the symptom is "the worker finished without reporting a result." Theory data must be a function of the code alone; a stable sentinel is resolved inside the test body instead.

### What the CI rounds taught Bobcat (fed upstream, not worked around)

- Idle lanes sit at peak working set through the retry passes, so a fresh-process retry held `workers+1` full test hosts and OOM-killed 16GB runners → `Supervisor.ReleaseIdleLanes` in Bobcat 0.6.1 (JasperFx/bobcat#81).
- A warm-process retry can never pass for a test whose failure pollutes its process (MySql's `end_to_end_from_scratch` cannot run "from scratch" where scratch no longer exists) — tried, reverted, documented.
- The environment-dependent-identity shape is now bug #3 in Bobcat's `docs/parallel-ready-suites.md`; the uid-instead-of-display-name report for never-reporting tests is JasperFx/bobcat#82.

## Deliberately *not* done here

- **Every other target stays at 1 worker.** Raising a suite's `workers` is a per-suite decision after profiling: the largest class is a hard floor (`sum(durations) / largest class`), and broker-backed suites need per-lane resource isolation first.
- **Pulling back is cheap by design.** For any well-behaved suite where the supervisor buys nothing, the default is already sequential (`workers: 1`); a target can also drop back to a bare `DotNetTest` call without touching any other target, since the MTP host property changes nothing about `dotnet test`.
- **`Flaky` tags stay on** — the quarantine ledger is what will make removing them evidence-based, starting with the ASB compliance class named above.
- Marten's sibling databases (`tenant1-3`, `players`, `things`) are still shared across lanes; scoping them per-lane is follow-on hardening.
- CIKafka's DLQ failure is pre-existing and handed off separately.

https://claude.ai/code/session_012vmmUkRtSFMk1Pbf9E5eor


jeremydmiller:
Session additions on top of the original branch (all verified locally with `--disable-test-retry` where applicable):

**The systemic find — leaked compliance hosts.** Thirteen fixtures across ASB/Kafka/NATS/Redis hid `TransportComplianceFixture.DisposeAsync` behind `public new ValueTask DisposeAsync() => CompletedTask` no-ops (pre-dating xUnit v3; the migration converted them mechanically). xUnit's class-fixture disposal dispatched into the no-op, so every such class leaked BOTH hosts process-long and the `AfterDisposeAsync` cleanup hooks never ran. A leaked receiver is a competing consumer — the ASB topic suites literally share `topic1`/`subscription1` — which was the entire mechanism behind the flaky ledger's headline entry (9 tests of the custom-rule class failing attempt 1 as a group, each passing alone in a fresh process). Retries-off before/after on the local emulator: 228/239 → 238/239 twice, custom-rule class fully green raw. Also widened `LeadershipElectionCompliance`'s forwarded-handling poll from a 1s ceiling to 20s (same early break).

**CISqlServer parallelized: 14m24s → 5m37s locally (390/390, zero retries at 4 workers).** Profiled first (60 classes, 931s total, largest 167s → 5.6× ceiling; ~2× at the hosted-runner clamp of 2). `sqlServerDatabasePerLane` mirrors the Postgres lane mechanism; `Servers.SqlServerDatabaseName` + `LaneDatabases.Name` de-literal the catalog/sibling-database names (three bug-shape-#1 assertions fixed, plus one test that only ever passed by test-ordering luck — its raw `delete from` bypassed the schema API's ensure-storage). `StartDockerServices` split into launch/await so the container boots during compilation.

**CIKafka**: green via #3760 on main (retry-tier native-DLQ inheritance product bug + per-test DLQ topic isolation), merged into this branch.

Expected matrix effect: CIKafka red gone, ASB flaky ledger ~10 → ≤1, CISqlServer ~16m → ~9-10m, and the pole is now solely CIAzureServiceBus's sequential runtime.

https://claude.ai/code/session_013eR4GL278688VhyhrGcttJ

Issue #3764, opened 2026-08-01
HTTP: opt-in ProblemDetails (409) mapping for Marten concurrency exceptions on [WriteAggregate] endpoints


uniquelau:
**Is your feature request related to a problem?**

A `[WriteAggregate]`/`[Aggregate]` HTTP endpoint that loses an optimistic-concurrency race surfaces `EventStreamUnexpectedMaxEventIdException` (a `JasperFx.ConcurrencyException`) from the transactional `SaveChangesAsync()` postprocessor as an unhandled exception → HTTP **500**.

Repro: two concurrent POSTs to the same aggregate's endpoint — the ordinary two-clients case, no version headers involved. We hit this with model-based testing of a production-shaped API (Wolverine 6.24.2 / Marten 9.22.2): racing generated operation pairs against `[WriteAggregate]` endpoints produced 57 raw 500s across 78 races. Data integrity held — optimistic concurrency did its job — but the HTTP contract leaks the internal failure. Message handlers have failure rules for exactly this (`OnException<ConcurrencyException>().RetryWithCooldown(...)` per the concurrency tutorial), but an `HttpChain` has no failure rules, so every app has to discover and hand-roll a mapping; 409 seems the semantically right default.

Prior art: #1490 (same escape path via a missing stream, acknowledged but closed unprioritized as "didn't come from real usage" — this report is the real-usage case) and discussion #2033 (asks exactly this, currently unanswered).

The `OnException` middleware convention *can* catch it — the generated try wraps the commit frame, as `use_stream_collision_policy` demonstrates for stream-id collisions — but there's no out-of-the-box mapping, no OpenAPI 409 metadata, low discoverability (hence #2033), and the obvious recipe (catch `ConcurrencyException`) silently misses `StreamLockedException` on the `FetchForExclusiveWriting` path, which derives from `MartenException`.

**Describe the solution you'd like**

An opt-in policy in Wolverine.Http.Marten:

```csharp
opts.UseProblemDetailsForConcurrencyExceptions(); // default 409, configurable
```

- Built on the existing `chain.GetOrCreateTryCatchFinallyFrame().AddCatchBlock(...)` machinery — essentially `StreamCollisionExceptionPolicy`, productized on the `OnException` convention.
[...]

PR #3765, opened 2026-08-01
Opt-in 409 ProblemDetails for Marten concurrency exceptions in Wolverine.Http

uniquelau:
Addresses #3764.

> ⚠️ **Please review with a healthy dose of skepticism.** This comes from consumers of Wolverine, not maintainers: we reproduced the behavior, studied the surrounding code (`StreamCollisionExceptionPolicy`, the `OnException`/`TryCatchFinallyFrame` machinery, `RequiredEntityPolicy`), and validated against the test suite — but we don't have deep knowledge of these internals or their design history, so take the approach with a pinch of salt. If a different shape fits better, we're glad to rework it, or treat this purely as a working starting point.

## What

`opts.UseProblemDetailsForConcurrencyExceptions(statusCode = 409)` — an opt-in `IHttpPolicy` in Wolverine.Http.Marten mapping commit-time concurrency failures to ProblemDetails instead of an unhandled 500.

- **Broad catch, scoped metadata**: any chain committing a Marten session catches `JasperFx.ConcurrencyException` (covers `EventStreamUnexpectedMaxEventIdException` and `DcbConcurrencyException` via inheritance); `ProducesProblem(statusCode)` is stamped only on write-shaped chains (aggregate handling present, or verbs beyond GET/HEAD) so `[ReadAggregate]` GETs under `AutoApplyTransactions` keep a clean OpenAPI surface.
- **`StreamLockedException`** (a `MartenException`, not a `ConcurrencyException`) caught solely on `ConcurrencyStyle.Exclusive` chains.
- **User code wins**: an endpoint's own `OnException` for the same exception type suppresses both the catch and the metadata stamp.
- **Never masks a streaming failure**: rethrows when `Response.HasStarted`, mirroring ASP.NET Core's `ExceptionHandlerMiddleware`.
- **Telemetry**: handled conflicts log at Information.

## Tests

8 tests in `concurrency_exception_problem_details.cs`, modelled on `use_stream_collision_policy`: deterministic stale-version 409 repro with body + OpenAPI metadata assertions; a live contended `FetchForExclusiveWriting` repro (competing session holds the stream lock, `Command Timeout` bounds the wait) → 409; a default-off pin proving the 500 escape is unchanged without the opt-in; a non-default status (412) test; and user-catch-wins pins at chain level. Two pre-existing tests that pinned escape-as-500 on version-carrying endpoints now assert the 409 ProblemDetails.

Full `Wolverine.Http.Tests`: **952 passed / 0 failed / 10 skipped** across four consecutive runs. One candid note: the first full run after the final build showed 2 failures we failed to capture and could not reproduce across four subsequent runs including a cold build (suspected schema-migration contention against the StreamLocked test host's initial 3s command timeout, since widened to 5s) — flagging for a CI observation cycle rather than pretending it didn't happen.

## Docs

New "Concurrency Exceptions" section in `docs/guide/http/marten.md` plus a cross-link from `exception-handling.md`, via region-tagged WolverineWebApi samples (mdsnippets-clean), including the hand-rolled `OnException` alternative and the supertype/`HasStarted` caveats.

### DIFF
--- a/docs/guide/http/exception-handling.md
+++ b/docs/guide/http/exception-handling.md
@@ -2,7 +2,7 @@
 
 Wolverine supports an `OnException` / `OnExceptionAsync` naming convention for middleware methods that allows you to handle exceptions thrown during endpoint execution. This is the recommended approach for structured exception handling in Wolverine HTTP endpoints.
 
-For mapping Marten concurrency exceptions to a `409 Conflict` with a `ProblemDetails` body across every Marten-backed endpoint at once, see [Concurrency Exceptions](/guide/http/marten.html#concurrency-exceptions).
+For mapping Marten concurrency exceptions to a `409 Conflict` with a `ProblemDetails` body across the endpoints using the Marten aggregate handler workflow or transactional middleware, see [Concurrency Exceptions](/guide/http/marten.html#concurrency-exceptions).
 
 ## Handler-Level Exception Handling
 
--- a/docs/guide/http/exception-handling.md
+++ b/docs/guide/http/exception-handling.md
@@ -2,6 +2,8 @@
 
 Wolverine supports an `OnException` / `OnExceptionAsync` naming convention for middleware methods that allows you to handle exceptions thrown during endpoint execution. This is the recommended approach for structured exception handling in Wolverine HTTP endpoints.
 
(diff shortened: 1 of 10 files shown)


jeremydmiller:
We're just going to make the docs suggest using an exception filter. Really cleaner than trying to do this in the codegen pipeline IMO

uniquelau:
Makes sense. I've pushed another commit to see what things look like when removing the try/catch, using the  IExceptionHandler and the IHttpPolicy to handle the OpenApi side things.

Issue #2307, opened 2026-03-16
Message Batching Listener Registration Fails When Using Service Bus + Conventional Routing


zjohnson-messagedigital:
I'm unable to convert an existing handler to support to support message batching
`public async Task HandleAsync(TransactionEtlEventV1 message, ITenantContext tenantContext, CancellationToken cancellationToken)` => `public async Task HandleAsync(TransactionEtlEventV1[] messages, ITenantContext tenantContext, CancellationToken cancellationToken)`
I'm following instructions from https://wolverinefx.net/guide/handlers/batching.html. I'm using Azure Service Bus Topic listeners configured with convention based routing. 

Verified the following:
* Code Gen works as expected
* No errors or warnings observed
* Inspecting WolverineRuntime Routing and not seeing listener registered.

PR #3789, opened 2026-08-03
Sanitize Azure Service Bus entity names, and stop losing the reason (GH-3786)

### LINKED ISSUE #3: Publish event messages for meaningful actions in Wolverine
Publishing messages will allow users to easily register handlers for big events. Thinking:

* circuit tripped off
* circuit resumed
* endpoint paused
* endpoint restarted
* endpoint failed?
* endpoint too deep?

All of these are available for the tracker too? The observable path is beneficial too maybe.

### LINKED ISSUE #3786: Conventional routing on Azure Service Bus fails broker initialization (hidden behind [Flaky] since March)
Conventional routing against Azure Service Bus fails to initialize the broker. Every affected test fails **100% of the time**, deterministically, on a clean emulator — this is not flakiness, and it has been hidden behind `[Trait("Category", "Flaky")]` since 2026-03-21.

## The failure

```
Wolverine.Transports.BrokerInitializationException : Unable to initialize the Broker asb in time
```

21 of 24 failures in the suite carry that exact message. Each one burns **~2m11s–2m27s** on the timeout before failing, which is why the affected classes dominate the suite's wall clock.

## Measured evidence

Full `Wolverine.AzureServiceBus.Tests` run, 2026-08-02, `net9.0`, Release, serial, against a **freshly recreated** emulator (`servicebus-emulator:2.0.1`) with the GH-3783 readiness gate in place, retries off:

**304 tests, 24 failed, 67.6 minutes — of which 57.0 minutes is these classes.**

| class | tests | failed | time |
|---|---|---|---|
| `ConventionalRouting.conventional_listener_discovery` | 6 | 5 | 12.55m |
| `ConventionalRouting.when_discovering_a_listening_endpoint_with_all_defaults` | 4 | 4 | 10.55m |
| `Bugs.Bug_1933_multi_tenant_conventional_routing` | 2 | 2 | 7.70m |
| `ConventionalRouting.when_discovering_a_sender_with_all_defaults` | 3 | 3 | 7.13m |
| `ConventionalRouting.when_discovering_a_listening_endpoint_with_overridden_queue_naming` | 3 | 3 | 6.09m |
| `ConventionalRouting.Broadcasting.end_to_end_with_conventional_routing` | 1 | 1 | 2.99m |
| `ConventionalRouting.end_to_end_with_conventional_routing_with_prefix` | 1 | 1 | 2.78m |
| `ConventionalRouting.end_to_end_with_conventional_routing` | 1 | 1 | 2.67m |
| `ConventionalRouting.discover_with_naming_prefix` | 1 | 1 | 2.39m |
| `end_to_end` | 6 | 2 | 2.03m |
| `Bugs.Bug_2307_batching_with_conventional_routing` | 1 | 1 | 0.08m |

Every single one is a conventional-routing test.

## A clue worth starting from

`ConventionalRouting.when_using_handler_type_naming` (2 tests) was **never tagged**, has therefore been running in CI all along, and **passes**. Whatever it does differently from its eleven neighbours is probably the shortest path to the cause.

`Bug_2307_batching_with_conventional_routing` is also worth noting separately: it fails in **0.08m**, not on the 2-minute broker timeout, so it may be a different (or downstream) failure.

## Why nobody noticed

These were tagged in a bulk sweep on 2026-03-21 (commits with messages like *"Tag all ConventionalRouting tests as Flaky"*), so they have not run in CI for over four months. None of the tags recorded a reason, a number, or an issue link, so there was nothing to indicate the tag was hiding a hard failure rather than a timing race.

Separately, GH-3783 found the ASB readiness gate had been declaring the emulator ready ~26s before its management API could serve provisioning. That is plausibly what the March sweep was *actually* measuring — but it does not explain these: the numbers above were taken **after** that fix, on a cold, clean emulator.

## Reproducing

```bash
docker compose up -d asb-emulator
dotnet test src/Transports/Azure/Wolverine.AzureServiceBus.Tests/Wolverine.AzureServiceBus.Tests.csproj \
  --framework net9.0 --filter "FullyQualifiedName~ConventionalRouting"
```

Recreate the emulator (`docker compose rm -sf asb-emulator asb-sql`) between runs — accumulated entities from prior runs change the timings.

## Scope

Fixing this restores 29 tests and, more importantly, tells us whether conventional routing on Azure Service Bus works at all — the tests are the only coverage of that combination and they have been dark since March.

jeremydmiller:
Closes #3786. Restores 29 tests: `CIAzureServiceBus` goes **271 → 300 passing**.

## The bug

Conventional routing derives broker entity names from the message type. Azure Service Bus accepts
only letters, numbers, `.`, `-` and `_` in an entity segment, so a handler taking an **array** —
`Handle(BatchedItem[] items)` — asked for a queue named `...bugs.batcheditem[]`:

```
400 SubCode=40000. '...batcheditem%5B%5D' contains character(s) that is not allowed by
Service Bus. Entity segments can contain only letters, numbers, periods (.), hyphens (-),
and underscores (_).
```

One such handler **anywhere in an assembly** took down broker startup for every conventionally-routed
host in it. Generic and nested message types (`<`, `>`, `,`, `+`) have the same shape.

That explains the clue in the issue: `when_using_handler_type_naming` passed all along because it
calls `DisableConventionalDiscovery().IncludeType(...)` and therefore never sees `BatchedItem[]`.

The dates confirm it. `Handle(BatchedItem[])` arrived in commit `5aeefe306` on **2026-03-16**; the
bulk *"Tag all ConventionalRouting tests as Flaky"* sweep was **2026-03-20/21**. The sweep was
mis-tagging a hard regression that had landed four days earlier.

## Why it hid for four months

`BrokerTransport.InitializeAsync` retries startup 20 times over two minutes and then throws
`BrokerInitializationException: Unable to initialize the Broker asb in time`. The actual cause went
to a log line — inside a job already dismissed as flaky. A `400 Bad Request` was never going to
succeed on attempt 20.

`BrokerInitializationException` now carries the last attempt's exception. That single change is what
made the rest of this tractable: the two further failures stacked behind the first became readable in
minutes.

## Fixes

1. `AzureServiceBusTransport.SanitizeIdentifier` substitutes illegal characters rather than only
   lowercasing. **Substituting, not stripping**, so `Foo[]` stays separable from `Foo`. No name that
   works today changes — a name containing an illegal character could never have been provisioned.
   `/` is preserved, since ASB allows hierarchical entity paths.
2. `AzureServiceBusTopicBroadcastingRoutingConvention` sanitizes a caller-supplied
   `SubscriptionNameForListener` result. `MessageRoutingConvention` runs every listener *identifier*
   through `MaybeCorrectName`, but the subscription name is computed inside this convention and
   skipped it — so a correctly-named topic got a subscription named `batcheditem[]`. Found only
   because fix #3 made it visible.
3. `BrokerInitializationException` carries the inner exception.

## Test changes

Every conventional-routing host is narrowed to the message types it actually asserts on.
Unrestricted discovery asked for **28 queues per host**, and the emulator caps a namespace at 50 —
so two classes in one process hit `SubCode=40300 ... Actual: 50, Max allowed: 50`, reported (again)
as a broker-init timeout.

The narrowing is an `ExcludeTypes`, **not** an `IncludeTypes`: `CompositeFilter` ANDs excludes and
ORs includes, so an include would have silently widened each test's own filter and made
`conventional_listener_discovery.include_types` pass whether the include filter worked or not.

`discover_with_naming_prefix` and `when_using_handler_type_naming` now purge in teardown. The first
left 28 *prefixed* queues behind, which is what pushed the classes after it over the cap.

Side effect worth having: the `ConventionalRouting` namespace went from **32m38s to 6m3s**.

## Still tagged — with tags corrected to what was measured

These were all labelled "broker timeout". None of them is one any more:

| class | now | cause |
|---|---|---|
| `end_to_end` | 2 of 6 | **session handling** — never related to this issue |
| `Bug_2307` | 1 of 1, in **5s** | plain assertion: `EndpointName` is the raw type name. A naming bug in the GH-2307 fix |
| `Bug_1933` | 2 of 2 → **1 of 2** | `TrackedSession` timeout; message Sent, never Received |

Each needs its own issue. I did not fold them into #3786.

## Verification

- Full `CIAzureServiceBus` locally against a **freshly recreated** emulator: **300 passed, 0 retries**.
- Every class verified **by name in isolation** as well as in the set — running "the set" is what
  produced the wrong conclusion twice on this issue.
- New unit tests cover the sanitizer directly, including the no-op guarantee for legal names and the
  `Foo[]` vs `Foo` non-collision.
- `dotnet build wolverine.slnx -c Release`: 0 warnings, 0 errors.

**Watch on this PR:** local wall clock for the whole suite was 16m26s against the 20-minute cap. CI
timings differ from local, so the `CIAzureServiceBus` duration here is the number to check before
merging.

https://claude.ai/code/session_0116vfBcKwcjWn8msM4ZjkuA

### DIFF
--- a/.github/workflows/dotnet.yml
+++ b/.github/workflows/dotnet.yml
@@ -36,6 +36,9 @@ jobs:
           echo "### Version: $version" >> $GITHUB_STEP_SUMMARY
 
       - name: Test
+        # Names this run in the retry ledger written to the job summary (build/RetryLedger.cs).
+        env:
+          CI_JOB_NAME: dotnet-ci
         run: ./build.sh ci
 
       # The CI target now also runs MessageRoutingTests, which boots a real
--- a/.github/workflows/http.yml
+++ b/.github/workflows/http.yml
@@ -35,11 +35,17 @@ jobs:
       - name: Exercise the openapi command
         run: ./build.sh OpenApiCommand --framework net9.0
 
+      # CI_JOB_NAME names each run in the retry ledger written to the job summary
+      # (build/RetryLedger.cs).
       - name: Run HTTP Tests
+        env:
+          CI_JOB_NAME: CIHttp
         run: ./build.sh CIHttp --framework net9.0
 
       # Asp.Versioning tests are pinned to net10.0
       - name: Run HTTP Asp.Versioning Tests
+        env:
+          CI_JOB_NAME: CIHttpAspVersioning
         run: ./build.sh CIHttpAspVersioning
 
       - name: Stop containers
--- a/.github/workflows/slow-tests.yml
+++ b/.github/workflows/slow-tests.yml
@@ -45,6 +45,9 @@ jobs:
           dotnet-version: 10.0.x
 
       - name: Run Tests
+        # Names this run in the retry ledger written to the job summary (build/RetryLedger.cs).
+        env:
+          CI_JOB_NAME: CISlowTests
         run: ./build.sh CISlowTests --framework net9.0
 
       - name: Stop containers
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -91,6 +91,9 @@ jobs:
       - name: Run Tests
         env:
           MEMORY_TELEMETRY: ${{ matrix.memory_telemetry }}
+          # Names this job in the retry ledger (build/RetryLedger.cs). GITHUB_JOB is the matrix's
+          # job id -- the same string, "test", for all thirty of these -- so it cannot be used.
+          CI_JOB_NAME: ${{ matrix.target }}
         run: |
           # The sampler has to live inside this step: when the runner is killed, later steps are
           # skipped, so only what has already been streamed to this step's log survives. See #3771.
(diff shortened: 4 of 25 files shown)


jeremydmiller:
### Timing follow-up — resolved, and a follow-up filed

The concern flagged in the description was real. Measured `CIAzureServiceBus`:

| | duration | headroom under the 20m cap |
|---|---|---|
| `main`, post-#3784 (271 tests) | 12.8 min | 7.2 min |
| this PR as first pushed (300 tests) | **18.7 min** | **1.3 min** |
| this PR now | **16.8 min** | **3.2 min** |

The recovered ~2 minutes came from sharing one host per class where the convention is identical.
Three classes assert several things about the *same* discovered endpoint — four facts about one
listener, three about one sender — and every assertion was starting Wolverine, provisioning the
emulator and purging it again. Ten hosts where three would do. Sharing is safe there precisely
because those assertions only read runtime state; `conventional_listener_discovery` still builds a
host per test, since each of its six tests configures a different convention.

Locally the whole suite went 16m26s → 14m38s, still 300 passed / 0 retries.

**This is a one-off win, not a trajectory change.** The durable answer is sharding the job, filed as
#3790. The handoff had deliberately rejected sharding on the grounds that *"57 of the 68 minutes were
broker timeouts, not capacity"* — true then, and no longer true now that this PR removes those
timeouts. That issue also records the two constraints any split has to respect: 264 of 306 tests sit
in the root namespace so the Polecat namespace-shard pattern doesn't port, and each shard needs its
own emulator because the namespace caps at 50 queues.

Issue #3797, opened 2026-08-03
Pulsar: requeue, scheduled retry and dead-letter routing are unimplemented


jeremydmiller:
Found while burning down the `Category=Flaky` ledger (#3763).

## What was hidden

On 2026-06-08, commit `02f531842` ("Disable Pulsar compliance tests" — no other rationale recorded) tagged all three Pulsar compliance classes `[Trait("Category", "Flaky")]`:

- `PulsarTransportComplianceTests`
- `InlinePulsarTransportComplianceTests`
- `with_cloud_events`

That is **66 of the project's 224 tests** — `CIPulsar` was running 158. Untagging them shows the exclusion was far too broad: **213 pass, 11 fail**, and the 11 are the *same four behaviours* repeated across the three fixtures.

## The four

| Test | Failure |
|---|---|
| `will_requeue_and_increment_attempts` | `Expected ending activity was not detected` |
| `can_schedule_retry` | `Expected ending activity was not detected` |
| `will_move_to_dead_letter_queue_without_any_exception_match` | `No ending activity detected` |
| `will_move_to_dead_letter_queue_with_exception_match` | `No ending activity detected` |

(`with_cloud_events` shows three of the four; it already overrode the fourth for an unrelated CloudEvents serialization limit.)

These are **not flaky**. They fail deterministically, every run, alone or in a suite. They look like genuinely unimplemented transport behaviour: native requeue with attempt tracking, scheduled retry, and dead-letter routing.

## What the burn-down PR does

Restores all three classes to CI and skips **only these four**, with the reason on each skip. That puts 55 previously-excluded Pulsar tests back into `CIPulsar` and leaves a precise, greppable record of what is actually missing instead of three blanket exclusions.

This issue tracks implementing the four behaviours (or, if any of them genuinely cannot apply to Pulsar, recording that on the skip and closing).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_0116vfBcKwcjWn8msM4ZjkuA

jeremydmiller:
Retitling: the CI half of this issue is done, and only the transport work remains.

[...]

Issue #3787, opened 2026-08-02
CI retry counts are invisible: a job burning 22 of 25 retries is indistinguishable from a clean green


jeremydmiller:
`CIAzureServiceBus` reported **green for four consecutive `main` runs** while burning **22 of its
25-retry budget** on every one of them — the same 22 tests, failing on first attempt and passing
alone in a fresh process. That was **85% of all flakiness in the repository**, and it was invisible.

Nothing in the GitHub checks UI distinguishes a job at 22/25 retries from one at 0/25. Both are a
green tick.

### What exists today, precisely

`build/SupervisedTests.cs` logs, via Serilog:

```
=== Wolverine.AzureServiceBus.Tests: 239 passed (0 retries, 2 worker processes) ===
  [FLAKY] <test> — passed on attempt 2
```

Nuke's GitHub Actions sink *does* turn those warnings into real `::warning` annotations, so the
Annotations panel is not empty. It is worse than empty — **it is truncated at exactly the number
that carries the signal.** On run 30759855844, the job that spent 22 retries:

```
$ gh api ".../actions/jobs/91528323381/logs" | grep -c FLAKY
66                       # 22 flaky tests, logged three ways

$ gh api ".../check-runs/91528323381/annotations?per_page=100" --paginate -q length
10                       # GitHub caps warning annotations at 10 per check run, first-wins
```

So the panel showed ten `[FLAKY]` lines. It would have shown ten at 10 retries, and ten at 50. The
**count** — the only thing that moved — is the exact quantity GitHub throws away, and the summary
line that does carry it (`0 retries` / `22 retries`) is a plain log line inside a job that passed,
which nobody has a reason to open. The ASB problem was found by opening one anyway, by accident.

### What matters

The absolute retry count mattered less than **the change in it**. The count stepped `1 → 22`
between two adjacent `main` runs and then held at exactly 22 for four runs. That step was the real
signal, and no baseline existed anywhere to notice it against.

### Proposal

1. **Job summary.** Each supervised run appends a retry ledger to `$GITHUB_STEP_SUMMARY`, so the
   numbers are on the run page without opening a log, and without competing for the ten annotation
   slots.
2. **A counted annotation, emitted first.** One `::warning` carrying the count, before the per-test
   `[FLAKY]` warnings, so it is inside the ten that survive.
3. **Run-over-run.** Each job publishes a machine-readable ledger artifact; a roll-up job aggregates
[...]

Issue #3763, opened 2026-08-01
Flaky-tag removal campaign: the Category=Flaky exclusion list is a debt ledger


jeremydmiller:
Every CI target filters `Category!=Flaky`, so a `[Trait("Category", "Flaky")]` tag is a permanent exclusion — those tests run nowhere. With the Bobcat supervisor now reporting pass-on-retry honestly (the `[FLAKY]` ledger), the tag should shrink toward zero: a test either gets fixed or gets deleted, not warehoused.

Findings from the 2026-08-01 stabilization work that reframe this list:

- **Two of the biggest "flaky" generators turned out to be real, fixable bugs**: the shared Kafka DLQ topic masked a product bug (retry tiers never inherited native-DLQ enablement — #3760), and the ASB group flake was thirteen fixtures leaking their hosts through `public new ValueTask DisposeAsync()` no-ops (fixed on #3758). Assume tagged tests hide the same two shapes: shared broker state and disposal/ordering luck.
- **Concrete first target**: `batch_processing_with_kafka.end_to_end` fails even in isolation (`This Wolverine application has a configuration for batching messages of type TestMessage, but there is no known handler for TestMessage[]` under `AlbaHost.For<Program>`) — that is not flakiness, it is broken, and the tag is hiding it.
- The #3726 CIRabbitMQ set (three chronic first-round failures) belongs to the same review.

Suggested mechanics: burn the list down a test at a time — untag, run it under `--disable-test-retry` locally ten times, fix or delete. The supervisor's ledger keeps any regression visible instead of silently retried.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_013eR4GL278688VhyhrGcttJ

jeremydmiller:
Two passes landed in #3780 (merged as `0d95ad0bf`). **Ledger: 53 → 44.** Both passes confirm the issue's framing — in every case so far the tag was hiding something with a mechanism, not a coin flip.

## Kafka — the concrete first target named in this issue

`batch_processing_with_kafka.end_to_end` failed on **every** run in isolation, as predicted. The cause was not batching:

```
APP=Wolverine.Kafka.Tests | DISCOVERY=Wolverine.RuntimeCompilation,Wolverine.Kafka.Tests
```

[...]

PR #3798, opened 2026-08-03
Restore 55 Pulsar tests and skip only the four behaviours that are missing (GH-3763)

jeremydmiller:
The Pulsar pass of #3763, after #3791 (AWS), #3794 (Kafka) and #3795 (RabbitMQ).

## An exclusion with no recorded reason, and far too broad

Commit `02f531842` on 2026-06-08 — message *"Disable Pulsar compliance tests"*, no other rationale — tagged all three Pulsar compliance classes `Category=Flaky`. That is **66 of the project's 224 tests**; `CIPulsar` has been running 158.

Untagged: **213 pass, 11 fail** — and the 11 are the *same four behaviours* repeated across the three fixtures.

| Test | Failure |
|---|---|
| `will_requeue_and_increment_attempts` | `Expected ending activity was not detected` |
| `can_schedule_retry` | `Expected ending activity was not detected` |
| `will_move_to_dead_letter_queue_without_any_exception_match` | `No ending activity detected` |
| `will_move_to_dead_letter_queue_with_exception_match` | `No ending activity detected` |

They are **not flaky** — they fail deterministically, every run, alone or in a suite, and they look like genuinely unimplemented transport behaviour: native requeue with attempt tracking, scheduled retry, and dead-letter routing. Filed as **#3797**.

## So: all three classes come off the ledger, only those four are skipped

Each skip carries the reason and the issue number, so what's missing is greppable instead of buried under three blanket exclusions. `can_schedule_retry` had to become `virtual` in `TransportCompliance` to be overridable; the other three already were.

`with_cloud_events` already opted out of `will_move_to_dead_letter_queue_with_exception_match` by overriding it with a bare `return Task.CompletedTask` — which **reports as a PASS**, so the suite was counting a test that ran nothing. That's now a real `Skip` carrying its original explanation.

**Net: `CIPulsar` goes from 158 executing tests to 213, plus 11 named skips.**

## Two things worth knowing that came out of this

- **`CIPulsar` runs Pulsar in Testcontainers, not docker-compose** (#3467). A local run hangs silently at `224 batched` if Docker is short on memory — nine stale containers were holding 4.7 of 7.7 GiB here and no Pulsar container could start. Stopping them fixed it; nothing in the output said so.
- **The supervisor's summary counts SKIPPED tests inside its "passed" number.** Run 2 below reports "224 passed" for 213 executed + 11 skipped. The per-assembly runner reports them correctly (`18 succeeded / 4 skipped` for one fixture). Worth a look given #3787's honest-reporting goal — not changed here.

## Verification

| | Result |
|---|---|
| CIPulsar, run 1 | 223 passed, 1 retry, **exit 0** |
| CIPulsar, run 2 | **224 passed, 0 retries**, exit 0 |

Repo flaky tags **16 → 13**. `wolverine.slnx` builds clean in Release.

https://claude.ai/code/session_0116vfBcKwcjWn8msM4ZjkuA

### DIFF
--- a/src/Testing/Wolverine.ComplianceTests/Compliance/TransportCompliance.cs
+++ b/src/Testing/Wolverine.ComplianceTests/Compliance/TransportCompliance.cs
@@ -578,7 +578,7 @@ public abstract class TransportCompliance<T> : IAsyncLifetime where T : Transpor
     }
 
     [Fact]
-    public async Task can_schedule_retry()
+    public virtual async Task can_schedule_retry()
     {
         throwOnAttempt<BadImageFormatException>(1);
 
--- a/src/Transports/Pulsar/Wolverine.Pulsar.Tests/InlinePulsarTransportComplianceTests.cs
+++ b/src/Transports/Pulsar/Wolverine.Pulsar.Tests/InlinePulsarTransportComplianceTests.cs
@@ -40,5 +40,22 @@ public class InlinePulsarTransportFixture : TransportComplianceFixture, IAsyncLi
 }
 
 [Collection("acceptance")]
-[Trait("Category", "Flaky")]
-public class InlinePulsarTransportComplianceTests : TransportCompliance<InlinePulsarTransportFixture>;
+public class InlinePulsarTransportComplianceTests : TransportCompliance<InlinePulsarTransportFixture>
+{
+    // GH-3763. These four are not flaky -- they fail deterministically, every run, in all three Pulsar
+    // compliance fixtures, with "No ending activity detected" / "Expected ending activity was not
+    // detected". They are the requeue, retry-scheduling and dead-letter behaviours the transport has not
+    // implemented. Skipping just these restores the other 55 tests in this file's three fixtures, which
+    // pass; the whole classes used to be excluded for them. See GH-3797.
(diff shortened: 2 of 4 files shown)


jeremydmiller:
Follow-ups filed out of this work, so the PR itself stays scoped to the tag burn-down:

- **#3797** — the four compliance behaviours that are actually missing (`will_requeue_and_increment_attempts`, `can_schedule_retry`, both `will_move_to_dead_letter_queue_*`). These are the 11 skips this PR adds; the issue is what un-skips them.
- **#3799** — the test containers: one heavy Pulsar broker per *worker process*, an unpinned `apachepulsar/pulsar:latest` that disagrees with the `4.0.3` pin in `docker-compose.yml`, and the silent hang at "224 batched" when Docker is out of memory. Includes a cheap fix: the compose `pulsar` service already exists and is unused, and the fixture already honours `WOLVERINE_PULSAR`.
- **#3800** — `will_move_to_dead_letter_queue_with_exception_match` under CloudEvents, split out because it is a shared-harness gap rather than a Pulsar defect: `ErrorCausingMessage` carries a `Dictionary<int, Exception>` that does not round-trip through `System.Text.Json`.

PR #3840, opened 2026-08-05
A deliberately paused listener reports the distinct ListeningStatus.Paused

### LINKED ISSUE #3832: ListeningStatus cannot distinguish an administrative pause from back-pressure TooBusy
Found while diagnosing two CritterWatch field issues (JasperFx/CritterWatch#915, #919/#922).

`IListenerCircuit.PauseAsync` leaves a listener reporting `Status == TooBusy`, and `DurableLocalQueue` maps its latched state to `TooBusy` unconditionally (`Latched ? ListeningStatus.TooBusy : ListeningStatus.Accepting`). So three very different conditions are indistinguishable to anything reading listener state:

1. **Back-pressure latched** (BufferingLimits exceeded — should self-resume),
2. **Administratively paused** (operator action / durable pause restriction — must NOT self-resume),
3. **Durable local queue latched** (agent restriction).

Two concrete costs already paid:
- A guard written against "paused" checked `Stopped`/`GloballyLatched`, compiled cleanly, and did nothing against the exact condition it existed for — because pause reports TooBusy (CritterWatch#915, the silent DLQ fan-out).
- Operators (and monitoring products) reading `TooBusy` cannot tell "transient back pressure, will recover" from "someone paused this and it will never recover on its own" — which is precisely the distinction that mattered in the CritterWatch#922 latched-ingest incident.

**Ask:** a distinct `ListeningStatus.Paused` (or an orthogonal `PauseReason`/`LatchReason` on the circuit) so state readers can distinguish self-recovering back pressure from operator intent. Happy to PR it if the shape is agreed — the main cost is auditing existing `TooBusy` switch/if sites (health checks in ExclusiveListenerFamily / DynamicListenerAgent / LeaderPinnedAgentFamily, WolverineTransportHealthCheck, BackPressureAgent's resume branch, which must NOT resume a Paused listener).

https://claude.ai/code/session_016v2Aijyo8MX2AdPUZL5VtG

jeremydmiller:
Closes #3832, with the shape agreed there: a distinct `ListeningStatus.Paused` rather than an orthogonal reason property.

## The problem

Three very different listener conditions were indistinguishable to state readers — back-pressure latched (self-resumes), administratively paused (must NOT self-resume), and durable-local-queue latched. `ListeningAgent.PauseCoreAsync` left a paused agent reporting a bare `Stopped`, and `DurableLocalQueue` mapped every latch to `TooBusy` unconditionally. Two costs already paid in the field: a guard written against "paused" compiled cleanly and did nothing (CritterWatch#915), and monitoring surfaces could not tell "transient back pressure, will recover" from "someone paused this and it will never recover on its own" (the CritterWatch#922 latched-ingest incident).

## The change

- **`ListeningStatus.Paused`**, appended last so the existing members keep their numeric wire values.
- **`ListeningAgent.PauseCoreAsync`** sets and publishes `Paused`; the `StopAndDrainCoreAsync` guard treats `Paused` as already-stopped-underneath.
- **`DurableLocalQueue`** tracks the administrative pause separately from `Latched`: `PauseAsync` → `Paused`; `LatchReceiver()` (the agent-restriction latch) deliberately keeps its `TooBusy` mapping in this PR — flagging that as a follow-up question rather than widening the change.
- **`BackPressureAgent`** needed no code change — its resume branch keys on `TooBusy` — but a new test pins that a `Paused` listener is never resumed by back-pressure relief, no matter how empty the queue.
- The three listener health checks (`ExclusiveListenerFamily`, `DynamicListenerAgent`, `LeaderPinnedAgentFamily`) report `Paused` as **Degraded** with an explicit "administratively paused and will not self-resume" message instead of falling through to Healthy.

## Audit of existing readers

Every `ListeningStatus` read site was audited: the `== Accepting` / `!= Accepting` guards (ListenerInboxRecovery, RecoverIncomingMessagesCommand, EndpointCollection, GlobalPartitionedRoute) already treat `Paused` correctly as not-accepting; `WolverineTransportHealthCheckAdapter` reads `TransportHealthStatus`, untouched.

## Tests

- `PausedListeningStatusTests` — live-host transitions for both circuit types: Accepting → `PauseAsync` → **Paused** → `StartAsync` → Accepting.
- `BackPressureAgentTests.never_restart_a_paused_listener_even_when_fully_drained`.
- Full CoreTests: 2251 green on net9.0 and net10.0.

https://claude.ai/code/session_016v2Aijyo8MX2AdPUZL5VtG

### DIFF
--- a/src/Testing/CoreTests/Transports/BackPressureAgentTests.cs
+++ b/src/Testing/CoreTests/Transports/BackPressureAgentTests.cs
@@ -108,6 +108,20 @@ public class BackPressureAgentTests
     }
 
     [Fact]
+    public async Task never_restart_a_paused_listener_even_when_fully_drained()
+    {
+        // GH-3832 — Paused is operator intent (or a circuit-breaker trip with its own Restarter).
+        // Back pressure relief must NOT resume it, no matter how empty the queue is.
+        theListeningAgent.Status.Returns(ListeningStatus.Paused);
+        theListeningAgent.QueueCount.Returns(0);
+
+        await theBackPressureAgent.CheckNowAsync();
+
+        await theListeningAgent.DidNotReceive().MarkAsTooBusyAndStopReceivingAsync();
+        await theListeningAgent.DidNotReceive().StartAsync();
+    }
+
+    [Fact]
     public async Task restart_when_too_busy_but_below_the_restart_threshold()
     {
         theListeningAgent.Status.Returns(ListeningStatus.TooBusy);
--- a/src/Testing/CoreTests/Transports/PausedListeningStatusTests.cs
+++ b/src/Testing/CoreTests/Transports/PausedListeningStatusTests.cs
@@ -15,9 +15,14 @@ namespace CoreTests.Transports;
 
 /// <summary>
 /// GH-3832 — a deliberate pause must report the distinct <see cref="ListeningStatus.Paused"/>,
-/// never the self-recovering <see cref="ListeningStatus.TooBusy"/> (and never linger as a bare
-/// Stopped). Before this, a guard written against "paused" had nothing to check for, and
-/// monitoring products could not tell operator intent from transient back pressure.
+/// never <see cref="ListeningStatus.TooBusy"/> (and never linger as a bare Stopped). Before this,
+/// a guard written against "paused" had nothing to check for, and monitoring products could not
+/// tell a timed pause from transient back pressure.
+///
+/// Both states do recover on their own; what differs is the trigger. A paused listener resumes
+/// when its interval elapses, a TooBusy one only once the queue drains below the restart
+/// threshold — so these tests pause for five minutes and restart explicitly rather than waiting
+/// out a timer.
 /// </summary>
 public class PausedListeningStatusTests : IAsyncLifetime
 {
@@ -79,8 +84,8 @@ public class PausedListeningStatusTests : IAsyncLifetime
 
         await circuit.PauseAsync(5.Minutes());
 
-        // The whole point of GH-3832: an administrative pause is not the self-recovering
-        // back-pressure latch, and must not read as one.
+        // The whole point of GH-3832: a timed pause is not the back-pressure latch, and must not
+        // read as one -- they recover on different triggers and only one is about load.
         circuit.Status.ShouldBe(ListeningStatus.Paused);
 
         await circuit.StartAsync();
--- a/src/Testing/CoreTests/Transports/PausedListeningStatusTests.cs
+++ b/src/Testing/CoreTests/Transports/PausedListeningStatusTests.cs
@@ -1,0 +1,90 @@
+using JasperFx.Core;
+using Microsoft.Extensions.Hosting;
+using Shouldly;
+using Wolverine;
+using Wolverine.ComplianceTests;
+using Wolverine.Runtime;
+using Wolverine.Tracking;
+using Wolverine.Transports;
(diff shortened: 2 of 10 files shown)


jeremydmiller:
Pushed `ec9711472`. The `CICircuitBreaking` failure was real and caused by this PR — but the mechanical fix would have baked in a bug, so the contract moved rather than just the assertions.

## The five failures

Four were the intended change (`agent.Status should be ListeningStatus.Stopped but was ListeningStatus.Paused`). The fifth, `pause_local_listener_on_matching_error_condition`, was the same cause wearing a different hat — it waits via `WaitForListenerStatusAsync("local", ListeningStatus.Stopped, 1.Minutes())`, and since `DurableLocalQueue` now publishes `Paused` that waiter never completed, so it surfaced as `TimeoutException` instead of an assertion.

## Why I didn't just update them

`ListeningStatus.Paused` documented itself as *"will NOT self-resume"* and named a circuit breaker trip as an example. But `PauseCoreAsync` ends with:

```csharp
Status = ListeningStatus.Paused;
...
_restarter = new Restarter(this, pauseTime);   // <-- schedules its own restart
```

Same in `DurableLocalQueue.PauseAsync`. Every caller of `PauseAsync(TimeSpan)` — the circuit breaker, `PauseListenerContinuation`, `RequeueContinuation` — installs a restart timer and does self-resume. The failing tests are the proof: each pauses, then waits for `Accepting` and gets it, and the circuit-breaker test is named `should_trip_and_restart`.

Left as written, a monitoring surface told "Paused means it will not recover on its own" would report every circuit-breaker trip as a permanent operator action, then watch it recover seconds later — the inverse of the CritterWatch#922 signal this PR exists to provide.

So: reworded, not redesigned. `Paused` now says the pause is for a fixed interval and resumes when it elapses, and states what the distinction actually buys — **recovery on a clock rather than on load**, so a reader watching a queue grow can tell "waiting for the pause to elapse" from "waiting for the backlog to clear". The three in-code comments and the two in `PausedListeningStatusTests` got the same correction. I also dropped the `<see cref>` to `BackPressureAgent`, which is `internal` and shouldn't appear in a public enum's docs.

## One test that was *not* failing

`everything_is_wonderful_even_though_there_are_some_failures_so_do_not_ever_trip` asserts:

```csharp
_observer.RecordedStates.ShouldNotContain(ListeningStatus.Stopped);
```

It stayed green and would have stayed green forever — trips publish `Paused` now, so it no longer observed anything. Moved to `Paused`, and verified it isn't vacuous by inverting it to `ShouldContain`: 7 of 7 variants fail, so the observer really does record `Paused`.

That's the pair worth noting — five tests went red because they asserted a value that changed, and this one went *silently vacuous* for exactly the same reason.

## Left alone

`stop_with_no_restart` and `force_restart_listener_3232` stay on `Stopped`. Both go through `StopAndDrainAsync`, which is a real stop with no restart timer.

## Verification

`CircuitBreakingTests` **24/24** (net9.0, against local Postgres + RabbitMQ), `CoreTests` `PausedListeningStatusTests` + `BackPressureAgentTests` **10/10**.

Two things in the original that I checked and left as-is because they're right: appending `Paused` last preserves the existing wire values, and the `or ListeningStatus.Paused` guard in `StopAndDrainCoreAsync` is necessary rather than incidental — `PauseCoreAsync` calls `StopAndDrainCoreAsync` first, so without it `pause_repeatedly` re-drains an already-paused agent.

Issue #3698, opened 2026-07-29
Agent assignment stalls at a few hundred of ~5650 agents and the leader goes idle (Wolverine 6.23.0, Marten multi-tenanted databases)


erdtsieck:
## Summary

With ~5650 distributed agents (Marten event subscriptions across 512 tenant databases + one durability agent per database), Wolverine's leader never converges on a full assignment.

It converges, but at roughly **9–10 agents per minute** and heavily skewed: two and a half hours after a rollout, 1481 of ~5240 agents are assigned and 1114 of those sit on the leader. Followers go 40 minutes at a stretch without gaining a single agent while the leader keeps adding to itself, and during those stretches **no `AssignmentChanged` record is written at all** — then it bursts back to ~50/s. Nodes stay healthy and no exceptions are logged. At this rate a full assignment takes about nine hours, and any rollout restarts it from scratch, so in practice most projections never run.

Two concrete leads, both detailed below:

- **5987 dead-lettered `Wolverine.Runtime.Agents.AgentsStarted` messages**, every one with `System.UriFormatException: Invalid URI: The URI is empty.` — i.e. the reply that confirms an agent start failing to parse. Dated 10–14 July, an earlier release, but on this same cluster and agent set.
- **Projection version bumps as the trigger.** The previous release had none and rolled out fine; every stuck agent belongs to one of the two projections whose version changed in the release that is stuck.

The per-node counts also land on exact multiples of `AgentStartBatchSize` early on, which is what first made us look at the chunked assignment path.

## Environment

- **WolverineFx 6.24.0, Marten 9.20.2, JasperFx 2.36.2**, .NET 10, PostgreSQL 18 (AWS RDS) — verified from the running container image, not from configuration. The five nodes in the measurements below started 16:26–16:30 local, which is exactly the rollout of that build.
  <br>_(Correction: an earlier version of this issue said 6.23.0 / 9.20.0 / 2.36.1. That came from a `ddpv` tag inside a persisted `mt_event_progression.pause_reason`, which is written once and never cleared, so it named a build from earlier in the day. The stall reproduces on the current 6.24.0.)_
- `IntegrateWithWolverine(opts => { opts.MainDatabaseConnectionString = ...; opts.UseWolverineManagedEventSubscriptionDistribution = true; })` — always on; we never register `AddAsyncDaemon(...)`
- `MultiTenantedWithShardedDatabases` — **512** tenant databases on one server, ~854 tenants
[...]

Issue #3753, opened 2026-07-31
Node assignment is very slow, or never completes, when a release contains a projection version bump


erdtsieck:
Filing this as the plain statement of the problem, deliberately without diagnosis, so the outcome we need does
not get lost among the individual defects.

## What we need

After a deploy, every agent assigned across every node, within minutes.

## What happens when a projection version bump is in the release

Node assignment either takes hours or does not converge at all. Nodes sit at zero while others carry
everything; the total number of assigned agents can go *down* over time; and the cluster never reaches a state
where all agents are running.

Measured today on a canary (restored production copy, 512 tenant databases, ~6,500 agents, 5 nodes), across a
series of attempts:

| | assigned, of ~6,500 | shape |
|---|---|---|
| stock settings, 40 minutes in | 498 | 157 / 311 / 376 |
| after tuning, 12 minutes in | 814 | 799 / 0 / 0 / 0 / 1 — two nodes never started a single agent |
| a later attempt, stable for 7 minutes | 799 | one node holding everything, four at zero |
| a four-minute window | 2,401 → 1,885 | 400+ agents stopped against ~5 started |
| best configuration found, 8 minutes in | 2,788 | 795 / 798 / 400 / 399 / 396, still climbing |

The last row is after a day of tuning derived from reading the source, and it is on track to converge in
roughly 35 minutes. That is the *good* case.

## What it looks like without a version bump

**Production** — which this canary is a restored copy of: same shape, same 512 tenant databases, same
~6,485 agents — sits fully converged at **1,301 / 1,301 / 1,295 / 1,294 / 1,294** across five nodes, with no
version bump in flight. So distribution itself is healthy — this is specific to a release in which a
projection's version has changed.

## Why this matters as its own issue

We cannot ship a schema-affecting release to production. Every release that touches a projection is a release
that cannot be deployed with confidence, and the workaround we found is a set of four settings whose correct
values we could only derive by reading three source files — not something anyone will reconstruct during an
incident.

I have filed what I believe are the contributing causes separately (#3748, #3749, #3750, and
JasperFx/jasperfx#594), but I would like this one to stay open until a deploy carrying a projection version
bump converges quickly on its own, because that is the actual requirement. Happy to be the verification site:
this cluster reproduces at full production scale and we can iterate on it quickly.

jeremydmiller:
[...]

Issue #3749, opened 2026-07-31
ReassignAgent blocks the destination node's lane on a round trip to the source, and is the one command type never batched


erdtsieck:
## Problem

`ReassignAgent` runs in the **destination** node's dispatcher lane but does its blocking round trip against
the **source** node. Lanes are strictly serial, so a source that is slow to answer — or gone — freezes every
command queued for an entirely *healthy* destination, including the batched `AssignAgents` chunks behind it.

- `ReassignAgent.DestinationNodeId` → `ActiveNode.NodeId` (`ReassignAgent.cs:22`), unless the narrow
  `StopInSourceLane` flag is set.
- `ReassignAgent.ExecuteAsync` (`:28`) → `InvokeAsync(OriginalNode, new StopAgent(AgentUri))`, no explicit
  timeout, and only `UnknownWolverineNodeException` is caught. A source still in `wolverine_nodes` but dead or
  merely busy just waits.
- `AgentCommandDispatcher`'s doc comment states the invariant the lanes exist for — *"a lane wedged on a dead
  node drains nothing but its own queue"* — and this is the one command that breaks it, because the node it
  wedges on is not the node whose lane it is in.
- `NodeAgentController.EvaluateAssignments.batchCommands` (lines 309–334) batches `AssignAgent` and
  `StopRemoteAgent` but has **no case for `ReassignAgent`**, so a rebalance emits thousands of these and
  `AgentStartBatchSize` has no effect on any of them (verified at 50, 100, 400, 1500).

## Evidence

A stable five-node cluster, every pod `Ready 1/1`, every heartbeat 1–7s old, sat **frozen for seven minutes**:
one node holding 799 agents, four holding zero, ~5,600 agents unplaced. The leader's decisions were correct
and evenly spread the whole time — parsing `DestinationNodeId` out of the `AssignmentChanged` descriptions:

| minute | → 513 | → 514 | → 515 | → 516 | → 517 |
|---|---|---|---|---|---|
| 15:26 | 2,401 | 2,701 | 1,708 | | |
| 15:27 | 1,639 | 1,661 | 1,638 | 1,769 | |
| 15:28 | 275 | 162 | 165 | 56 | 1,747 |

Four nodes handed 1,600–2,700 placements each landed **zero**, and logged nothing at all — consistent with
the commands never reaching them, since `StartAgents.ExecuteAsync` logs per agent either a success or
`Failed to start requested agent`.

### A controlled comparison inside one cluster, one moment

[...]

Issue #3604, opened 2026-07-23
Exclusive listener agent does not reassign reliably to a survivor after its node leaves the cluster


jeremydmiller:
## Summary

While building the multi-node end-to-end coverage for #3597, I could not get an `ExclusiveListenerFamily` agent (`ListenWithStrictOrdering()` / `ExclusiveNodeWithParallelism()`) to reassign reliably to a surviving node after the node hosting it left the cluster. Instead of settling on a survivor, the listener agent **flapped** — it would start briefly on another node, then drop again — and never reached a stable running state within a generous timeout.

This is independent of the #3590 inbox-recovery fix (that concerns which node recovers dormant `owner_id = 0` rows; this is about which node *runs the listener agent* at all after a failover).

## Repro shape

Three `Balanced` hosts sharing one PostgreSQL store, a Rabbit queue configured `.ListenWithStrictOrdering().UseDurableInbox()`:

1. Wait for the `ExclusiveListenerFamily` agent to settle on some non-leader node.
2. Stop that node (both a graceful `StopAsync()` and the `DisableHealthChecks()` + `StopAsync()` teardown were tried).
3. Wait for the listener agent to appear as running on a surviving node.

Observed: log lines like `Successfully started agent wolverine-listener://rabbitmq/<queue> on Node 2` followed by the agent no longer being reported by `RunningAgents()` on any node — i.e. it starts and then leaves, repeatedly. Tightening `StaleNodeTimeout` (→ 5s) and `CheckAssignmentPeriod` (→ 1s), releasing any agent pin before the stop, and extending the wait to 90s did not produce a stable reassignment.

## Why it matters

`ExclusiveListenerFamily` is the mechanism behind strictly-ordered / single-active-node listeners. If its agent does not fail over cleanly when a node leaves, a strictly-ordered listener can be left not running anywhere until something else nudges assignment — which is exactly the kind of silent stall #3590 was chasing on the recovery side.

## Coverage gap

There is currently **no** failover test for `ExclusiveListenerFamily` — `exclusive_listeners.cs` covers steady-state spreading (`spread_out_the_exclusive_listeners`) but never kills a node. #3597's end-to-end test deliberately reproduces the *inbox-recovery ordering* deterministically (seeding rows under a not-in-cluster owner id, then releasing them) rather than encoding a flaky node-kill, so this failover behavior is still unverified.

## Ask

[...]

PR #3853, opened 2026-08-05
Extend the pending-assignment ledger to reassignments (GH-3852)

### LINKED ISSUE #3852: Leader re-decides ~45,000 ReassignAgent placements in six minutes during a ramp — is it still true now that jasperfx#610 has landed?
Split out of #3785 so it does not close with that issue. Offered there as an observation without a conclusion, and it is the one thread from that report which is neither retracted nor resolved elsewhere.

## The observation

From a 512-database, 5-node production cluster during an assignment ramp (Wolverine 6.24.4 / Marten 9.22.2 / JasperFx 2.37.2):

> over a six-minute window during the ramp we counted **~45,000 `ReassignAgent` decisions** in the `AssignmentChanged` descriptions, essentially all aimed at the nodes that already held everything and about twenty at each new node

For scale, the cluster holds ~8,700 agents. Forty-five thousand reassignment *decisions* in six minutes is roughly five decisions per agent per six minutes, overwhelmingly re-aimed at nodes that were already saturated.

## The standing hypothesis

The reporter's own reading, which I think is the right first thing to test:

> Given that the outcome converges, this may simply be the leader re-deciding against a fleet whose starts have not confirmed yet — in which case it is a symptom of [jasperfx#610] too, and only worth looking at if it is cheap to make the evaluation aware of pending starts.

That is: while a bumped projection's shards warm up, the replay ran **inside** the agent start path, so an agent did not count as assigned until its replay finished. The leader re-evaluates against a fleet that looks emptier than it is, and re-decides the same placements repeatedly.

## Why it is testable now rather than speculative

jasperfx#610 — *"Move the blue/green side-effect gate warm-up out of the agent start path"* — is **closed**, landed in JasperFx 2.38.0, and reached Wolverine in **6.24.6** (JasperFx 2.39.5). The original measurement was taken on 6.24.4, which carried JasperFx 2.37.2 and therefore still had the warm-up on the start path.

So the hypothesis has a clean test: re-count `ReassignAgent` decisions over a comparable ramp window on 6.24.6.

- **Decision count collapses** → it was #610, nothing further to do here, close this.
- **Decision count stays high** → the leader is genuinely re-deciding against stale state, and the "make the evaluation aware of pending starts" question becomes real work rather than a maybe.

## Why it is worth keeping even if benign

Each decision writes an `AssignmentChanged` row. #3658 already established that the node-record table is a real capacity concern on this cluster shape — the reporting cluster wrote ~12.8M rows/day and reached 36M rows / 16 GB inside the retention window, which is what `TrimNodeRecordsCommand` and the row cap exist for. 45,000 decisions per six minutes during every ramp feeds directly into that, whether or not the decisions themselves are harmful.

## Not in scope

The convergence and delay complaints from #3785 — those were retracted or resolved (jasperfx#610, #3753). This issue is only the decision *volume* and whether the leader should account for pending starts.

Related: #3785 (origin), #3753 (outcome-level slow assignment), #3658 (node-record volume), JasperFx/jasperfx#610.

jeremydmiller:
Closes #3852.

## The gap

The GH-3698 pending-assignment ledger closed the re-decision hole for first-time assignments and silently excluded every reassignment. An agent being moved is still listed in its **source** node's persisted `ActiveAgents`, so `Agent.OriginalNode` is set, and `applyPendingAssignments` skipped it outright:

```csharp
// Already running somewhere: not a pending placement.
if (agent.OriginalNode != null) continue;
```

The ledger *armed* on a `ReassignAgent` but could never *apply* one, and `TryBuildAssignmentCommand`'s `OriginalNode != null` branch consulted no ledger at all — it went straight from "assigned != original" to a fresh `ReassignAgent`. So the leader re-decided the same move from scratch on every evaluation until the source's assignment row finally disappeared.

## Scale

Against the reported cluster shape — 512 databases x 17 agents across five nodes, group affinity, three incumbents holding everything and two newcomers ramping in — with the persisted snapshot frozen:

```
total agents: 8704
cycle 1: reassign decisions = 3468
cycle 2: reassign decisions = 3468
...
cycle 6: reassign decisions = 3468
```

Roughly 13 cycles of that is the reported ~45,000 in six minutes. Each decision writes an `AssignmentChanged` row — `_observer.AssignmentsChanged` runs *before* `batchCommands` and before the dispatcher, so nothing downstream dedupes it — feeding the node-record volume problem of #3658 at full rate.

The outcome converged because `ReassignAgents` carries hand-written set-based value equality, so the dispatcher collapses an identical re-emitted batch while its lane is busy. That hold ends the moment the batch completes, though, so a re-decision landing before the snapshot catches up re-runs a real `StopAgents` round trip against a source that has already let go.

**On the version framing in the issue:** 6.24.4 already carried the GH-3698 ledger (landed 7/29, ahead of the 8/1 bump), so the original measurement was taken *with* the intended mitigation in place. jasperfx#610 shortens the staleness window and reduces the absolute count on 6.24.6, but cannot close the gap — any slow ramp re-opens it at the same per-cycle rate.

## Changes

- `applyPendingAssignments` ends the wait only on a **matching** node, so a move in flight sets `PendingNode` and holds its placement (before the family distributes, so the distribution balances around it rather than re-splitting and yanking it back).
- `TryBuildAssignmentCommand` returns false for a move already dispatched and still live. `PendingRetryDue` re-drives one that is never confirmed.
- `AgentCommandDispatcher` tracks reassignments in a `_moving` map, keyed to the node the agents are moving **to** — not the command's lane, which is the source. Kept deliberately separate from `_inFlight` because `Enqueue` must never suppress a reassignment: it carries a stop, and dropping that is the two-copies bug `StartedAgentsOf` guards against. Without this map the hold falls back to the ledger TTL (2 x `CheckAssignmentPeriod`, 60s), which a batch's own reply window trivially outlives.

## Coverage

New `pending_reassignment_ledger`. 4 of the 8 fail without the fix; the rest guard against over-suppression — a pause still stops an agent mid-move, confirmation still clears the ledger, an unconfirmed move is still re-driven, and the dispatcher still collapses an identical in-flight batch.

Full `CoreTests` green (2272 passed), `wolverine.slnx` Release build clean.

https://claude.ai/code/session_01WHAuhdWS3XeAk16swV9G8m

### DIFF
--- a/src/Testing/CoreTests/Runtime/Agents/pending_reassignment_ledger.cs
+++ b/src/Testing/CoreTests/Runtime/Agents/pending_reassignment_ledger.cs
@@ -1,0 +1,309 @@
+using CoreTests.Transports;
+using JasperFx.Core;
+using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.Logging.Abstractions;
+using NSubstitute;
+using Shouldly;
+using Wolverine.ComplianceTests;
+using Wolverine.Configuration;
+using Wolverine.Runtime;
+using Wolverine.Runtime.Agents;
+using Xunit;
+
+namespace CoreTests.Runtime.Agents;
+
+/// <summary>
+/// GH-3852. The pending-assignment ledger (GH-3698) closed the re-decision hole for first-time assignments
+/// but silently excluded every reassignment: an agent being moved is still listed in its SOURCE node's
+/// persisted ActiveAgents, so <c>Agent.OriginalNode</c> is set, and <c>applyPendingAssignments</c> skipped
+/// it outright. The ledger armed on a <see cref="ReassignAgent" /> and could never apply one, so the leader
+/// re-decided the same move from scratch on every evaluation until the source's assignment row finally
+/// disappeared — one <c>AssignmentChanged</c> row per moved agent per cycle, plus a redundant
+/// <c>StopAgents</c> round trip once the dispatcher's in-flight hold had been released.
+///
+/// <para>Reported from a 512-database, 5-node cluster: ~45,000 reassignment decisions in a six-minute ramp
+/// window against ~8,700 agents.</para>
+/// </summary>
+public class pending_reassignment_ledger
+{
+    private readonly WolverineOptions _options;
+    private readonly IWolverineRuntime _runtime;
+    private readonly INodeAgentPersistence _persistence = Substitute.For<INodeAgentPersistence>();
+    private readonly FakeAgentFamily _family = new("fake", 40);
+    private readonly NodeAgentController _controller;
+    private readonly WolverineNode _incumbent;
+    private readonly WolverineNode _newcomer;
+
+    public pending_reassignment_ledger()
+    {
+        _options = new WolverineOptions { ApplicationAssembly = GetType().Assembly };
+        _options.Transports.NodeControlEndpoint = new FakeEndpoint("fake://self".ToUri(), EndpointRole.System);
+        _options.Durability.DurabilityAgentEnabled = false;
+
+        _runtime = Substitute.For<IWolverineRuntime>();
+        _runtime.Options.Returns(_options);
+        _runtime.DurabilitySettings.Returns(_options.Durability);
+        _runtime.Observer.Returns(Substitute.For<IWolverineObserver>());
+
+        _controller = new NodeAgentController(
+            _runtime, _persistence, [_family], NullLogger<NodeAgentController>.Instance, CancellationToken.None);
+
+        _incumbent = new WolverineNode
+        {
+            NodeId = _options.UniqueNodeId,
+            AssignedNodeNumber = 1,
+            ControlUri = _options.Transports.NodeControlEndpoint!.Uri
+        };
+        _incumbent.Capabilities.AddRange(_family.AllAgentUris());
+
+        _newcomer = new WolverineNode
+        {
+            NodeId = Guid.NewGuid(),
+            AssignedNodeNumber = 2,
+            ControlUri = "fake://two".ToUri()
+        };
+        _newcomer.Capabilities.AddRange(_family.AllAgentUris());
+    }
+
+    private Task<AgentCommands> evaluateAsync()
+        => _controller.EvaluateAssignmentsAsync([_incumbent, _newcomer], new AgentRestrictions());
+
+    private static int reassignedAgentCount(AgentCommands commands)
+        => commands.OfType<ReassignAgents>().Sum(x => x.AgentUris.Length)
+           + commands.OfType<ReassignAgent>().Count();
+
+    /// <summary>
+    /// The ramp shape: one node holding everything, a newcomer holding nothing. The rebalance is decided
+    /// once; while the moves are still executing, an unchanged node-state snapshot must decide nothing.
+    /// </summary>
+    [Fact]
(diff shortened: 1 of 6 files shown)


jeremydmiller:
## SlowTests follow-up — found and fixed a real regression, then a pre-existing one

`SlowTests` runs in no CI workflow, so I ran the agent suites by hand. `agent_reassignment_at_scale` (the GH-3753/GH-3749 harness) went **red** on the first commit. Baselined against `origin/main` in a worktree to be sure it wasn't ambient noise — main passed, the branch failed on two assertions.

The timelines are identical for the first 15 seconds, then split hard:

```
        main                              branch (ledger fix only)
  16s | 1: 234  2: 238  3:   2  | 474    16s | 1: 234  2: 238  3:   0  | 472
  20s | 1: 232  2: 236  3:   4  | 472    20s | 1: 232  2: 236  3:   0  | 468
  30s | 1: 227  2: 231  3:  13  | 471    30s | 1: 227  2: 231  3:   0  | 458
  40s | 1: 222  2: 226  3:  23  | 471    40s | 1: 222  2: 226  3:   0  | 448
```

Main's node 3 climbs while the total stays flat at 471 — stops and starts interleave. The branch's node 3 sits at **0** while the total *declines* 2/sec: the stops land and the starts never do, then arrive in big step jumps (90 → 110 → 120 → 140 → 160) at the very end.

### Cause

`StopAgents.ExecuteAsync` ran a plain serial `foreach`. `StartAgents.StartBatchAsync` got bounded parallelism in GH-3604/D3 for exactly this reason — *"daemon-agent starts are I/O bound, so a 50-agent chunk started one-at-a-time was seconds of dead wall-clock that blew the reply window"* — and the stop side, which is the same shape, never did.

So at `AgentStartBatchSize = 50` against a source slow to let go, the entire chunk's stop cost runs in series before `ReassignAgents` can cascade a single start, and every agent in the chunk is down for the duration.

**That was survivable only because of the churn this PR removes.** The leader re-deciding the same move every cycle was trickling agents onto the destination through individual `ReassignAgent` commands alongside the batch. Take the churn away and the serial stop is exposed as the actual cost. The bug was always there; the wasteful re-decisions were hiding it.

### Result

`agent_reassignment_at_scale`:

| | main | ledger fix | + parallel stops |
|---|---|---|---|
| fresh nodes >=20 each | 38.0s | 74.1s | **14.0s** |
| converged at | 176.3s | 176.3s | **31.1s** |
| worst backslide | 8 | 72 | 40 |

Convergence is **5.7x faster than main**, not merely restored. The backslide is deeper than main's 8 but well inside the test's `<= 70` bar ("never worse than one reassignment chunk in flight"), and it now lasts seconds instead of the ~100s a serial chunk took — far fewer agent-seconds spent down.

`StopLocallyAsync` is per-agent independent (a `ConcurrentDictionary` lookup and removal, the agent's own `StopAsync`, a persistence delete), so the fan-out is safe for distinct URIs exactly as on the start side. The per-agent `try`/`catch` is preserved so one failing stop still costs only itself.

### Verification

- All 3 `SlowTests.Agents` suites green.
- Full `CoreTests` green (2275 passed).
- New `stop_agents_batch_parallelism` — 3 tests, and the bounded-parallelism one is red against the old serial loop.
- `wolverine.slnx` Release build clean.

Issue #3871, opened 2026-08-07
RabbitMQ: UseQuorumQueues conflicts with native DLQ after upgrade to 6.24.10


leorg99:
**Describe the bug**

When RabbitMQ native dead-letter queues are enabled (the default) together with `UseQuorumQueues()` and `AutoProvision()`, Wolverine can declare the shared `wolverine-dead-letter-queue` as a classic queue and later redeclare it as a quorum queue.

RabbitMQ rejects the second declaration because queue type is immutable. The listener then fails to start and no consumer is created.

**Regression**

This first appeared after upgrading our messaging packages from:

- WolverineFx `6.16.0` → `6.24.10`
- RabbitMQ.Client `7.2.1` → `7.2.2`

With `6.24.10`, the failure reproduces consistently against a newly created empty RabbitMQ `3.11` broker.

We have not yet isolated whether the behavioral change is in Wolverine or RabbitMQ.Client, so this is reported as a suspected regression rather than attributed definitively to one package.

**To Reproduce**

Use a fresh RabbitMQ broker and this topology:

```csharp
builder.Services.AddWolverine(options =>
{
    options
        .UseRabbitMq(new Uri("amqp://guest:guest@localhost:5672/"))
        .UseQuorumQueues()
        .AutoProvision();

    options.ListenToRabbitQueue("orders");
});
```

Environment:

- WolverineFx.RabbitMQ `6.24.10`
- RabbitMQ.Client `7.2.2`
- RabbitMQ `3.11`
- .NET `10`
- Empty broker with no pre-existing queues

Start the host, then inspect the queues:

```text
wolverine-dead-letter-queue     durable=true  arguments=[]
orders                          durable=true  x-queue-type=quorum,
                                             x-dead-letter-exchange=wolverine-dead-letter-queue
```

Startup fails with:

```text
PRECONDITION_FAILED - inequivalent arg 'x-queue-type' for queue
'wolverine-dead-letter-queue': received the value 'quorum' of type 'longstr'
but current is none
```

The same failure is reproduced by our integration host with the focused `AlertTriggerMessagingTests` messaging tests against a newly created, worktree-scoped RabbitMQ container.

**Expected behavior**

Startup should complete successfully, and every auto-provisioned queue should be declared consistently. If the shared native DLQ is intended to remain classic, it should not subsequently receive the quorum policy. If it is intended to be quorum, its first declaration should include `x-queue-type=quorum`.

**Additional context**

The relevant implementation path appears to be:

[...]

PR #3876, opened 2026-08-07
#3871 follow-up: narrow the system endpoint compile to Rabbit MQ

jeremydmiller:
Follow-up to #3874.

## What changes

The #3871 fix compiled **every** endpoint in `BrokerTransport.InitializeEndpointsAsync`, which changed start-up behavior for all ten broker transports in order to solve a RabbitMQ problem.

`Endpoint.Compile()` is latched by `_hasCompiled`, so the only endpoints actually affected were the ones created inside `tryBuildSystemEndpoints` — for RabbitMQ the dead letter queue and its exchange, for Kafka the `wolverine.topics` endpoint, and so on for the rest. Only the RabbitMQ ones needed it.

This moves the compile into `RabbitMqTransport.tryBuildSystemEndpoints`, right where those two endpoints are created, and leaves every other transport on the code path it had before. The base class keeps a comment describing the contract so the next transport that builds system endpoints knows it owns compiling them.

No behavior change for RabbitMQ — `Bug_3871_quorum_queues_with_native_dead_lettering` passes on both shapes.

## Why now, and what is *not* established

This was prompted by `CIKafka` hitting its 20-minute cap on the #3871 PR — 18.5 minutes of test execution against ~8.5 minutes on a passing main run, with the job cancelled before it could report.

**I have not established that the #3871 change caused that**, and the evidence is genuinely mixed:

| Probe | Result |
|---|---|
| CIKafka on 5 recent main runs | ~10m20s, tight cluster |
| CIKafka on #3874 (with the change) | 18.5 min → timeout |
| CIKafka control on #3875 (no `BrokerTransport` change), same hour | **10m14s, normal** — rules out "slow runners today" |
| Local A/B, 75 Kafka tests (compliance + DLQ + broadcast) | **4m49s with the change vs 4m54s without** — a tie, the reverted run marginally slower |

The control says CI was healthy; the local A/B says this code costs nothing on the tests I can run. Those pull in opposite directions. The local subset is 75 of 301 tests and excludes the per-tenant tests, which need a second broker at `localhost:9192` that the repo's compose file doesn't provide — so a regression confined to that portion would not show up in my measurement.

CIKafka on main post-merge is the outstanding probe; it has been stuck in the GitHub queue and had not run at the time of writing.

**The narrowing is worth doing on scope grounds regardless of how that lands**: fixing a RabbitMQ declaration bug inside `RabbitMqTransport` is better-scoped than changing start-up for ten transports. If CIKafka on main turns out to be fine, this PR is still the right shape; if it times out again, this removes the change from Kafka's path entirely.

https://claude.ai/code/session_01JKfy5EzLX1i149gjUb3Tfg

### DIFF
--- a/src/Transports/RabbitMQ/Wolverine.RabbitMQ/Internal/RabbitMqTransport.cs
+++ b/src/Transports/RabbitMQ/Wolverine.RabbitMQ/Internal/RabbitMqTransport.cs
@@ -345,6 +345,16 @@ public partial class RabbitMqTransport : BrokerTransport<RabbitMqEndpoint>, IAsy
             deadLetterQueue.ConfigureExchange?.Invoke(dlqExchange);
 
             dlq.BindExchange(dlqExchange.Name, deadLetterQueue.BindingName);
+
+            // These two were created after the caller's Compile() loop had already run, so they are
+            // the only endpoints in the transport that would otherwise never have the endpoint
+            // policies applied to them. Resource setup declares whatever it finds here without
+            // compiling anything, so leaving them uncompiled means UseQuorumQueues() (which targets
+            // every Application role queue, and the lazily created DLQ is one) does not reach the
+            // dead letter queue until start up redeclares it -- and Rabbit MQ rejects a second
+            // declaration that changes a queue's type. See GH-3871.
+            dlq.Compile(runtime);
+            dlqExchange.Compile(runtime);
         }
     }
 
--- a/src/Wolverine/Transports/BrokerTransport.cs
+++ b/src/Wolverine/Transports/BrokerTransport.cs
@@ -126,17 +126,12 @@ public abstract class BrokerTransport<TEndpoint> : TransportBase<TEndpoint>, IBr
             endpoint.Compile(runtime);
         }
 
+        // A transport that builds system endpoints needing the endpoint policies applied to them --
+        // a dead letter queue whose type must match what the runtime declares later, say -- has to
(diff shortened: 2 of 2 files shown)


jeremydmiller:
## Verdict on the Kafka question: this PR's premise is gone

`CIKafka` on **main, with the #3871 change merged**, succeeded in **10m33s** (17:49:57 → 18:00:30) — right at the ~10m20s norm.

That is the decisive probe, and all three now agree that the #3871 change costs Kafka nothing:

| Probe | Result |
|---|---|
| Local A/B, 75 Kafka tests | 4m49s with vs 4m54s without — tie |
| Control on #3875 (no `BrokerTransport` change), same hour | 10m14s — normal |
| **Main post-merge, with the change** | **10m33s — normal** |

The `CIKafka` timeout on #3874 was an unreproducible one-off.

## Recommendation: close this without merging

With the performance premise removed, the remaining argument was scope — but that argument now cuts the *other* way. `RabbitMqTransport` is not the only transport that builds dead letter endpoints inside `tryBuildSystemEndpoints`; Azure Service Bus and SQS do the same thing, and they have the identical latent bug (a system endpoint that never receives the endpoint policies, declared pre-policy by `BrokerResource.Setup()` and redeclared post-policy at start up). The version already on main fixes that whole class for every broker transport. **This PR would narrow the fix back to RabbitMQ and re-expose the others.**

Unless there's a reason to prefer the smaller blast radius, main is in the better shape and this should be closed.

Issue #3888, opened 2026-08-10
Agents lost when a node dies are not picked up by a healthy peer that advertises the same capability


erdtsieck:
> **Edited after the fact.** This originally blamed memory pressure on one node and proposed raising a limit. That was wrong; the cause is JasperFx/jasperfx#644. The gap described here is real and separate, and the issue now states it without the mistaken framing. Original diagnosis preserved in the comments.

## What happened

During a blue/green projection rollout on 512 shard databases, two fleets of the same application ran side by side: one serving `provided-cares:V23`, one rebuilding `V24`. `invoices:V16` was **not** bumped, so both fleets advertise it and the leader split those agents between them.

One fleet's process died (JasperFx/jasperfx#644 — the per-tenant high-water poller OOMs and takes the host down, exit 139). It was holding **53 agents of that shared projection version**.

The other fleet runs the same version, advertises the same capability, and was completely healthy. It never received them.

The serving fleet's own backlog had been falling steadily — 102 → 79 → 53 shards — and then sat at **exactly 53 for sixteen minutes**, unchanged, while the disk went idle at 1% IOPS. Those are customer-facing read models, frozen, with a capable node beside them. A `kubectl rollout restart` was the only thing that moved them.

## Why nothing recovers it

`EventSubscriptionAgent` detects the stall and, because `ShardFailureCategory.Other` is self-healing, attempts recovery:

```csharp
// EventSubscriptionAgent.AttemptAutoRestartAsync
await StopAsync(cancellationToken);
await StartAsync(cancellationToken);
```

That restart is **node-local**. Nothing hands the agent to a different node.

For a per-event failure that is exactly right — a poison event reproduces anywhere, and GH-3638's reasoning about not churning applies. But `Other` is by definition *not* bound to the event; `canSelfHeal`'s own doc comment calls it "a transient infrastructure fault … a database blip, a timeout". Those are frequently properties of the *node* or its process, not of the shard. Retrying in the same place re-runs the same conditions.

And when the process itself dies there is no local retry at all — the agents simply stop existing on that node until something restarts the pod.

## Suggested

After a run of failed local restarts of a self-healing failure, release the assignment so the leader can place the agent on another capable node. `RemoveAssignmentAsync` already exists in the stop path; what is missing is the policy that reaches for it — including whatever keeps the leader from handing the agent straight back to the node that just failed it. At 13,572 agents a policy that bounces would be worse than the freeze it replaces, which is why I have not guessed at one.

## Related

[...]

Issue #3638, opened 2026-07-26
WO-8: surface a paused shard's ShardFailure to the assignment plane and health checks


jeremydmiller:
### Summary

WO-8, the Wolverine half of JasperFx/jasperfx#565 (implemented in JasperFx/jasperfx#567). Surface an `ApplyEventException`-stopped shard to the assignment/observer plane **with a reason**, so CritterWatch can alert on it and progress doesn't silently flatline. This was deliberately deferred to ship as a coordinated pair with the JasperFx.Events accessor, which has now landed.

### What JasperFx.Events gives us

- `ISubscriptionAgent.Failure` → `ShardFailure?` — a default interface member (non-breaking), set alongside `Status` when a shard pauses or stops and cleared on start/replay.
- `ShardFailure` carries `Category` (`ApplyEvent` / `EventSerialization` / `UnknownEventType` / `ProgressionOutOfOrder` / `Other`), `ExceptionType`, `RootExceptionType`, `Message`, `Detail` (the full text `ShardState.PauseReason` has always carried), `Event` (sequence, event type, stream, tenant, version when known) and `OccurredAt`. It is a plain serializable record, not an `Exception`, so it survives the hop to the assignment plane and into a UI.
- `ShardStateTracker.CurrentState(...)` / `TryGetCurrentState(...)` / `CurrentStates()` — a synchronous snapshot for pollers that aren't subscribed.
- `ShardState.Failure` rides along on the published paused/stopped state, and Marten/Polecat persist the classified fields onto the extended progression row (JasperFx/marten#5048, JasperFx/polecat#368) so it is readable even when the publishing node is down.

### Proposed work in Wolverine

**1. Delegate the reason on the wrapper.** `EventSubscriptionAgent.Status` already delegates to the live inner agent (GH-3519); do the same for the reason:

```csharp
public ShardFailure? Failure => _innerAgent?.Failure;
```

Expose it on `IEventSubscriptionAgent` so CritterWatch can read it through the agent surface it already resolves via `IEventSubscriptionAgentFamily.FindAgentUriAsync`.

**2. Make the health check say something actionable.** `EventSubscriptionAgent.CheckHealthAsync` currently returns a fixed string:

```csharp
if (Status == AgentStatus.Paused)
    return HealthCheckResult.Unhealthy($"Projection {Uri} paused due to errors");
```

With the failure in hand that becomes the category, the failing event sequence and the root exception type — the difference between an alert an operator can act on and one they have to go dig for.

[...]

PR #3890, opened 2026-08-10
Make the stalled-agent auto-restart path testable, and cover it

erdtsieck:
Groundwork for #3888, split out so the policy discussion there isn't tangled up with a testability change.

## The problem

`EventSubscriptionAgent.CheckHealthAsync` has an auto-restart branch that nothing covers, and nothing *could* cover: reaching it takes `MaxConsecutiveStallsBeforeRestart` (3) stalls at `StallTimeout` (60s) apart, and the clock was `DateTimeOffset.UtcNow` read inline. The existing agent tests all stop at the `Degraded` branch just short of it.

That is also why the GH-3638 distinction — retry an infrastructure blip, surface a poison event — has never been asserted anywhere.

## The change

A `TimeProvider` property defaulting to `TimeProvider.System`, and the three inline clock reads routed through it.

One subtlety worth flagging: `_lastAdvancedAt` moved out of its field initialiser into a first-observation anchor inside the health check. Initialising it at construction takes the value from the system clock, so a substituted clock afterwards compares against a baseline on a different timeline and every check reads as stalled since year one. Anchoring on first use means whichever clock is in place sets the baseline.

## The tests

Three passing:

- an `Other`-category failure is retried — that is what auto-restart exists for, per `canSelfHeal`'s own doc comment
- an `ApplyEvent` failure is surfaced with "Auto-restart suppressed" instead of restart-looped (GH-3638)
- an agent that is keeping up never reaches the branch at all — the counterweight, so the first two cannot pass against a detector that fires unconditionally

One skipped, stating what #3888 asks for: after a run of failed local restarts the agent should stop being retried on the node that keeps failing it. The comment carries the measurement — a warming fleet held 53 agents of a shared projection version, went down on memory, and every retry landed on the same starved node while a healthy peer advertising the same capability stood idle. Those read models were frozen for sixteen minutes.

I have deliberately not implemented that policy. The mechanism is easy — `RemoveAssignmentAsync` is already in the stop path — but what keeps the leader from handing the agent straight back is a real design decision about the assignment grid, and I would be guessing at which shape you want. Happy to build it once you say.

`dotnet test --filter FullyQualifiedName~Runtime.Agents`: 324 passed, 1 skipped, 0 failed.

### DIFF
--- a/src/Testing/CoreTests/Runtime/Agents/stalled_agent_auto_restart.cs
+++ b/src/Testing/CoreTests/Runtime/Agents/stalled_agent_auto_restart.cs
@@ -1,0 +1,160 @@
+using JasperFx;
+using JasperFx.Events.Daemon;
+using JasperFx.Events.Projections;
+using Microsoft.Extensions.Diagnostics.HealthChecks;
+using NSubstitute;
+using Shouldly;
+using Wolverine.Runtime.Agents;
+using Xunit;
+using IProjectionDaemon = JasperFx.Events.Daemon.IProjectionDaemon;
+using JasperFxSubscriptionAgent = JasperFx.Events.Daemon.ISubscriptionAgent;
+
+namespace CoreTests.Runtime.Agents;
+
+/// <summary>
+/// Coverage for the auto-restart branch of <see cref="EventSubscriptionAgent.CheckHealthAsync" />, which
+/// until now had none: reaching it takes three consecutive stalls sixty seconds apart, and the clock was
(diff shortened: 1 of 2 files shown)


erdtsieck:
Context that arrived after opening this: the crash behind #3888 turned out to be JasperFx/jasperfx#644 — the per-tenant high-water poller OOMing the host process, not the projection work and not Wolverine.

That does not change this PR. The auto-restart branch still has no coverage and still cannot be reached without a settable clock, and #3888's reassignment question is if anything sharper now: a process that dies in the poller takes every projection agent on that node with it, so "can a healthy peer pick these up" is the difference between a pod restart and sixteen minutes of frozen read models.

Issue #3899, opened 2026-08-10
Partitioned listeners: allow per-message-type slot-key derivation (or exempting types from partitioning)


jeremydmiller:
Filed on Jeremy's behalf, out of the JasperFx/CritterWatch#969 throughput investigation.

**The problem:** `PartitionProcessingByGroupId` routes **every** message on the listener through the GroupId-keyed slot. That's correct for message types that need per-group ordering — but when one GroupId dominates, the whole listener collapses to parallelism 1 for *all* message types, including the ones that never needed ordering.

**Field arithmetic (CritterWatch#969):** CritterWatch keys GroupId = ServiceName to serialize per-service stream appends. A customer fleet has **one** monitored service and 2,173 tenants — so effectively every envelope lands on one slot. Measured: ~207 msg/min drained against ~7,870/min arriving at effective parallelism ~2 on an 8-core pod, listener permanently latched, queue +8k/min from a purged start. The messages doing the drowning (metrics samples, progression updates, DLQ counts) are all document/accumulator paths that need **no ordering at all** — they're only serialized because they share the listener with the few types that do.

**Ask — a way to scope partitioning to the types that need it.** Either shape works:
1. **Per-message-type slot-key derivation:** let the endpoint declare, per message type, how to derive the slot key (e.g. metrics keyed by service+store rather than service — spreading the dominant group across slots while keeping each sub-key ordered); or
2. **A per-type bypass:** let listed message types skip the partitioned block entirely and run at the endpoint's plain parallelism.

Related but distinct from #2580 (which asked for *stricter* same-group serialization) and #3481 (broker-native partitioning) — this is the opposite lever: *relaxing* partitioning for types that never asked for it.

**Landing:** the GroupId partitioning configuration surface + `ShardedExecutionBlock`/slot-routing path.

**Effort:** M.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #3904, opened 2026-08-10
GH-3900: parallelize the sharded execution block's deserialize stage without giving up per-group FIFO

### LINKED ISSUE #3900: ShardedExecutionBlock's deserialize stage is single-threaded
Filed on Jeremy's behalf, out of the JasperFx/CritterWatch#969 throughput investigation.

**The bottleneck:** `ShardedExecutionBlock.DeserializeFirst` builds its upstream deserialize stage with `PushUpstream` **without a `parallelCount`**, which produces a 1-worker block (the JasperFx Blocks default). So every envelope's decompress + deserialize runs single-threaded *in front of all slots* — regardless of how many slots the sharded block fans out to. On a listener carrying compressed telemetry payloads (CritterWatch's Brotli/MessagePack pushes), the serialized deserialize stage caps throughput before the slots ever see work.

In the CritterWatch#969 profile this stacks with the slot-collapse problem (one dominant GroupId): even after partitioning is fixed, an 8-core pod still funnels all decompress+deserialize through one worker.

**The fix is nearly wired already:** the `PushUpstream(parallelCount, ...)` overload exists — the deserialize stage just doesn't use it. Wire a sensible degree (e.g. min(slot count, ProcessorCount), or the endpoint's execution parallelism). Deserialization is stateless per envelope, so parallelizing it carries no ordering hazard — ordering is enforced downstream at the slots.

**Landing:** `ShardedExecutionBlock.DeserializeFirst` (`src/Wolverine/Runtime/WorkerQueues/` or current home) + a throughput-shaped test or at least a pin that the block is constructed with parallelCount > 1.

**Effort:** S–M.

jeremydmiller:
Closes #3900. Field context: JasperFx/CritterWatch#969 — on a listener carrying compressed telemetry payloads, the single-threaded deserialize stage caps ingest throughput in front of ALL execution slots, regardless of slot count.

## The ordering analysis (why this was never a one-liner)

`ShardedExecutionBlock.DeserializeFirst` built its upstream stage with the parameterless `PushUpstream`, i.e. a 1-worker `Block<Envelope>`. That single worker deserializes and then `await`s `PostAsync` inline, which means **envelopes reach the slot enqueue — where the GroupId → slot hash runs — in exact arrival order**. Since every slot is a single-worker block and all envelopes of a group hash to the same slot, arrival-ordered slot enqueue is precisely what yields the per-GroupId FIFO the sharded structure exists to provide.

Two tempting fixes are unsound:

1. **The issue's literal suggestion, `PushUpstream(parallelCount, ...)`** — a JasperFx `Block<T>` with `parallelCount = N` runs N independent `Task.Run` reader loops over one shared channel. Completion order (and therefore downstream `PostAsync` order) is unordered. Two same-group envelopes arriving back-to-back could swap before the slot hash ever sees them — a silent break of the ordering guarantee, exposed only under load. The issue's "no ordering hazard" claim is wrong: ordering is enforced *by* arrival-ordered slot enqueue, not downstream of it.

2. **Hash-partitioning the deserialize stage by group** (N serial deserialize lanes) — the `GroupId` wire header is only present when the *sender* set it. Receiver-side grouping rules (`ByMessage`, `ByPropertyNamed` — which NREs on a null `Message` — inferred grouping, custom `IGroupingRule`s) can only resolve a group **after** deserialization, so the lane key doesn't exist yet for exactly those workloads. It also breaks mixed header/no-header same-group traffic during rolling upgrades.

## The design: pipelined deserialization with ordered emission

- A **single-worker ingress** stage starts up to N concurrent deserialization tasks (gated by a `SemaphoreSlim`, N = `min(slotCount, ProcessorCount)` by default) and posts `(envelope, task)` pairs downstream **in arrival order**.
- A **single-worker emit** stage awaits each task **in that same order** and hands the envelope to the slot hash. Awaiting in order *is* the re-sequencing — an out-of-order completion simply waits its turn.

Slot-enqueue order is byte-for-byte what the old serial stage produced — for every case, including groups resolvable only post-deserialize — while the CPU-bound decompress/deserialize work runs N-wide. Throughput becomes pipelined (~N× when deserialize is the bottleneck) at the cost of head-of-line latency on a slow envelope, which is inherent to preserving order. No JasperFx Blocks change required; this composes from the existing `PushUpstream` primitives against JasperFx 2.46.0.

Failure-path behavior is preserved: a deserialization failure surfaces on the emit await and is routed to the sharded block's `OnError` (the receivers' logging) with the actual envelope; a terminal block fault (jasperfx#506) still reports with a null item so `HasFaulted` detection in the receivers (CritterWatch#942/#922) keeps working. `parallelism <= 1` (e.g. a 1-core container) falls back to the original serial stage.

## Tests (`sharded_execution_parallel_deserialization`)

- **Ordering stress, header case**: 10 groups × 200 interleaved envelopes, jittered deserialize cost, parallelism 8 — per-group sequences must come out exactly in posted order.
- **Ordering stress, rules case**: same, but no GroupId on the wire; the group is resolvable only after deserialization via `ByMessage` — the case that forbids lane-hashing.
- **Single-group traffic**: strictly sequential, ordered, and never executed concurrently (max observed execution concurrency == 1).
- **Parallelism pin**: fixed 20 ms per-envelope deserialize cost; asserts observed deserialize concurrency > 1 and a coarse wall-time bound well under the serial floor (not a benchmark).
- **Default overload + serial fallback** covered.

All 6 pass on net9.0 and net10.0, along with the full `CoreTests.Runtime.Partitioning` suite (90/90 both TFMs). The Postgres-backed `SlowTests.Bug_2580_tenant_partitioning_with_inferred_grouping` — end-to-end through the sharded receiver with a rules-resolved group — also passes. `SlowTests.Bug_concurrency_with_global_partitioning` failed locally only because it needs a live Kafka broker at localhost:9092 (died in `BrokerTransport.InitializeAsync`/`KafkaTopic.SetupAsync` at startup, before any envelope reaches the deserialize stage) — environmental, not related to this change.

### DIFF
--- a/src/Testing/CoreTests/Runtime/Partitioning/sharded_execution_parallel_deserialization.cs
+++ b/src/Testing/CoreTests/Runtime/Partitioning/sharded_execution_parallel_deserialization.cs
@@ -1,0 +1,299 @@
+using System.Collections.Concurrent;
+using Wolverine.ComplianceTests;
+using Wolverine.Runtime;
+using Wolverine.Runtime.Partitioning;
+using Xunit;
+
+namespace CoreTests.Runtime.Partitioning;
+
+/// <summary>
+/// GH-3900: the decompress/deserialize stage in front of the sharded execution slots runs
+/// N-wide instead of single-threaded. These tests pin the two things that matter about that
+/// change: (1) per-GroupId FIFO ordering into the slots is preserved byte-for-byte -- including
+/// when the group can only be resolved AFTER deserialization by a message-based grouping rule --
+/// and (2) the deserialization work itself actually runs concurrently.
+/// </summary>
+public class sharded_execution_parallel_deserialization
+{
+    // Simulates the real HandlerPipeline.TryDeserializeEnvelope: the envelope's Message is null
+    // until this stage runs. An optional per-envelope delay stands in for decompress/parse cost
+    // and, when jittered, forces out-of-order completions that would expose any unordered fan-out
+    private class StubDeserializingPipeline : IHandlerPipeline
+    {
+        private readonly ConcurrentDictionary<Guid, object> _payloads = new();
+        private readonly Func<int> _delayInMilliseconds;
+        private int _active;
+        private int _maxActive;
+
+        public StubDeserializingPipeline(Func<int>? delayInMilliseconds = null)
+        {
+            _delayInMilliseconds = delayInMilliseconds ?? (() => 0);
+        }
+
+        public int MaxObservedConcurrency => Volatile.Read(ref _maxActive);
+
+        public Envelope EnvelopeFor(object message, string? groupId = null)
+        {
+            var envelope = ObjectMother.Envelope();
+            envelope.GroupId = groupId;
+            _payloads[envelope.Id] = message;
+            return envelope;
+        }
+
+        public Task InvokeAsync(Envelope envelope, Wolverine.Transports.IChannelCallback channel)
+            => throw new NotSupportedException();
+
+        public Task InvokeAsync(Envelope envelope, Wolverine.Transports.IChannelCallback channel,
+            System.Diagnostics.Activity activity)
+            => throw new NotSupportedException();
+
+        public async ValueTask<IContinuation> TryDeserializeEnvelope(Envelope envelope)
+        {
+            var current = Interlocked.Increment(ref _active);
+            InterlockedMax(ref _maxActive, current);
+
+            try
+            {
+                var delay = _delayInMilliseconds();
+                if (delay > 0)
+                {
+                    await Task.Delay(delay);
+                }
+                else
+                {
+                    // Still hop threads so completions can interleave
+                    await Task.Yield();
+                }
+
+                envelope.Message = _payloads[envelope.Id];
+                return NullContinuation.Instance;
+            }
+            finally
+            {
+                Interlocked.Decrement(ref _active);
+            }
+        }
+
+        private static void InterlockedMax(ref int location, int value)
+        {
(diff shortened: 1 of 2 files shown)


jeremydmiller:
Merged `origin/main` (#3902 exempt lane, #3903 batch DeliverBy) into this branch — conflict resolved in `ShardedExecutionBlock.cs`.

**Interaction design (GH-3899 x GH-3900):** exempted envelopes ride the same ordered deserialize pipeline, because exemption is keyed on the message *type*, which — exactly like the group — is only knowable after deserialization; there is no earlier point to divert them. They are deserialized exactly once, and the exempt-lane routing happens inside `PostAsync`, downstream of the ordered emission. Posting to the multi-worker exempt lane never blocks emission any harder than posting to a slot does (same bounded-channel back pressure), and exempt types have no ordering requirement to begin with, so riding the ordered emission is harmless.

Pinned by a new test, `exempt_types_ride_the_ordered_pipeline_once_and_still_reach_the_parallel_exempt_lane`: a barrier that deadlocks unless two exempt executions overlap (proving the exempt lane still runs parallel behind the ordered stage), per-group FIFO intact for interleaved grouped traffic, and exactly one deserialization per envelope.

Post-merge results: full `CoreTests.Runtime.Partitioning` — including all of #3902's `exempting_types_from_partitioned_processing` tests — **103/103 on net9.0 and net10.0**; #3903's `BatchDeliverByExpiry`/`batch_deliver_by_expiry` tests 13/13 on both TFMs.

Issue #3907, opened 2026-08-11
Tracking: implement the aggregate handler workflow once in Wolverine core instead of once per store


jeremydmiller:
Wolverine ships two near-identical copies of the aggregate handler workflow — one in `Wolverine.Marten`, one in `Wolverine.Polecat`. This issue tracks pulling the shared design down into Wolverine's core library so it is implemented **once**, with each store integration keeping its own public vocabulary and delegating to the common code.

The immediate forcing function is a third store: [Fisher](https://github.com/JasperFx/fisher), the SQLite-backed sibling. Writing `Wolverine.Fisher` as a third copy-paste of ~7,000 lines is the outcome this issue exists to prevent.

## The evidence

Normalizing names (`Marten`→`Polecat`, `IDocumentOperations`→`IDocumentSession`) and diffing every paired file between the two integrations:

| | |
|---|---|
| `Wolverine.Marten` | 8,649 LOC / 68 files |
| `Wolverine.Polecat` | 7,000 LOC / 66 files |
| Byte-identical after normalization | **9 files** |
| Under 20% divergence | ~30 more |

The byte-identical set includes `Codegen/BoundaryEventCapture.cs` (118 lines), `Codegen/LoadBoundaryFrame.cs`, `Codegen/RegisterEventsFrame.cs`, `Events.cs`, `Persistence/Sagas/DocumentSessionOperationFrame.cs`, `UnknownAggregateException.cs`, `ConcurrencyStyle.cs`.

**The drift is the real cost, not the line count.** In `AggregateHandling.cs` — the heart of the workflow — the two copies have diverged asymmetrically:

- the Polecat copy has proper nullable annotations (`required`, `[NotNullWhen(true)]`) that the Marten copy lacks;
- the Marten copy carries a null guard on `AggregateId`/`AggregateType` in `Modify` that the Polecat copy dropped.

Improvements and fixes are landing on one copy at a time. That compounds with every store added.

Genuinely store-specific and staying put: the op library (`IMartenOp.cs`, 850 lines), the DI registration surface (`WolverineOptionsMartenExtensions.cs`), session creation, `MartenIntegration.cs`.

## Scope and constraints

[...]

PR #3914, opened 2026-08-12
GH-3907 increment 2: promote the aggregate workflow attributes into core as [WriteModel]/[ReadModel]/[DeciderFunction]

jeremydmiller:
Increment 2 of #3907. Increment 1 (#3910) moved the first shared mechanism into core and created `IEventSourcingFrameProvider`; this promotes the **attributes** onto it and reduces the per-store ones to shells.

Naming is per Jeremy's call:

| new, in core | the store attributes that now inherit it |
|---|---|
| **`[WriteModel]`** | `[WriteAggregate]` (Marten, Polecat) |
| **`[ReadModel]`** | `[ReadAggregate]` (Marten, Polecat) |
| **`[DeciderFunction]`** | `[AggregateHandler]` (Marten, Polecat) |

`[DeciderFunction]` is a deliberate pull toward the functional decider vocabulary — `decide(command, state) -> events`.

**Additive.** `[WriteAggregate]`, `[ReadAggregate]` and `[AggregateHandler]` keep compiling and behaving identically.

Net **+2,109 / −1,573** across 46 files. The two integrations alone lose **1,642 lines**.

## What had to come down with the attributes

Increment 1's seam was two members (`StoreName`, `EventsCollectionType`) plus an extension method — enough for event capture, not enough for the attributes, which are a thin front on `AggregateHandling`. That file was still duplicated at **398 lines in each store**, so `AggregateHandling`, `MissingAggregateCheckFrame`, `TagAggregateOtelFrame` and the concurrency enum came down too. That is the bulk of the deletion.

The seam grew additively to seven members (two optional): `UnknownAggregateExceptionType`, `BuildLoadAggregateFrame(AggregateLoadRequest)`, `BuildFetchLatestFrame`, `DetermineStreamIdentity`, and optional `TryDetermineNaturalKeyType`.

The two frame-builders are the load-bearing ones. **No JasperFx session abstraction declares the `.Events` accessor**, so core would otherwise have to write `session.Events.FetchForWriting<T>(…)` itself and name a store. Handing back a finished `Frame` also keeps Marten's `IBatchableFrame` enlistment — which Polecat has no equivalent of — private to Marten.

The seam is now **public** and lives on `MartenPersistenceFrameProvider` / `PolecatPersistenceFrameProvider` rather than the standalone classes increment 1 created, so it is discoverable through the persistence strategies already on `GenerationRules` with no second registry.

One thing it did **not** need: each store's `IPersistenceFrameProvider.DetermineSagaIdType` turned out to be byte-for-byte the identity resolution its `WriteAggregateAttribute` had inlined. That seam already existed.

## Two reasons the shells are not literally empty

Both are deliberate and worth review.

**1. `ConcurrencyStyle` is public vocabulary in each integration's namespace.** `[WriteAggregate(LoadStyle = ConcurrencyStyle.Exclusive)]` and `[AggregateHandler(ConcurrencyStyle.Exclusive)]` resolve it to the store's own spelling, so retiring those enums would be a **source break**. Each store keeps its enum and bridges: `[WriteAggregate]` shadows `LoadStyle`, `[AggregateHandler]` forwards through its constructor.

This also forced a rename in core. `ConcurrencyStyle` there **failed the `CoreTypeNameCollisionCompliance` test increment 1 installed** — decision 2 names `ConcurrencyStyle` in its own doc comment — so core's is `ModelConcurrencyStyle`.

**2. A store attribute must name its own provider — and this one is a genuine finding.**

`AddMarten(...)` **without** `IntegrateWithWolverine()` is a supported configuration (`Bug_505`, `Bug_2182` and eight others depend on it), and in it **nothing ever registers a persistence strategy**. `[Aggregate]` / `[WriteAggregate]` have always worked there because they named `MartenPersistenceFrameProvider` directly. Resolving purely through the registry silently narrowed that — it surfaced as **10 red `Wolverine.Http.Tests`**.

So the three core attributes expose `protected virtual ResolveEventSourcingProvider(...)`, defaulting to the registry, and each shell overrides it in one line. That is the honest division: **the agnostic attribute asks who owns the model; the store attribute already knows.**

Final shells: `[ReadAggregate]` 34 lines, `[AggregateHandler]` 41, `[WriteAggregate]` 52 — identical in both integrations. Everything that was *workflow* is gone from all six.

## A real bug fixed on the way

Polecat's `[Identity]` does not derive from `JasperFx.IdentityAttribute`. Carrying the pair over the seam as a list works on the codegen path and **silently fails on `IMayInferMessageIdentity`**, where `HandlerChain` reflects fresh attribute instances and has no `GenerationRules` — so Polecat's own `[Identity]` was being lost on the `TryInferMessageIdentity` path. Core now matches on the attribute's **name**, which is call-site-independent, picks up a third store's spelling for free, and is one member smaller.

## A third store can plug in with no core change

This was an explicit design constraint, because **Fisher** (SQLite-backed, unreleased) is intended as a third persistence option and has no Wolverine integration yet. Getting this in first means it never needs its own copies of these attributes.

`Wolverine.Fisher` would implement `IEventSourcingFrameProvider` on its `IPersistenceFrameProvider`, register with `AddPersistenceStrategy<T>()`, and derive its shells. Nothing in core enumerates stores: the owning store resolves from the same registry `[Entity]` uses, `[Identity]` matches by name, `EventsCollectionType` defaults to null (made optional precisely so a store need not invent an `Events` type), and `TryDetermineNaturalKeyType` defaults to null.

**Caveat:** two catch-all event stores in one app is unresolvable by the registry — but that configuration is already unsupported (they would fight over sagas, entities and transaction middleware), and the per-store overrides mean the store-named attributes are immune regardless.

## Docs

`docs/guide/handlers/persistence.md` gains a five-row vocabulary table — `[Entity]` / `[WriteModel]` / `[ReadModel]` / `[DeciderFunction]` / `Storage.*` — and an **Event Sourced Models** section, so the features read as one vocabulary rather than four unrelated ones. Samples are compiled `#region`s in `src/Samples/DocumentationSamples/EventSourcedModelSamples.cs` per the mdsnippets convention. Both stores' "Aggregate Handlers and Event Sourcing" pages get a pointer to the agnostic form plus an explicit note that everything on those pages still applies.

## Verification

- Full solution (**213 projects**) builds clean. `Wolverine.csproj` still **0 IL warnings** under `EnableTrimAnalyzer` / `EnableAotAnalyzer` — the moved code needed ~20 leaf `[UnconditionalSuppressMessage]` annotations in the repo's established style.
- **MartenTests 566/566** · **PolecatTests 262 passed / 2 pre-existing skips** · **Wolverine.Http.Tests 951 passed / 10 pre-existing skips / 0 failed**
- **20 new tests**, a deliberately near-identical pair at `src/Persistence/{MartenTests,PolecatTests}/AggregateHandlerWorkflow/agnostic_model_attributes.cs` (10 each). Four behavioural per store, including `read_model_resolves_the_current_state_without_appending`, which asserts the stream version is unchanged — `FetchLatest` vs `FetchForWriting` is exactly the distinction that would rot silently. Six pin the shells, including the `ConcurrencyStyle` bridge in both directions: an unforwarded setter would downgrade every exclusive lock to optimistic and nothing else would notice.

Rebased onto `main` so the 6.26.0 version bump from the original base branch is not carried in; core and both integrations re-built clean after the rebase.

## Known loose end

`src/Persistence/PolecatTests/MARTEN_VS_POLECAT_REPORT.md` still describes the duplicated `AggregateHandling` and is now stale. Left alone as a tracking doc rather than edited blind.

## Context

Driven by [CritterWatch#999](https://github.com/JasperFx/CritterWatch/issues/999), where the same source compiles into a Marten flavour and a Polecat flavour rebound by a per-flavour `GlobalUsings.cs` aliasing seven types — four of them Wolverine integration types. This retires two (`AggregateHandlerAttribute`, `Events`); `Storage.*` already retires the other two with no upstream work.


jeremydmiller:
**Deliberately held as a draft: this does not merge until it is proven out in a real consumer.**

CritterWatch is that consumer. It will consume this branch via a **local NuGet pack** and convert its own handlers to `[WriteModel]` / `[ReadModel]` / `[DeciderFunction]`, and only once that adoption is real does this merge.

The reasoning: this PR is judged green by its own tests, which is the weakest kind of evidence for an API change. The two judgement calls in it — each store keeping its `ConcurrencyStyle` as a bridge rather than taking a source break, and shells overriding `ResolveEventSourcingProvider` to name their own provider — are exactly the sort of thing that reads fine in a test fixture and turns out wrong against a codebase that did not write the tests.

So adoption feedback is **input to this PR, not a downstream problem**. If the attributes do not fit CritterWatch's real handler shapes, the right outcome is changing this branch rather than contorting the consumer around it.

Concretely, what CritterWatch is proving:

- that `[WriteModel]` / `[DeciderFunction]` genuinely remove the need for an injected session in an event-sourced handler — the point of the whole exercise, since that is what lets `CritterWatch.Services` stop aliasing store types
- that the attributes behave identically across **both** flavours, since the same source compiles into a Marten build and a Polecat/SQL Server build
- that they compose with `[Transactional]` ancillary-store routing, where removing an injected session can remove the very thing that made the chain transactional

Tracking on the consumer side: [CritterWatch#999](https://github.com/JasperFx/CritterWatch/issues/999).

Issue #3918, opened 2026-08-12
[WriteModel] ignores [Identity] while [DeciderFunction] honours it


jeremydmiller:
Found while adopting the GH-3907 attributes in a real consumer (CritterWatch), the proving ground for #3914.

## What happens

The two forms resolve the model's identity differently:

- **`[DeciderFunction]`** goes through `AggregateHandling.DetermineAggregateIdMember`, which **honours `[Identity]`** on the command property. So `UpdateLabelName([property: Identity] string ServiceName, ...)` resolves with no annotation at the call site.
- **`[WriteModel]`** uses `FindIdentity`, which tries only `RouteOrParameterName`, `{aggregateType}Id`, `id`, and then strong-typed-id matching. It **ignores `[Identity]` entirely.**

So the same command, against the same model, needs an explicit `[WriteModel("ServiceName")]` where `[DeciderFunction]` needed nothing.

## Why it matters

`[Identity]` is the declared, discoverable way to say "this property is the identity" — it lives on the message, where it is true regardless of which handler form consumes it. Having one attribute respect it and its sibling not is the kind of asymmetry that only shows up when you migrate between the two forms, which is precisely the path #3914 encourages.

It also bit in a way worth noting: an earlier cut of #3914 carried the identity attribute pair over the seam as a *list*, which worked on the codegen path and silently failed on `IMayInferMessageIdentity` — where `HandlerChain` reflects fresh attribute instances and has no `GenerationRules`. That was fixed by matching on the attribute's **name** (`cd0c2abbc`), which is call-site independent and picks up a third store's spelling for free. The same name-based match is what `[WriteModel]`'s identity resolution wants.

## Suggested

Have `FindIdentity` consult `[Identity]` before falling back to the name conventions — ideally via the same name-based match `[DeciderFunction]` already uses, so a third store's `[Identity]` spelling works without core knowing about it.

Refs #3907, #3914.

jeremydmiller:
Fixed and shipped in #3920 (merged as `0f9682efd`).

`WriteModelAttribute.FindIdentity` now consults `[Identity]`, using the same **name-based** attribute match `[DeciderFunction]` uses (the one `cd0c2abbc` installed so a store's own `[Identity]` spelling works without core enumerating stores) — exactly as this issue suggested. Resolution order is now: explicit `[WriteModel("orderId")]`, then `[Identity]`, then `{Model}Id`, then `id`, then a strong-typed-id match.

[...]

Issue #3917, opened 2026-08-12
"Every return value is an event" should not be the default under [WriteModel]


jeremydmiller:
Found while adopting the GH-3907 attributes in a real consumer (CritterWatch), the proving ground for #3914.

## What happens

With no `IEnumerable<object>` in the return, `EventCaptureActionSource` treats **every** returned value as an event to append.

Under `[DeciderFunction]` that reads correctly — the method *is* the decider, so its return is its events.

Under `[WriteModel]` it does not. That attribute is now `AttributeTargets.Parameter` (#3914 + `0e9ba7403`), and it claims only *"this parameter is a model I write to."* It says nothing about the return. But the catch-all applies anyway.

## The concrete failure

CritterWatch's `BrokerHealthSnapshotsHandler` returns `Task<BrokerHealthChanged?>` — a **SignalR push**, meant for `EnqueueCascadingAsync`. Adding `[WriteModel]` to its model parameter would have **silently appended that push to the `ServiceSummary` event stream** instead of publishing it.

Nothing errors. Nothing warns. The push simply stops arriving and an unrelated event turns up in the stream.

Worked around with a `(IReadOnlyList<object>, TReply?)` tuple, which generates correctly (`AppendMany` + `EnqueueCascadingAsync`) — but that is a workaround for a default that pointed the wrong way.

## Why this is worth changing now

This is inherited behaviour, not new in #3914. But it is exactly the migration a `[WriteModel]` user will attempt — take a handler that returns a reply, add the attribute to its model parameter — and it fails **silently** rather than loudly. The attribute being parameter-scoped makes the mismatch sharper, not softer.

It is also why CritterWatch did **not** convert `AlertEndpoint.ClearAlert`: on an HTTP chain the same default would target the response body.

## Suggested

Scope the catch-all "every return is an event" default to **`[DeciderFunction]`**, and require an explicit opt-in under `[WriteModel]` — or at minimum refuse to compile when a `[WriteModel]` handler returns a type that is registered as a message, which is the unambiguous case.

Refs #3907, #3914.

jeremydmiller:
Closing as **by design**, per the determination that `[WriteModel]` must be identical to `[WriteAggregate]` in all respects.

The suggested fix — scoping the "every return value is an event" catch-all away from `[WriteModel]` — cannot be done without one of two unacceptable outcomes:

[...]

PR #3920, opened 2026-08-12
GH-3916/3918: [WriteModel] takes Required from the parameter's nullability and honors [Identity]

### LINKED ISSUE #3916: [WriteModel].Required defaults to true and ignores the parameter's nullability
Found while adopting the GH-3907 attributes in a real consumer (CritterWatch), which was the proving ground for #3914.

## What happens

`[WriteModel].Required` defaults to `true` and does **not** read the parameter's nullability annotation. So:

```csharp
public static async Task<...> Handle(
    SomeCommand command,
    [WriteModel("ServiceName")] ServiceSummary? service)   // nullable — "I will handle absence"
```

still generates an `EntityIsNotNullGuard<ServiceSummary>.Assert(...)` → `HandlerContinuation.Stop`, **plus a logged warning per message**. The handler's own null branch becomes dead code, and the author gets no signal — it compiles clean.

A parameter annotated nullable, with `Required = true`, is a contradiction. Today it resolves silently in favour of the attribute default.

## Why it mattered downstream

CritterWatch had to write `Required = false` on **all eight** converted handlers. These deliberately drop a push for an unregistered service, and two must throw `ServiceNotYetRegisteredException` specifically so a scoped `RetryWithCooldown` → `Discard` policy sees it. The generated `Stop` short-circuits before the handler body, so the intended behaviour silently disappeared.

## Suggested

Either:
- **default `Required` from the nullable annotation** — `T?` ⇒ `Required = false`, `T` ⇒ `Required = true`; explicit `Required` still wins; or
- **throw at codegen** when an explicit `Required = true` meets a nullable-annotated parameter, since that combination cannot be satisfied.

The first is friendlier and matches what the annotation already communicates. Either beats silently generating a guard the author didn't ask for.

## Related

There is a second-order effect worth knowing while fixing this: with `Required = true` the Marten batch-query enlistment fires (`documentSession.CreateBatchQuery()` → `batchedQuery.Events.FetchForWriting(...)`); with `Required = false` it degrades to a plain `documentSession.Events.FetchForWriting(...)`. So every `Required = false` conversion also gives up the batched form — probably an unintended consequence of how the not-null guard rearranges frames, and worth a look in the same pass.

Refs #3907, #3914.

jeremydmiller:
Two asymmetries found while adopting the GH-3907 attributes in a real consumer (CritterWatch, the proving ground for #3914) — both the kind you only see when migrating between the two handler forms, which is precisely the path #3914 encourages.

## GH-3916 — `Required` ignored the nullable annotation

`Required` defaulted to an unconditional `true`, so

```csharp
Handle(Cmd command, [WriteModel] ServiceSummary? service)
```

still generated `EntityIsNotNullGuard` → `HandlerContinuation.Stop`, plus a logged warning per message. The handler's own null branch was dead code and the author got no signal — it compiles clean. CritterWatch had to write `Required = false` on all eight converted handlers.

A nullable annotation with `Required = true` is a contradiction that used to resolve silently in favour of the attribute default. Now the annotation *is* the default, and an explicit `Required` still wins over it.

## GH-3918 — `[WriteModel]` ignored `[Identity]`

`[DeciderFunction]` honours `[Identity]` on the command member; `[WriteModel]` did not, so the same command against the same model needed an explicit `[WriteModel("ServiceName")]` under one form and nothing under the other.

`FindIdentity` now consults it, using `AggregateHandling`'s existing **name-based** match — the one `cd0c2abbc` installed so a store's own `[Identity]` spelling works without core enumerating stores. Slotted behind an explicit `[WriteModel("...")]`, ahead of the name conventions.

## GH-3917 is deliberately *not* here

Its suggestion — scope the "every return value is an event" catch-all away from `[WriteModel]` — cannot be done without either changing `[WriteAggregate]` too (it derives from `[WriteModel]`) or making the two diverge. **`[WriteModel]` must be identical to `[WriteAggregate]` in all respects**, so the catch-all stays exactly as it is and the trap is documented instead: the Event Sourced Models section now states outright that a returned value which is also a message lands in the stream rather than being published, and shows the tuple return that does both. #3917 should be closed as by-design.

## Bonus: the batching note in #3916 points the other way

#3916 observed `Required = true` producing a Marten batch query and `Required = false` producing a plain `FetchForWriting`, and read that as `Required = false` **giving up** batching. It is the reverse.

`MartenBatchingPolicy.sortThroughFrames` counted a required aggregate load **twice** — once as the bare `LoadAggregateFrame`, once as the `Creator` of the `LoadEntityFrameBlock` wrapping it for the guard — so a handler with a single aggregate cleared the "two or more operations" gate *on a duplicate* and got `CreateBatchQuery()` + `Execute()` around one lone `FetchForWriting`. `MartenBatchFrame.Enlist` already deduped, so only the gate was fooled and the batch always held exactly one item.

Counting each operation once means a single aggregate load now generates the same plain fetch whether or not it is required. Genuine multi-operation batching is untouched.

## Verification

Red first, deliberately: with the fix stashed, **2 of the 4 new tests fail and the 2 that pin unchanged behaviour pass** — which is the split you want from a change that is supposed to alter only the nullable and `[Identity]` paths.

- 105/105 `MartenTests.AggregateHandlerWorkflow`
- 7/7 `batch_querying_support` + `fetch_specifications_tests`
- `dotnet build wolverine.slnx -c Release -f net9.0` clean

Refs #3907, #3914. Closes #3916, #3918.


jeremydmiller:
Adding to the verification list in the description: **950/950 `Wolverine.Http.Tests` pass** (10 skipped, 960 total). That is the suite worth naming here — it is where `[Aggregate]` / `[WriteAggregate]` get the heaviest workout, including nullable model parameters, so it is the most likely place the `Required`-from-nullability default would have shown up as a regression.

jeremydmiller:
**CIMarten caught a regression, and I have reverted the part that caused it.**

`Bug_2387_write_aggregate_throw_exception_codegen` — `[WriteAggregate]` with `OnMissing.ThrowException` — failed to generate:

```
CS0841: Cannot use local variable 'stream_entity' before it is declared
```

The diagnosis in the description was right about the *mechanism* and wrong about the *consequence*. `MartenBatchingPolicy` really did count a required aggregate load twice, and a single-aggregate handler really did clear the "two or more operations" gate on that duplicate. But the resulting batch-of-one is **not** merely wasteful — it is load-bearing.

`MartenBatchFrame.GenerateCode` calls `LoadAggregateFrame.GenerateCodeForBatchResolution`, which declares `var stream_entity = await ...BatchItem` **before** the guard frames run — the ordering `86d84899e` installed precisely so the not-found guard can reference the variable. Take the batch away and the declaration reverts to where the plain path emits it, which on the `ThrowException` path is *after* the guard.

So the accidental duplicate was masking a real frame-ordering bug, and removing it exposed rather than caused the problem. Fixing it properly means making the plain path hoist its own declaration — its own piece of work, and not what #3916 asked for. #3916 only flagged the batching difference as "worth a look in the same pass"; it has now been looked at and the finding is recorded in the revert commit for whoever picks it up.

**The #3916 and #3918 changes are untouched and green** — 105/105 `MartenTests.AggregateHandlerWorkflow`, `Bug_2387` 2/2, 950/950 `Wolverine.Http.Tests`.

Separately: the `build` job failure here is **not from this PR**. It is `NU1903` — a newly-published high-severity advisory against `SSH.NET` 2025.1.0, which `TreatWarningsAsErrors` turns into a build error. `main` is failing the same way, so it needs its own fix.

Issue #3001, opened 2026-06-01
Codegen: prime scoped IServiceProvider via IRequireScopingFrame so MessageContext and IDocumentSession aren't duplicated under service location


jeremydmiller:
## Context

Two recurring bugs in user code that share a root cause: when a dependency chain in a handler forces Wolverine to fall back to service location, the resolved instances of certain "should-be-singleton-per-message" types diverge from the ones the generated `MessageHandler` is already using.

### Problem 1 — Duplicated `MessageContext` / `IMessageBus` / `IMessageContext`

A handler's dependency takes `IMessageBus` (or `IMessageContext`) and is reached via service location. The scoped `IServiceProvider` resolves a **second** `MessageContext` independent of the one driving the handler. Writes through the service-located bus enrol with a *different* outbox (or none), so cascaded messages silently drop on the floor.

### Problem 2 — Duplicated Marten / Polecat `IDocumentSession` (and `IQuerySession`)

The Wolverine integration (`MartenIntegration` / `PolecatIntegration`) is responsible for building the `IDocumentSession` enrolled with the active outbox — that's what `MartenOutbox` / `PolecatOutbox` are constructed against. When a service-located dependency pulls `IDocumentSession` from the container, it bypasses that path and gets a **separate** session with its own (empty) outbox enrolment. Transaction boundaries are defeated.

### What's already in place (partial mitigation, runtime)

#2583 added an `AsyncLocal`-based fallback: `ServiceLocationAwareExecutor` sets `Wolverine.Runtime.MessageContext.Current` for the duration of each invocation, and the scoped factories in `HostBuilderExtensions.cs:216-226` prefer it:

```csharp
services.AddScoped<IMessageBus>(sp =>
    Wolverine.Runtime.MessageContext.Current ?? sp.GetRequiredService<MessageContext>());
services.AddScoped<IMessageContext>(sp =>
    Wolverine.Runtime.MessageContext.Current ?? sp.GetRequiredService<MessageContext>());
services.AddScoped<MessageContext>();
```

That fixes the `MessageContext` case at runtime via `AsyncLocal` flow, but:

- It's fragile under async work that escapes the original flow.
- It doesn't help the Marten/Polecat `IDocumentSession` case — those sessions still come from their own scoped registrations.
- It's not a *structural* fix at the codegen layer.

## Goal

[...]

PR #3962, opened 2026-08-16
GH-3956: recognise the JasperFx.Events.Documents contracts in all three stores

### LINKED ISSUE #3956: Persistence frame providers don't recognise JasperFx.Events.Documents — a handler taking IDocumentSessionOperations silently loses its writes
JasperFx 2.47.0 shipped `JasperFx.Events.Documents` (jasperfx#647), and all three stores already implement it on their own session types. Verified by reflecting on the pinned assemblies:

| | Marten 9.24.0 | Polecat 5.15.1 | Fisher 0.7.0 |
|---|---|---|---|
| `IQuerySession` | `IDocumentReadOperations` | same | same |
| `IDocumentSession` | `IDocumentReadOperations`, `IDocumentWriteOperations`, `IDocumentSessionOperations` | same | same |
| `IDocumentStore` | `IDocumentSessionFactory` | same | same |

So a handler *can* declare `JasperFx.Events.Documents.IDocumentSessionOperations` today and it will bind to the store's real session. **What it will not get is a commit.**

## The defect

`PersistenceFrameProvider.CanApply` decides whether a chain gets `CreateDocumentSessionFrame` + the `DocumentSessionSaveChanges` postprocessor by matching against a fixed type list. `MartenPersistenceFrameProvider.cs:111`:

```csharp
var serviceDependencies = chain
    .ServiceDependencies(container, new[] { typeof(IDocumentSession), typeof(IQuerySession),
        typeof(IDocumentOperations), typeof(global::JasperFx.Events.IEventOperations),
        typeof(global::JasperFx.Events.IEventStoreOperations),
        typeof(global::Marten.Events.IEventStoreOperations) }).ToArray();

return serviceDependencies.Any(x => x == typeof(IDocumentSession) || x == typeof(IDocumentOperations)
                                    || x.Closes(typeof(IEventStream<>))
                                    || x == typeof(global::JasperFx.Events.IEventOperations)
                                    || x == typeof(global::JasperFx.Events.IEventStoreOperations)
                                    || x == typeof(global::Marten.Events.IEventStoreOperations));
```

`JasperFx.Events.Documents.*` is in neither the probe list nor the predicate. A handler declaring the abstraction therefore compiles, resolves from DI, runs, stores into the session's unit of work — and the unit of work is never committed. No exception.

⚠️ **`chain.IsTransactional` is not a usable check for this.** `[Transactional]` sets it unconditionally while `GetPersistenceProviders` falls back to a no-op provider, so the flag reads `true` with no commit emitted. (We've also seen the mirror image — `false` with a commit present, on a chain whose writes are returned `IStorageAction<T>`.) The emitted frames are the only evidence in either direction, which makes this a genuinely silent failure mode.

## The precedent is in the same method

`JasperFx.Events.IEventOperations` and `IEventStoreOperations` — the store-agnostic **event** abstractions — are already in that list, and the comment immediately below explains why they were added:

> A handler that takes the event operations straight as a parameter ... is unambiguously using this store, but none of those types appeared here, so `AutoApplyTransactions` skipped the chain and nothing was ever committed. Appending through the parameter queued into the session's unit of work and then silently vanished, with no exception.

That is word-for-word the bug being reported here, on the document side rather than the event side.

## Identical in all three integrations

- `src/Persistence/Wolverine.Marten/Persistence/Sagas/MartenPersistenceFrameProvider.cs:111`
- `src/Persistence/Wolverine.Polecat/Persistence/Sagas/PolecatPersistenceFrameProvider.cs:93`
- `src/Persistence/Wolverine.Fisher/Persistence/Sagas/FisherPersistenceFrameProvider.cs:85`

## What's needed

1. **Add `JasperFx.Events.Documents.IDocumentSessionOperations` / `IDocumentReadOperations` / `IDocumentWriteOperations`** to the probe list and predicate in all three providers, for the same reason the event abstractions are already there.
2. **Variable-type matching on the created session.** `CreateDocumentSessionFrame` creates a variable typed as the store's `IDocumentSession`; a handler parameter typed as the abstraction needs to be satisfied by it. (Assignment is legal since the store interface derives from the abstraction — this is about how codegen matches the variable, not about the type relationship.)
3. **DI registration.** The stores register their own interfaces, not the abstractions, so `IDocumentSessionFactory` and friends don't resolve on a stock host. jasperfx#430 solved exactly this for `IProjectionCoordinator` by registering the JasperFx base interface alongside the store-specific subtype; the same move here would work. Consumers currently need a local `services.TryAddSingleton<IDocumentSessionFactory>(sp => sp.GetRequiredService<IDocumentStore>())` shim, which is easy to omit and fails at runtime rather than compile time.

## Why it matters

Without this, a handler in store-agnostic source has no way to take a session — it must name a concrete store type, which is precisely what `JasperFx.Events.Documents` exists to avoid. In CritterWatch this is what forces the whole console into a linked-source-plus-`global using` arrangement (one shared source tree compiled three times with a different store aliased in per project) rather than a single store-agnostic assembly with thin provider add-ons. A handful of our handlers load N documents by ids computed inside a loop, so they cannot use the declarative `[Entity]`/`[WriteModel]` route that sidesteps the session entirely — for those, taking the abstraction is the only path.

Happy to do the PR if the shape above is agreeable.

jeremydmiller:
Closes #3956.

A handler could not take `JasperFx.Events.Documents.IDocumentSessionOperations`, `IDocumentWriteOperations` or `IDocumentReadOperations` as a parameter — which is the whole point of those contracts existing, since store-agnostic source has no other way to take a session without naming a concrete store type.

## Two independent gaps, both of which had to close

1. **Variable matching.** Codegen matches a variable by its *exact* type, so nothing satisfied the parameter from the `IDocumentSession` the chain had already created. Added a `SharedDocumentOperationsSource` per store, alongside the existing `SharedEventOperationsSource` — the document-side counterparts to the event ones.
2. **`CanApply`.** The fixed probe list/predicate named none of them, so `AutoApplyTransactions` skipped the chain and no `SaveChangesAsync` postprocessor was attached.

## Measured, not assumed

The issue describes a silent write loss. On a **stock** host it's actually harder than that — the handler fails codegen outright with `UnResolvableVariableException`, because the stores don't register the abstractions in DI either.

The silent-loss half is real and was confirmed directly: with the variable source in place but the predicate reverted, the handler binds, runs, and the write is queued into the session's unit of work and discarded with no exception — two of the new tests go red while the other three stay green. Both halves are load-bearing.

## Design notes

- **All three contracts resolve from the chain's single `IDocumentSession`, including the read-only one.** Resolving reads from a separate `IQuerySession` would hand a handler that takes read and write contracts together two different sessions, and its reads would not see its own pending writes. There's a test pinning that they're the same instance.
- **`IDocumentReadOperations` is probed but deliberately NOT matched in the predicate**, exactly as `IQuerySession` has always been: a read-only parameter is not evidence that the chain writes anything.
- **DI registration** delegates to `IDocumentSession`/`IQuerySession` rather than opening a fresh session, so Marten's GH-3001 scope priming still applies and a service-located contract gets the handler's outbox-enrolled session. `IDocumentStore` already implements `IDocumentSessionFactory`, so that one is a straight singleton alias — the same move jasperfx#430 made for `IProjectionCoordinator`.

## No package bump needed

The handoff notes assumed this would need `JasperFx.Events` 2.48.0. It doesn't — 2.47.0 already carries every type, and Marten 9.23.0 / Polecat 5.12.0 / Fisher 0.7.0 all implement them. Verified by reflecting on the pinned assemblies.

## One thing worth flagging for adopters

Importing the `JasperFx.Events.Documents` **namespace** makes `ToListAsync()` ambiguous between `DocumentQueryableExtensions` and each store's own queryable extensions (`CS0121`). The tests alias the individual contracts instead of importing the namespace. That's upstream in JasperFx.Events, not something this PR can fix, but it will bite anyone converting a file to the store-agnostic contracts.

## Verification

- New `store_agnostic_document_session_parameters` (Marten): 5/5, confirmed red first
- Polecat + Fisher document contract tests added to the existing `all_queryable_and_event_store_operations` class (that suite balances CI shards by test-*class* count): 7/7 each
- Full `FisherTests`: 25/25

### DIFF
--- a/src/Persistence/FisherTests/all_queryable_and_event_store_operations.cs
+++ b/src/Persistence/FisherTests/all_queryable_and_event_store_operations.cs
@@ -1,5 +1,12 @@
 using JasperFx;
 using JasperFx.Events;
+// NB: importing the JasperFx.Events.Documents NAMESPACE here would make ToListAsync() ambiguous
+// between DocumentQueryableExtensions and Fisher's own queryable extensions (CS0121), so alias
+// just the contracts under test.
+using IDocumentSessionOperations = JasperFx.Events.Documents.IDocumentSessionOperations;
+using IDocumentWriteOperations = JasperFx.Events.Documents.IDocumentWriteOperations;
+using IDocumentReadOperations = JasperFx.Events.Documents.IDocumentReadOperations;
+using IDocumentSessionFactory = JasperFx.Events.Documents.IDocumentSessionFactory;
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.Hosting;
 using Fisher;
@@ -108,8 +115,52 @@ public class all_queryable_and_event_store_operations : IAsyncLifetime
         events.Count.ShouldBe(1);
         events[0].Data.ShouldBeOfType<FiLedgerEntryRecorded>().Note.ShouldBe("opening");
     }
-}
 
+    /// <summary>
+    ///     GH-3956. The store-agnostic JasperFx.Events.Documents contracts bind AND commit. Before this,
+    ///     a handler declaring one failed codegen outright on a stock host; once bound by the variable
+    ///     source alone, its writes were queued into the session's unit of work and silently discarded.
+    /// </summary>
+    [Fact]
+    public async Task document_session_operations_parameter_is_committed()
+    {
+        var id = Guid.NewGuid();
+
+        await _host.InvokeMessageAndWaitAsync(new StoreFiNote(id, "session-ops"));
+
+        await using var session = _host.Services.GetRequiredService<IDocumentStore>().LightweightSession();
+        var note = await session.LoadAsync<FiNote>(id, TestContext.Current.CancellationToken);
+
+        note.ShouldNotBeNull();
+        note.Text.ShouldBe("session-ops");
+    }
+
+    /// <summary>
+    ///     The read and write contracts must be satisfied by the same session variable, or a handler that
+    ///     writes and then reads would silently be reading from a different unit of work.
+    /// </summary>
+    [Fact]
+    public async Task read_and_write_operations_resolve_to_one_session()
+    {
+        var result = await _host.MessageBus()
+            .InvokeAsync<FiSessionsCompared>(new CompareFiSessions(), TestContext.Current.CancellationToken);
+
+        result.SameInstance.ShouldBeTrue();
+    }
+
+    /// <summary>
+    ///     Fisher registers IDocumentStore, never the lifted factory contract it already implements.
+    /// </summary>
+    [Fact]
+    public void document_session_factory_contracts_are_registered()
+    {
+        var store = _host.Services.GetRequiredService<IDocumentStore>();
+        _host.Services.GetRequiredService<IDocumentSessionFactory>().ShouldBeSameAs(store);
+        _host.Services
+            .GetRequiredService<JasperFx.Events.Documents.IDocumentSessionFactory<IDocumentSession, IQuerySession>>()
+            .ShouldBeSameAs(store);
+    }
+}
 public class FiWidget
 {
     public Guid Id { get; set; } = Guid.NewGuid();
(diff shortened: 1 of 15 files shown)


jeremydmiller:
Full-suite regression results after the change:

| Suite | Result |
|---|---|
| `MartenTests` | 605 / 605 |
| `PolecatTests` | 286 passed, 2 skipped |
| `FisherTests` | 25 / 25 |

Plus the new tests: `store_agnostic_document_session_parameters` 5/5 (Marten), and the document-contract additions to `all_queryable_and_event_store_operations` 7/7 on both Polecat and Fisher.

Issue #3785, opened 2026-08-02
Database affinity is per agent family, not per database: 73% of shard databases have durability and projections on different nodes


erdtsieck:
> **Rewritten 2026-08-03 after measuring this on production for a full hour.** The original report claimed that
> nodes joining a running cluster get only the durability family and *never* take projection work. That is
> wrong and I am retracting it — see "Retracted: the imbalance does converge" below. What survives, and what
> this issue is now about, is that database affinity is applied per agent *family* rather than per database, so
> two nodes end up connecting to the same shard database. That part reproduced on production and held steady
> over an hour.

Agent assignment honours database affinity within the `wolverinedb://` durability family and within the
`event-subscriptions://` family, but not *between* them. The result is that for most shard databases one node
owns the durability agent while a different node owns that database's projection agents.

## What we measured

Production, 5 pods, one Marten store over 512 tenant shard databases with `Events.UseTenantPartitionedEvents`,
~8,700 agents (6 async projections × ~1,385 tenants, plus ~460 durability agents). Measured on a converged
cluster, then again every five minutes for an hour:

| | databases |
|---|---|
| durability agent on the **same** node as that database's projections | 71 |
| on a **different** node | **375** |
| only one family assigned | 65 |

That is 73% split, and it did not move over the hour (it was 397/81 an hour earlier, so it drifts a little with
rebalancing but the ratio holds). The pool count derived from it stayed flat between 902 and 945.

## What it costs, measured rather than estimated

Npgsql keeps one connection pool per `(connection string, database)` per process, so a split database means two
pools against one shard instead of one. Joining `pg_stat_activity` on the shard server to the assignment table:

| connections held by | count |
|---|---|
| the projection owner (the node doing the work) | 1,483 |
| the durability owner | 434 |
| nodes owning neither family (ordinary API traffic for that tenant) | 1,025 |

**The split costs 425 connections** — the connections a durability owner holds against databases it would
otherwise never open. That is 14% of the total, not a doubling: a durability pool holds ~1.1 connections on
average, against 3 for a busy projection pool. I want to be precise about that because "two pools instead of
one" sounds like twice the connections and it is not.

14% still mattered. Earlier the same day this server sat at **2,393 of its 2,400 `max_connections`** for half an
hour during a projection version bump, throwing 7,500+ `53300 remaining connection slots are reserved…` in
fifteen minutes, briefly taking two of five pods out of the load balancer, and dropping assigned agents from
[...]

PR #3963, opened 2026-08-16
GH-3954: stop assigning durability agents to nodes that cannot run them

### LINKED ISSUE #3954: Leader assigns the wolverinedb durability agent to nodes with DurabilityAgentEnabled = false — assignment never converges and outgoing recovery never runs
## Summary

In a Balanced cluster, the leader assigns the `wolverinedb://…` durability agent to nodes running `Durability.DurabilityAgentEnabled = false`. Such a node cannot start the agent (`ArgumentOutOfRangeException: Unrecognized agent scheme 'wolverinedb'`), the leader re-issues the identical assignment forever, and — because no durability agent ends up running anywhere for that store — `owner_id = 0` outgoing envelopes are never recovered. All queue tables read zero the whole time, so the failure is silent. We found a 9-day backlog of undelivered envelopes in a staging environment this way.

Observed on 6.17.3 and 6.24.4; from reading `main`, the relevant path looks unchanged.

## Setup

Two-node cluster on PostgreSQL persistence + the PostgreSQL queue transport:

- **Node A (worker/consumer)**: durability enabled, `ListenToPostgresqlQueue(...).ExclusiveNodeWithParallelism(...)`.
- **Node B (web producer)**: stages envelopes inside caller-owned EF transactions (so they land in `wolverine_outgoing_envelopes` with `owner_id = 0` and rely on durability-agent recovery), no listeners, conventional discovery disabled, and `Durability.DurabilityAgentEnabled = false` — the intent being "this node stages but runs no recovery/scheduled agents".

## Observed behavior

1. Node A happened to hold leadership. The leader assigned the `wolverinedb://postgresql/<host>/<db>/wolverine` agent to **node B**.
2. Node B throws on the `StartAgent` command:
   ```
   System.ArgumentOutOfRangeException: Unrecognized agent scheme 'wolverinedb' (Parameter 'uri')
      at Wolverine.Runtime.Agents.NodeAgentController.startWithRetriesAsync(Uri agentUri)
      at Wolverine.Runtime.Agents.NodeAgentController.StartAgentAsync(Uri agentUri)
      at Wolverine.Runtime.Agents.StartAgent.ExecuteAsync(...)
   ```
   and the envelope moves to the error queue; the leader logs `Unable to confirm that agent wolverinedb://… started on node …` and `Giving up on a batched agent command after no progress for 00:05:00`.
3. The leader re-evaluates every ~5 minutes and makes the **same** assignment again (`wolverine_node_records` shows an `AssignmentChanged` row every ~5m10s, indefinitely). It never converges to the capable node.
4. Result: no durability agent runs for the store → `CheckRecoverableOutgoingMessagesOperation` never executes → producer-staged envelopes sit in `wolverine_outgoing_envelopes` (`owner_id = 0`) forever, while the queue tables stay empty.

## Why (from source)

- A node with `DurabilityAgentEnabled = false` never registers the store as an agent family, so it genuinely cannot start the agent (`NodeAgentController` ctor only does `_agentFamilies[_runtime.Stores.Scheme] = _runtime.Stores` when the flag is true).
- Node capabilities correctly **exclude** the scheme in that case (`StartLocalProcessing` populates `WolverineNode.Capabilities` from the registered families' `SupportedAgentsAsync`), so the leader has the data to avoid the node.
- But the assignment path for the store family doesn't consult capabilities: `MessageStoreCollection.EvaluateAssignmentsAsync` uses `DistributeEvenly` (6.x) / `DistributeEvenlyWithAffinity` (main, GH-3785, falling back to the even spread when there are no projection agents), and both spread across **all** grid nodes. The remainder pass even prefers non-leader nodes (`_nodes.FirstOrDefault(x => !x.IsLeader …)`), which in a two-node cluster deterministically targets the incapable producer whenever the consumer is leader.
- Capability matching (`AllNodesHaveSameCapabilities` / `MatchAgentsToCapableNodesFor`) exists in `DistributeByGroupAffinity` and `DistributeEvenlyWithBlueGreenSemantics`, but not in the plain even spread used for `wolverinedb` agents.

## Repro sketch (deterministic, single node)

A **single-node** Balanced cluster with `DurabilityAgentEnabled = false` reproduces it without any second node: the node elects itself leader, assigns the durability agent to itself, throws `Unrecognized agent scheme`, and an envelope staged into `wolverine_outgoing_envelopes` with `owner_id = 0` is never recovered into the queue. (That's how we pinned it in a regression test.)

## Expected

Either of these would resolve it:

- The leader skips nodes that don't advertise the agent's scheme in their capabilities when distributing `wolverinedb` agents (mirroring what the blue/green and group-affinity paths already do) — related in spirit to #3746 / #3747, which scoped the inverse case.
- Or `DurabilityAgentEnabled = false` is documented as unsupported for nodes participating in a Balanced cluster whose store has durability work.

## Workaround

We removed `DurabilityAgentEnabled = false` from the producer nodes, so any node the leader picks can run the agent. That's safe in practice because a Main-role store's cluster-assigned agent doesn't auto-start scheduled-job polling (GH-3376/GH-3439) and incoming recovery no-ops for endpoints without a local listener — but it does mean the durability settings (e.g. `KeepAfterMessageHandling`) must be kept identical across all nodes, since the agent runs with whichever node hosts it.

Thanks for Wolverine — happy to provide more logs or turn the single-node repro into a PR-able failing test if useful.

jeremydmiller:
Closes #3954.

A node started with `Durability.DurabilityAgentEnabled = false` never registers the durability agent family, so it threw `Unrecognized agent scheme 'wolverinedb'` the moment the leader handed it one. The leader re-issued the identical assignment every five minutes forever, no durability agent ran anywhere for that store, and `owner_id = 0` outgoing envelopes were never recovered — silently, with every queue table reading zero. The reporter found a nine-day backlog this way.

## ⚠️ The issue's premise is wrong, and following it would have made things worse

The issue states that the incapable node's capabilities correctly omit the scheme, so *"the leader has the data to avoid the node."*

**It does not.** `WolverineNode.Capabilities` is populated only from `_agentFamilies.Values.OfType<IStaticAgentFamily>()` (`NodeAgentController.StartLocalProcessing.cs:10`), and `MessageStoreCollection` is declared `: IAgentFamily, IAsyncDisposable` — **not** `IStaticAgentFamily`. There are exactly two sites that add to `Capabilities`, both fed by that same filter.

So no `wolverinedb://` Uri has ever appeared in `Capabilities` on *any* node, capable or not. The suggested fix — per-agent `Node.Capabilities.Contains(agent.Uri)` — would have found **zero** capable nodes everywhere and stranded the durability agent cluster-wide.

Per-agent matching is wrong here for a second, independent reason: `Capabilities` is a **startup snapshot**, while this family's agent list *grows* at runtime as tenant databases are added. Per-agent matching would strand every later-added tenant database's durability agent.

## What this does instead

Whether a node can run these agents is genuinely a **per-node** property — the setting turns the whole family on or off — so that is what gets published:

- `MessageStoreCollection.DurabilityCapabilityUri`, written into `Capabilities` at startup when the family is actually registered
- consulted by a new capability-aware overload of `DistributeEvenlyWithAffinity`

Only `MessageStoreCollection` and `MultiTenantedMessageDatabase` call that method, so the change is scoped to the store/durability family exactly as intended — `DistributeEvenlyWithBlueGreenSemantics` and `DistributeByGroupAffinity` are untouched.

It is deliberately **not** gated on `AllNodesHaveSameCapabilities`, which returns true trivially both when there is a single node (`_nodes.Skip(1)` is empty) and when every node is equally incapable — precisely the two configurations in the report. A fix modelled on the blue/green path would have changed neither.

Two more cases handled:

- **The GH-3785 affinity preference** follows another family's placements and may name a node that cannot host durability agents. An incapable preference is now discarded rather than honoured — co-location is an optimisation, capability is not.
- **When no node is capable**, nothing is assigned and a warning names the condition and the setting. The original failure was silent in both directions; leaving the agents unassigned without saying why would only trade one quiet failure for another.

## Verification

- 9 new `AssignmentGrid` tests. Confirmed **5/9 fail against the ungated overload** and 9/9 pass with the gate, so the gate is demonstrably what's doing the work (a plain stash can't serve as the red baseline here, since the tests reference new API).
- 2 new end-to-end tests asserting a normal Balanced Postgres host **publishes the marker and still gets its durability agent** — the regression that would matter most if the marker ever stopped being published.
- Existing `durability_modes` suite still 7/7, including its Balanced-mode running-agent count.

### DIFF
--- a/src/Persistence/PersistenceTests/Agents/durability_agent_capability_publication.cs
+++ b/src/Persistence/PersistenceTests/Agents/durability_agent_capability_publication.cs
@@ -1,0 +1,115 @@
+using IntegrationTests;
+using JasperFx.Core;
+using JasperFx.Resources;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Hosting;
+using Npgsql;
+using Shouldly;
+using Weasel.Postgresql;
+using Wolverine;
+using Wolverine.Persistence;
+using Wolverine.Postgresql;
+using Wolverine.Runtime;
+using Wolverine.Tracking;
+using Xunit;
+
+namespace PersistenceTests.Agents;
+
+/// <summary>
+///     GH-3954. The end-to-end half of the capability gate: the leader now refuses to assign
+///     <c>wolverinedb://</c> agents to a node that has not published
+///     <see cref="MessageStoreCollection.DurabilityCapabilityUri" />, so the marker actually reaching
+///     <c>WolverineNode.Capabilities</c> at startup is load bearing for every durable Wolverine application.
+///     If it ever stops being published, durability agents stop being assigned everywhere — this is the test
+///     that says so directly rather than leaving it to an agent count somewhere else.
+/// </summary>
+public class durability_agent_capability_publication : PostgresqlContext, IAsyncDisposable
+{
+    private IHost? _host;
+
+    private async Task<WolverineRuntime> startAsync(Action<WolverineOptions>? configure = null)
+    {
+        await using (var conn = new NpgsqlConnection(Servers.PostgresConnectionString))
+        {
+            await conn.OpenAsync();
+            await conn.DropSchemaAsync("durability_capability");
+        }
+
+        _host = await Host.CreateDefaultBuilder()
+            .UseWolverine(opts =>
+            {
+                opts.PersistMessagesWithPostgresql(Servers.PostgresConnectionString, "durability_capability");
+                opts.Services.AddResourceSetupOnStartup();
+                opts.Durability.Mode = DurabilityMode.Balanced;
+
+                configure?.Invoke(opts);
+            }).StartAsync();
+
+        return _host.GetRuntime();
+    }
+
+    private async Task<bool> waitForDurabilityAgent(TimeSpan timeout)
+    {
+        using var cts = new CancellationTokenSource(timeout);
+        while (!cts.IsCancellationRequested)
+        {
+            if (_host!.RunningAgents().Any(x => x.Scheme == "wolverinedb"))
+            {
+                return true;
+            }
+
+            try
+            {
+                await Task.Delay(250.Milliseconds(), cts.Token);
+            }
+            catch (TaskCanceledException)
+            {
+                break;
+            }
+        }
+
+        return _host!.RunningAgents().Any(x => x.Scheme == "wolverinedb");
+    }
+
+    public async ValueTask DisposeAsync()
+    {
+        if (_host != null)
+        {
+            await _host.StopAsync();
+            _host.Dispose();
+        }
+    }
+
+    [Fact]
+    public async Task a_durability_enabled_node_publishes_the_capability_and_gets_its_agent()
+    {
+        var runtime = await startAsync();
+
+        var node = await runtime.Storage.Nodes.LoadNodeAsync(runtime.Options.UniqueNodeId, CancellationToken.None);
+        node.ShouldNotBeNull();
+        node.Capabilities.ShouldContain(MessageStoreCollection.DurabilityCapabilityUri);
+
+        await _host!.WaitUntilAssumesLeadershipAsync(30.Seconds());
+
+        // The assignment the gate must NOT block. A regression in publishing the marker shows up here as
+        // an empty list rather than as a subtly wrong agent count somewhere else.
+        (await waitForDurabilityAgent(TimeSpan.FromSeconds(30)))
+            .ShouldBeTrue("the durability agent was never assigned to a node that IS capable of running it");
+    }
+
+    [Fact]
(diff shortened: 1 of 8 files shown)


jeremydmiller:
Regression sweep after the change (including the `released_agent_failover` NRE fix in the second commit):

| Suite | Result |
|---|---|
| `CoreTests` (full) | 2406 passed, 2 skipped, **0 failed** |
| `PersistenceTests` (full) | 92 / 92 |
| `PostgresqlTests` multi-tenant durability | 3 / 3 |
| `MartenTests.Distribution` | 49 / 49 |

Plus the new tests: 9 `AssignmentGrid` unit tests (5/9 confirmed failing against the ungated overload) and 2 end-to-end capability-publication tests.

Issue #3950, opened 2026-08-15
One rejected delivery tag on a busy Rabbit channel takes down the whole connection (RabbitMQ.Client SessionManager race, code=541)


jeremydmiller:
Found while investigating why `Bug_189_fails_if_there_are_many_messages_in_queue_on_startup` has been a chronic CIRabbitMQ failure. The test was the messenger; the finding underneath is worth its own issue.

### What happens

When the broker rejects a `basic.ack` — `PRECONDITION_FAILED - unknown delivery tag` — it closes that channel. If deliveries are still in flight on the channel being torn down, RabbitMQ.Client races itself:

```
System.Collections.Generic.Dictionary`2.get_Item(TKey key)
RabbitMQ.Client.Impl.SessionManager.Lookup(Int32 number)
RabbitMQ.Client.Framing.Connection.ProcessFrameAsync(InboundFrame frame, CancellationToken)
RabbitMQ.Client.Framing.Connection.ReceiveLoopAsync(CancellationToken)
RabbitMQ.Client.Framing.Connection.MainLoop()
---- System.Collections.Generic.KeyNotFoundException : The given key '4' was not present in the dictionary.
```

An inbound frame arrives for the channel number that was just removed from the session map, `Lookup` throws, and the client escalates that into a **library-initiated close of the entire connection**:

```
AlreadyClosedException : Already closed: The AMQP operation was interrupted:
AMQP close-reason, initiated by Library, code=541,
text='The given key '4' was not present in the dictionary.'
```

So the blast radius of one bad delivery tag is not the channel — it is every listener and sender on that connection.

**Wolverine cannot catch this.** It is thrown on the client's own `MainLoop` thread, inside the receive loop, after the ack has already gone out. Wolverine's settle paths are already hardened for the tag rejection itself (`RabbitMqChannelCallback.Complete` and `RabbitMqInteropFriendlyCallback.MoveToErrorsAsync` both swallow `AlreadyClosedException` when the message matches `PRECONDITION_FAILED - unknown delivery tag`), and that hardening is not the gap — the connection dies regardless.

### Measurements

All local, `Wolverine.RabbitMQ.Tests`, RabbitMQ from `docker compose`, client 7.1.2:

| Scenario | `KeyNotFoundException` | `code=541` |
|---|---|---|
| Bug_189 as written (~5% of 1000 deliveries poisoned) | every run (3/3) | every run |
| Bug_189, poisoning removed entirely | 0/3 | 0/3 |
[...]

Issue #3391, opened 2026-07-12
RabbitMQ: tracking-invariant regression test + re-consume gap on successful eager channel restart (follow-ups to #3370)


jeremydmiller:
Two non-blocking follow-ups from the #3370 review (RabbitMQ listener ghosting fix, merged):

## 1. Regression test for the ConnectionMonitor tracking invariant

The full broker-restart-with-callback-exception repro isn't deterministically achievable in the shared docker suite, but the tracking invariant IS cheaply testable: **an agent must remain tracked by `ConnectionMonitor` after a callback-exception-triggered restart** (success or failure). Before #3370, `_monitor.Remove(this)` in `RabbitMqChannelAgent.HandleChannelExceptionAsync` had no compensating re-`Track()` anywhere — even a successful eager restart left the agent permanently untracked and one connection drop away from ghosting.

Needs an `internal` accessor on `ConnectionMonitor._agents`. Test pattern: `src/Transports/RabbitMQ/Wolverine.RabbitMQ.Tests/Bugs/Bug_3171_channel_only_shutdown_recovery.cs` (internal-state assertions against the live broker, poll-based `WaitForAsync`, unique queue names, no shared fan_out/direct_exchange state).

## 2. Pre-existing gap: successful eager restart of a listener never re-consumes

On a *successful* eager restart via `HandleChannelExceptionAsync`, only `startNewChannel()` runs (`RabbitMqChannelAgent.cs:135`) — no re-declare / `BasicConsumeAsync`. A callback-exception-only failure (channel survives, or the shutdown handler was unsubscribed before close) can therefore leave an open channel with **zero consumers** and `State = Connected`. The shutdown-event push-heal from #3171/#3187 usually covers it, but the eager-restart path itself should either re-consume for listeners or defer to the listener's `ReconnectedAsync`.

Neither introduced by #3370 — it fixed the far worse permanent-ghosting variant. Noted during its review.

Refs #3370, #3171.

PR #3964, opened 2026-08-16
GH-3950: quiesce a Rabbit channel that the broker has already closed

jeremydmiller:
Addresses suggestion 3 of #3950. (Suggestion 2 shipped as #3961; suggestion 1 is the upstream `SessionManager.Lookup` defect and is not ours.)

When the broker rejects a settle with `PRECONDITION_FAILED - unknown delivery tag`, it has **already closed** the channel that tag arrived on. Wolverine logged the rejection and carried on, leaving deliveries streaming into a channel being torn down — which is precisely the condition that makes RabbitMQ.Client race itself: an inbound frame arrives for a channel number just removed from the session map, `SessionManager.Lookup` does an indexer read and throws `KeyNotFoundException`, and the client escalates that into a library-initiated close of the **whole connection** (`code=541`), taking down every listener and sender on it.

A rejected settle now cancels that channel's consumer, tears it down and rebuilds, through the existing `ReconnectedAsync()` path. Latched per channel generation so a burst of rejected settles on one dead channel triggers exactly one rebuild, and dispatched off the settle path because `ReconnectedAsync` takes the reconnect lock and does real broker work. Wired into all three sites that recognise the rejection.

## What this does not do

**It does not prevent the connection death.** The `KeyNotFoundException` is raised on the client's own `MainLoop` after our ack has already gone out, and the issue's own measurements show a *single* poisoned tag killed the connection on 4 of 5 runs — far too early for anything here to intervene.

What it does is narrow the window for further frames on a channel already known to be dead, and get the listener back onto a healthy one sooner. I'd rather state that plainly than let the change read as a fix for the 541s. The real fix is upstream: `SessionManager.Lookup` should `TryGetValue` and drop frames for a dead channel rather than throwing.

## About the test

The new test guards the hazard this change **introduces**, not the bug it mitigates.

#3391 is the precedent: a rebuild that only swaps the channel leaves a listener sitting on an open channel with **zero consumers** while still reporting Connected — silently dead. The poisoned message is redelivered by the broker either way, so no existing assertion would catch that; only a fresh batch published *after* the rebuild distinguishes "recovered" from "quietly stopped listening."

It also needed a poll-based wait helper: the existing `WaitForAll` completes a single `TaskCompletionSource`, so a second call in the same test returns immediately on the already-completed task and asserts against nothing. The first version of this test "passed" in 985ms that way.

## Verification

- `stale_delivery_tag_settling` 2/2, including the existing test
- Full `Wolverine.RabbitMQ.Tests` suite results to follow in a comment

### DIFF
--- a/src/Transports/RabbitMQ/Wolverine.RabbitMQ.Tests/stale_delivery_tag_settling.cs
+++ b/src/Transports/RabbitMQ/Wolverine.RabbitMQ.Tests/stale_delivery_tag_settling.cs
@@ -74,6 +74,65 @@ public class stale_delivery_tag_settling
 
         await host.StopAsync(TestContext.Current.CancellationToken);
     }
+
+    /// <summary>
+    /// GH-3950. A rejected settle now proactively rebuilds the listener's channel rather than leaving
+    /// deliveries streaming into a teardown. This asserts the listener is still CONSUMING afterwards.
+    /// </summary>
+    /// <remarks>
+    /// This is the guard for the hazard that proactive rebuild introduces rather than for the bug it
+    /// mitigates. #3391 is the precedent: a rebuild that only swaps the channel leaves a listener sitting
+    /// on an open channel with ZERO consumers while still reporting Connected — silently dead, and no
+    /// existing assertion catches it because the poisoned message itself is redelivered by the broker
+    /// regardless. Publishing a fresh batch AFTER the rebuild is what distinguishes "recovered" from
+    /// "quietly stopped listening".
+    /// </remarks>
+    [Fact]
+    public async Task the_listener_keeps_consuming_after_a_rejected_settle_rebuilds_the_channel()
+    {
+        var queueName = RabbitTesting.NextQueueName();
+
+        StaleTagHandler.Reset(poisonAt: 2);
+
+        using var host = await Host.CreateDefaultBuilder()
+            .UseWolverine(opts =>
+            {
+                opts.UseRabbitMq().AutoProvision().AutoPurgeOnStartup();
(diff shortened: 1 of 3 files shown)


jeremydmiller:
Full `Wolverine.RabbitMQ.Tests` suite on this branch: **503 passed / 504**, with one failure in `ConventionalRouting.end_to_end_with_conventional_routing.send_from_one_node_to_another_all_with_conventional_routing`.

That test passes 3/3 when its class is run in isolation, so it looks like suite-level interference rather than a defect in this change — but that suite has a documented history of chronic first-round failures, and I'd rather not call it a flake from one data point. Running the same full suite against `main` with this change reverted to get an apples-to-apples baseline; will post the comparison here.

jeremydmiller:
Baseline comparison done — **the failure is pre-existing on `main`, not caused by this change.**

Same machine, same broker, same full-suite invocation:

| | total | failed | failing test |
|---|---|---|---|
| this branch | 504 | 1 | `end_to_end_with_conventional_routing.send_from_one_node_to_another_all_with_conventional_routing` |
| `main`, this change reverted | 503 | 1 | same test |

(The totals differ by exactly the one test this PR adds, which confirms the two runs are distinct.)

So the full-suite result for this branch is **503 passed with one pre-existing failure**, and `stale_delivery_tag_settling` is 2/2.

Worth filing separately: the handoff notes recorded this suite at 503/503 as recently as yesterday, so this test has started failing on `main` since — the candidates are the three RabbitMQ changes merged since (#3796, #3960, #3961). Recording that as a lead rather than a diagnosis; it is out of scope here.

PR #3967, opened 2026-08-16
GH-3965: stop the ConventionalRouting tests leaking Rabbit consumers

### LINKED ISSUE #3965: ConventionalRouting tests leak Rabbit consumers on the shared "routed" queue, so end_to_end_with_conventional_routing loses its message to a competing consumer
## Symptom

`ConventionalRouting.end_to_end_with_conventional_routing.send_from_one_node_to_another_all_with_conventional_routing` fails in a full `Wolverine.RabbitMQ.Tests` run and passes 3/3 when its class runs in isolation.

```
System.TimeoutException : This TrackedSession timed out before all activity completed.
Activity detected:
| Service (Node Id)  | Message Id  | Message Type | Time (ms) | Event |
| Sender (7db8333a…) | 08defbc1-…  | routed       |         2 | Sent  |
```

The message is **Sent and never Received by anyone**. Nothing consumed it in the receiver.

## Cause

Every message routed conventionally for `ConventionallyRoutedMessage` lands on a **fixed** queue name — the type carries `[MessageIdentity("routed")]`, so the conventional queue is literally `routed`. Five classes in `Wolverine.RabbitMQ.Tests.ConventionalRouting` stand up hosts against that same queue, several of them with conventional discovery enabled so `RoutedMessageHandler` gives them a live listener on it.

And every one of those classes tears down like this:

```csharp
ValueTask IAsyncDisposable.DisposeAsync() => ValueTask.CompletedTask;

public void Dispose()
{
    _sender?.Dispose();
    _receiver?.Dispose();
}
```

`IHost.Dispose()` does **not** run `IHostedService.StopAsync`, so the Rabbit listeners are never cancelled and their consumers stay attached to `routed` on the broker. A later class's message is then delivered to a leaked consumer belonging to an already-finished test, and the real receiver never sees it — hence a tracked session showing `Sent` with no `Received`.

This is the same fixed-queue-name interference class of problem tracked in #3763, and the same `IHost.Dispose()` gap that fixtures elsewhere in the repo work around by awaiting `StopAsync()`.

## Why now

The full suite was recorded at 503/503 as recently as 2026-08-15. The leak itself is older than that — what changed is presumably ordering or listener shutdown timing, with #3796, #3960 and #3961 as the candidates, all of which touched `RabbitMqListener` shutdown. Recording that as a lead, not a diagnosis; the leak is the defect regardless of what tipped it.

Note this is **test infrastructure**, not a product defect — no shipped code path depends on `Dispose()` stopping a host.

## Fix

Have these classes actually stop their hosts (`await host.StopAsync()`) in a real `DisposeAsync`, rather than leaving a no-op. `ConventionalRoutingContext` needs the same, and its derived classes must not shadow it with their own empty `DisposeAsync`.

jeremydmiller:
Closes #3965. Test infrastructure only — no shipped code path depends on `Dispose()` stopping a host.

## The evidence

`end_to_end_with_conventional_routing.send_from_one_node_to_another_all_with_conventional_routing` failed in a full `Wolverine.RabbitMQ.Tests` run and passed 3/3 in isolation. The tracked session is what pinned it:

```
System.TimeoutException : This TrackedSession timed out before all activity completed.
| Service (Node Id)  | Message Id | Message Type | Time (ms) | Event |
| Sender (7db8333a…) | 08defbc1-… | routed       |         2 | Sent  |
```

**Sent, and never Received by anyone.** Not "received by the wrong host" — nothing in the session consumed it at all.

## The cause

`ConventionallyRoutedMessage` carries `[MessageIdentity("routed")]`, so every conventionally routed instance lands on one **fixed** queue named `routed`. Five classes in this namespace stand up hosts against it, several with conventional discovery on, which gives them a live listener via `RoutedMessageHandler`.

Every one of them tore down like this:

```csharp
ValueTask IAsyncDisposable.DisposeAsync() => ValueTask.CompletedTask;
public void Dispose() => _host.Dispose();
```

`IHost.Dispose()` does **not** run `IHostedService.StopAsync`, so the Rabbit listeners were never cancelled and their consumers stayed attached to `routed` on the broker after the class finished. A later class's message was delivered to one of those orphans, and the real receiver never saw it.

## The trap in the fix

Making `ConventionalRoutingContext` implement `IAsyncDisposable` is **not sufficient on its own**. Its derived classes implement `IAsyncLifetime`, whose explicit `ValueTask IAsyncDisposable.DisposeAsync()` **shadows** the base implementation — so the base would simply never run. Each derived class therefore calls the new `protected DisposeHostAsync()` explicitly.

## Deliberately not touched

`discover_with_naming_prefix` and `Bug_3633_conventional_routing_respects_named_broker` leak the same way, but they use prefixed and named-broker queue names that don't collide with plain `routed`, so they don't contribute to this interference. They also build their hosts in a constructor, so converting them means restructuring to `IAsyncLifetime` — a larger change than this fix warrants. Worth a follow-up if the suite ever shows symptoms from them.

## Verification

- `Wolverine.RabbitMQ.Tests.ConventionalRouting` namespace: **33/34 → 34/34**
- Full suite result to follow in a comment

## Note on timing

The full suite was recorded at 503/503 on 2026-08-15, so something tipped the ordering or shutdown timing recently — #3796, #3960 and #3961 all touched `RabbitMqListener` shutdown and are the candidates. The leak itself predates all of them, and is the defect regardless.

### DIFF
--- a/src/Transports/RabbitMQ/Wolverine.RabbitMQ.Tests/ConventionalRouting/Bug_2681_handler_type_naming_binds_all_exchanges.cs
+++ b/src/Transports/RabbitMQ/Wolverine.RabbitMQ.Tests/ConventionalRouting/Bug_2681_handler_type_naming_binds_all_exchanges.cs
@@ -46,7 +46,13 @@ public class Bug_2681_handler_type_naming_binds_all_exchanges : IAsyncLifetime, 
         _runtime = _host.Services.GetRequiredService<IWolverineRuntime>();
     }
 
-    ValueTask IAsyncDisposable.DisposeAsync() => ValueTask.CompletedTask;
+    // GH-3965: IHost.Dispose() does not run StopAsync, so a synchronous teardown left this
+    // class's Rabbit consumers attached to a SHARED, fixed queue name and stealing later
+    // tests' messages. Stop the hosts for real.
+    async ValueTask IAsyncDisposable.DisposeAsync()
+    {
+        if (_host != null) await _host.StopAsync();
+    }
 
     [Fact]
     public void handler_queue_is_bound_to_every_handled_message_exchange()
@@ -127,7 +133,13 @@ public class Bug_2681_custom_bindings_are_not_double_added : IAsyncLifetime, IDi
         _runtime = _host.Services.GetRequiredService<IWolverineRuntime>();
     }
 
-    ValueTask IAsyncDisposable.DisposeAsync() => ValueTask.CompletedTask;
+    // GH-3965: IHost.Dispose() does not run StopAsync, so a synchronous teardown left this
+    // class's Rabbit consumers attached to a SHARED, fixed queue name and stealing later
+    // tests' messages. Stop the hosts for real.
+    async ValueTask IAsyncDisposable.DisposeAsync()
+    {
+        if (_host != null) await _host.StopAsync();
+    }
 
     [Fact]
     public void user_custom_binding_is_not_double_added_by_the_convention()
--- a/src/Transports/RabbitMQ/Wolverine.RabbitMQ.Tests/ConventionalRouting/Bug_3041_saga_and_handler_with_handler_type_naming.cs
+++ b/src/Transports/RabbitMQ/Wolverine.RabbitMQ.Tests/ConventionalRouting/Bug_3041_saga_and_handler_with_handler_type_naming.cs
@@ -39,7 +39,13 @@ public class Bug_3041_saga_and_handler_with_handler_type_naming : IAsyncLifetime
         _runtime = _host.Services.GetRequiredService<IWolverineRuntime>();
     }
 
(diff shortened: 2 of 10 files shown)


jeremydmiller:
Full `Wolverine.RabbitMQ.Tests` suite: **504 / 504, zero failures.**

The progression across this session, all on the same machine and broker:

| | total | failed |
|---|---|---|
| `main` before #3964 | 503 | 1 |
| `main` + #3964 (GH-3950) | 504 | 1 |
| **+ this fix** | **504** | **0** |

Namespace-level: `Wolverine.RabbitMQ.Tests.ConventionalRouting` went 33/34 → 34/34.

That also retires the open question from #3964, where the same test failed on both the branch and the unmodified baseline — it was this leak, not the channel quiesce.

PR #3969, opened 2026-08-17
GH-3941: a store-agnostic EventsToAppend return type in core

### LINKED ISSUE #3941: Promote a store-agnostic `Events` return type to core — the IEnumerable<object> fallback is positional and can silently pick the wrong collection
`Wolverine.Marten.Events` and `Wolverine.Polecat.Events` are identical (`class Events : List<object>, IWolverineReturnType`) and store-named, so a handler that wants to be store-agnostic cannot use either. The store-agnostic path exists — it just has no name.

## Where this lands today

`IEventSourcingFrameProvider.TryApplyEventsFromReturnValues` resolves the events return in two steps:

```csharp
var eventsVariable =
    firstCall.Creates.FirstOrDefault(x => x.VariableType == provider.EventsCollectionType)
 ?? firstCall.Creates.FirstOrDefault(x =>
        x.VariableType.CanBeCastTo<IEnumerable<object>>() &&
        !x.VariableType.CanBeCastTo<IWolverineReturnType>());
```

So a store-agnostic handler returns a bare `IReadOnlyList<object>` / `List<object>` and hits the second clause. That works, on every provider, and is what we do in CritterWatch — e.g. a handler under `[WriteModel]` returning `(IReadOnlyList<object>, SomeMessage?)`. Nothing is broken.

## The problem with relying on the fallback

**It is positional and implicit.** `FirstOrDefault` takes the first match in `Creates`, and because `IEnumerable<T>` is covariant, *every* reference-typed collection in the return is castable to `IEnumerable<object>`. So:

```csharp
public static (IReadOnlyList<object>, IReadOnlyList<string>) Handle(SomeCommand cmd, [WriteModel] Thing thing)
```

has **two** candidates. Whichever appears first in `Creates` silently becomes the events appended to the stream. Nothing fails at codegen and nothing fails at runtime — you simply get the wrong collection in the event stream.

`OutgoingMessages` is safe today only because it is an `IWolverineReturnType`, which the fallback explicitly excludes. That is a happy accident of an unrelated marker interface, not a designed guarantee, and it does not extend to a user's own collection type.

The `[WriteModel]` / `[DeciderFunction]` attributes are what make store-agnostic handlers attractive in the first place (GH-3907), so the number of handlers relying on this fallback is going up, not down.

## Proposal

Promote a core `Events` to `Wolverine.Persistence.EventSourcing`, mirroring the two store-specific ones including the `operator +` accumulation sugar:

```csharp
namespace Wolverine.Persistence.EventSourcing;

public class Events : List<object>, IWolverineReturnType
{
    public Events() { }
    public Events(IEnumerable<object> collection) : base(collection) { }
    public static Events operator +(Events events, object @event) { events.Add(@event); return events; }
}
```

Then `(Events, OutgoingMessages)` becomes a declared, unambiguous, store-agnostic signature.

⚠️ **The design detail that matters:** it cannot simply be added and left to the fallback. `IWolverineReturnType` is exactly what the fallback *excludes*, so a core `Events` would be skipped. The resolution needs a check for the core type ahead of the fallback, alongside `provider.EventsCollectionType` — roughly:

```csharp
var eventsVariable =
    firstCall.Creates.FirstOrDefault(x => x.VariableType == provider.EventsCollectionType)
 ?? firstCall.Creates.FirstOrDefault(x => x.VariableType == typeof(Events))   // new
 ?? firstCall.Creates.FirstOrDefault(x => /* existing IEnumerable<object> fallback */);
```

The same addition is needed in the sibling paths that repeat this shape — `DcbModelAttribute` has the identical `IEnumerable<object>` check, and `RegisterEventsFrame` / `BoundaryEventCaptureFrames` both branch on `CanBeCastTo<IEnumerable<object>>()`.

The store-specific `Wolverine.Marten.Events` / `Wolverine.Polecat.Events` should stay as-is for existing code, exactly as `WriteAggregateAttribute` was kept alongside `WriteModelAttribute` in GH-3907.

## Explicitly not proposed

This is the **single-stream** case only, where `[WriteModel]` pins the aggregate and every event goes to its stream. It does nothing for "append to a variable number of *different* streams", which is a separate gap — `SideEffectPolicy.lookForSingularSideEffects` walks individual return variables via `chain.ReturnVariablesOfType<ISideEffect>()`, so a `List<AppendEvents>` is not seen as a side effect at all. Worth keeping the two apart.

jeremydmiller:
Closes #3941 in mechanism — but **please read the naming section first, because the type name is a genuine problem and I think it should change before this merges.**

---

## ⚠️ The `Events` name collides, and it collides on exactly the use case this feature exists for

The issue proposed naming the core type `Events`, mirroring the three store-specific ones. Implementing it, that turns out to be a source-breaking change for the primary intended usage.

The two namespaces a store-agnostic handler needs are:

- `Wolverine.Persistence.EventSourcing` — where `[WriteModel]` / `[ReadModel]` / `[DeciderFunction]` live, i.e. the GH-3907 vocabulary that makes store-agnostic handlers possible at all
- `Wolverine.Marten` (or `.Polecat` / `.Fisher`) — which any real Marten app already imports

Put the core `Events` in the first and a handler importing both cannot name `Events` unqualified:

```
error CS0104: 'Events' is an ambiguous reference between
  'Wolverine.Persistence.EventSourcing.Events' and 'Wolverine.Marten.Events'
```

**Demonstrated, not theorized.** The regression test in this PR carries an explicit `using Events = Wolverine.Persistence.EventSourcing.Events;` alias. Removing that alias fails the build with CS0104 on **the handler signature lines themselves** — lines 113 and 119, i.e. the return types. Both TFMs, 8 errors.

So the collision does not land on some incidental call site. It lands on the declaration the feature is for, in any app that already uses a store integration — which is every app.

### Options

1. **Rename the core type.** `EventsToAppend` and `StreamEvents` are both free across the whole repo (checked). `AppendEvents` is **not** — it is already `Wolverine.Persistence.EventSideEffects.AppendEvents : ISideEffectAware`, so it is out. My preference is `EventsToAppend`: it reads correctly in a signature — `public static (EventsToAppend, OutgoingMessages) Handle(...)` — is self-documenting about *when* the append happens, and carries zero collision risk.
2. **Keep `Events`** and document that users combining both namespaces need a `using` alias. Symmetric with the store types, but it taxes precisely the users the feature targets, and the error they hit first is a compile break in code that used to build.

I have left it as `Events` so the diff matches the issue as written and the decision is yours. **Say the word and I will rename** — it is a mechanical change to one file plus the two resolution sites plus the test.

---

## The mechanism

`Wolverine.Marten.Events`, `Wolverine.Polecat.Events` and `Wolverine.Fisher.Events` are identical and store-named, so a handler that wants to be store-agnostic could not name any of them. The store-agnostic path existed — a bare `IEnumerable<object>` return is picked up by the fallback — but that fallback is **positional**: `IEnumerable<T>` is covariant, so every reference-typed collection in a return tuple is a candidate and `FirstOrDefault` takes whichever lands first in `Creates`. Nothing fails at codegen and nothing fails at runtime; the wrong collection simply becomes the appended events.

`OutgoingMessages` escapes that only because it is an `IWolverineReturnType`, which the fallback excludes — an accident of an unrelated marker rather than a guarantee, and one that does not extend to a user's own collection type.

### The design detail

The core type **cannot** just be added and left to the fallback: being an `IWolverineReturnType` is exactly what the fallback excludes, so it would be silently skipped. It needs an explicit match ahead of the fallback, added in both places that resolve an events return:

- `IEventSourcingFrameProvider.DetermineEventCaptureHandling` (single-stream)
- `DcbModelAttribute` (DCB boundary)

### Two sites the issue named that need no change

Worth recording so nobody re-investigates:

- `RegisterEventsFrame.FindMethod` and `RegisterBoundaryEventsFrame.FindMethod` branch on `CanBeCastTo<IEnumerable<object>>()` to choose `AppendMany` over `AppendOne` — and `Events : List<object>` already satisfies that.
- `HandlerChain` already skips `IEnumerable<object>` variables when collecting publishable message types, so a returned `Events` is not also routed as a message.

The store-specific types stay exactly as they are, the same way `WriteAggregateAttribute` was kept alongside `WriteModelAttribute` in GH-3907.

Single-stream only. Appending to a variable number of *different* streams is a separate gap and is deliberately untouched.

## Tests

`core_events_return_3941`, 3 facts. The one that matters is `ambiguous_sibling_collection_does_not_become_the_events` — a handler returning `(Events, IReadOnlyList<string>)`. The other two would pass against the old fallback too.

**Proven both ways:** with the new resolution line removed, all 3 fail — `Events` falls through to a fallback that excludes it, and the audit strings are appended to the event stream instead. That is the silent corruption the issue describes, reproduced.

Green on net9.0 and net10.0. Also built `MartenTests`, `PolecatTests` and `FisherTests` — the heaviest importers of both namespaces — with 0 errors, though see the naming section for why that is weaker evidence than it looks: those files happen not to reference `Events` as a type.

### DIFF
--- a/src/Persistence/MartenTests/AggregateHandlerWorkflow/core_events_return_3941.cs
+++ b/src/Persistence/MartenTests/AggregateHandlerWorkflow/core_events_return_3941.cs
@@ -5,19 +5,22 @@ using Shouldly;
 using Wolverine;
 using Wolverine.Marten;
 using Wolverine.Persistence.EventSourcing;
-using Events = Wolverine.Persistence.EventSourcing.Events;
 
 namespace MartenTests.AggregateHandlerWorkflow;
 
-// GH-3941: the store-agnostic Events return type. Wolverine.Marten.Events, Wolverine.Polecat.Events
-// and Wolverine.Fisher.Events are identical and store-named, so a handler compiled against more than
-// one store could not name any of them and had to fall back to a bare IEnumerable<object> return.
+// GH-3941: the store-agnostic EventsToAppend return type. Wolverine.Marten.Events,
+// Wolverine.Polecat.Events and Wolverine.Fisher.Events are identical and store-named, so a handler
+// compiled against more than one store could not name any of them and had to fall back to a bare
+// IEnumerable<object> return. That fallback works but is positional: IEnumerable<T> is covariant, so
+// EVERY reference-typed collection in a return tuple is a candidate and FirstOrDefault takes
+// whichever lands first in Creates. Nothing fails at codegen and nothing fails at runtime - the wrong
+// collection just becomes the events. ambiguous_sibling_collection_does_not_become_the_events is the
+// fact that matters here; the other two would pass against the old fallback too.
 //
-// That fallback works but is positional: IEnumerable<T> is covariant, so EVERY reference-typed
-// collection in a return tuple is a candidate and FirstOrDefault takes whichever lands first in
-// Creates. Nothing fails at codegen and nothing fails at runtime - the wrong collection just becomes
-// the events. ambiguous_sibling_collection_does_not_become_the_events is the fact that matters here;
-// the others would pass against the old fallback too.
+// Note that this file imports BOTH Wolverine.Marten and Wolverine.Persistence.EventSourcing with no
+// using alias, which is the combination a real store-agnostic handler needs -- the store integration
+// plus [WriteModel]. That compiles only because the core type is not also called Events: naming it
+// so collided with CS0104 on the two handler signatures at the bottom of this file.
 public class core_events_return_3941 : IAsyncLifetime
 {
     private IHost theHost = null!;
@@ -88,9 +91,9 @@ public class core_events_return_3941 : IAsyncLifetime
     {
         var streamId = await givenAccount(50m);
 
-        // The handler returns (Events, IReadOnlyList<string>). Both are castable to
+        // The handler returns (EventsToAppend, IReadOnlyList<string>). Both are castable to
         // IEnumerable<object>, so under the old fallback alone the audit lines could be appended as
-        // events instead - which throws nothing and simply corrupts the stream. Declaring Events is
+        // events instead - which throws nothing and simply corrupts the stream. Declaring EventsToAppend is
         // what makes the choice deterministic.
         await theHost.InvokeAsync(new RecordAuditedDeposit(streamId, 5m));
 
@@ -111,15 +114,15 @@ public record RecordAuditedDeposit(Guid AccountId, decimal Amount);
 
 public static class RecordCoreDepositHandler
 {
-    public static Events Handle(RecordCoreDeposit command, [WriteModel] Account account)
+    public static EventsToAppend Handle(RecordCoreDeposit command, [WriteModel] Account account)
         => new(Enumerable.Range(0, command.Repeat).Select(object (_) => new AmountDeposited(command.Amount)));
 }
 
 public static class RecordAuditedDepositHandler
 {
-    public static (Events, IReadOnlyList<string>) Handle(
+    public static (EventsToAppend, IReadOnlyList<string>) Handle(
         RecordAuditedDeposit command,
         [WriteModel] Account account)
-        => (new Events { new AmountDeposited(command.Amount) },
+        => (new EventsToAppend { new AmountDeposited(command.Amount) },
             new[] { $"deposit of {command.Amount} against {command.AccountId}" });
 }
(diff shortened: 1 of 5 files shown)


jeremydmiller:
## Renamed to `EventsToAppend` (747f856f6)

Per review. The naming concern was correct, and the rename resolves it properly rather than papering over it.

**The alias is gone.** The regression test now imports `Wolverine.Marten` and `Wolverine.Persistence.EventSourcing` **bare** — the combination a real store-agnostic handler needs, store integration plus `[WriteModel]` — and compiles. So the regression test is now also the standing proof that the collision is gone; if anyone renames this back, that file stops building.

**On the candidates:**

- `AppendEvents` was **not** available — `Wolverine.Persistence.EventSideEffects.AppendEvents : ISideEffectAware` already exists.
- `EventList`, `EventSet` and `EventBag` are all free across wolverine, marten, polecat, fisher and jasperfx (checked each).
- `EventsToAppend` says *when* the append happens, which a bare noun does not — and the tense matters here, because the value is a pending intention that the generated frame acts on, not a record of something already written.

The rationale is recorded in the doc comment on the type itself, so the name does not get "corrected" back to the symmetric-looking one later.

3/3 green on net9.0 and net10.0.

PR #3976, opened 2026-08-17
GH-3975: add a declarative way to run work AFTER the transactional commit

### LINKED ISSUE #3975: No declarative way to run work AFTER the transactional commit — After methods insert at the front, so they run before it
Add a declarative way to say "run this after the transactional commit" — a convention or attribute that **appends** to `Postprocessors` rather than inserting at the front.

## The problem

`After` reads like a post-handler hook that runs at the end. It does not run after the commit. `Chain.cs:360` and `MiddlewarePolicy.cs:94` both do:

```csharp
chain.Postprocessors.Insert(i, frame);
```

— front of the list. The transactional commit is itself a postprocessor (`DocumentSessionSaveChanges`, added by each store's `PersistenceFrameProvider.ApplyTransactionSupport`), so an `After` method runs **before** the write is durable.

That is a reasonable default for "after the handler," but it means there is currently **no supported way to express "after the commit"** — even though the position exists and Wolverine itself uses it: `ApplyTransactionSupport` does

```csharp
chain.Postprocessors.Add(new DocumentSessionSaveChanges());
chain.Postprocessors.Add(new FlushOutgoingMessages());
```

so `FlushOutgoingMessages` is a first-party post-commit postprocessor.

## What applications do instead, and why it's fragile

They hand-write a `Frame` and `Add` it from a policy sequenced after the persistence policy. That works — but the ordering is **positional and unstated**. Nothing in the source declares the intent, nothing validates it, and a change in policy application order silently moves the frame to the wrong side of the commit with every test still green.

The failure mode is the bad one: side effects that are supposed to follow a durable write instead run before it, and only diverge when a commit fails.

## Where this has cost real time

Two cases in CritterWatch, both of which reached for `After` first because the name implies it:

1. An alert raise/resolve latch that had to be set only after the append committed. Using `After` recorded "alert raised" for a transaction that could still fail. The workaround was a hand-written early `SaveChangesAsync` inside the handler.
2. An ingest cache write-back that must not publish state for a batch that was never persisted — same shape, same workaround. Its code comment calls the resulting explicit save "the most load-bearing one in the codebase."

In both cases the *correct* mechanism (a frame appended after the commit) existed but had to be discovered by reading Wolverine's own frame-provider source.

## Ask

Something like:

```csharp
public static class SomeHandler
{
    public static void Handle(...) { }

    [WolverineAfterCommit]                       // or: an `AfterCommit` conventional method name
    public static void AfterCommit(...) { }
}
```

that appends to `Postprocessors` after the persistence provider's commit frame, with the same variable-binding rules `After` already has.

Two properties matter more than the spelling:

- **It must land after the commit frame regardless of policy ordering** — that is the whole point; getting it by luck of sequencing is what applications already do.
- **It must not run when the commit throws.** Frames are concatenated without a `try`/`finally` (`HandlerChain.cs:709-717`), so unwinding already gives this — worth pinning with a test so it stays true.

## Nice-to-have

If `After`'s pre-commit position is kept (and it probably should be, for compatibility), consider mentioning the distinction in the middleware docs. The name is doing a lot of work and currently points the wrong way for anyone thinking about durability.

jeremydmiller:
Closes #3975.

`After` reads like a post-handler hook that runs at the end. It does not run after the commit. The commit is itself a postprocessor — each provider's `ApplyTransactionSupport` appends one — and `After` methods are *inserted at the front* of that list, so they run **before** the write is durable. There was no supported way to express the other side of it, even though Wolverine uses that position itself for `FlushOutgoingMessages`.

## What's added

`AfterCommit` / `AfterCommitAsync` as a conventional method name, and `[WolverineAfterCommit]` for methods named something else. Same parameter binding `After` already has; works on message handlers, sagas and HTTP endpoints.

```csharp
public static class RaiseAlertHandler
{
    public static void Handle(RaiseAlert command, IDocumentSession session)
        => session.Events.Append(command.Id, new AlertRaised(command.Reason));

    // Only runs if the append above actually committed
    public static void AfterCommit(AlertLatch latch, RaiseAlert command)
        => latch.MarkRaised(command.Id);
}
```

## The position is structural, not positional

This is the part that matters. Frames go into a new `IChain.PostCommitPostprocessors` list that is concatenated after every postprocessor at frame-assembly time — **not** appended to `Postprocessors` from a policy sequenced after the persistence policy.

That's the issue's actual complaint: an application that gets the position right by luck of policy ordering gets it wrong the moment that ordering changes, silently, with every test still green. A separate list can't drift that way.

`After`'s pre-commit position is deliberately unchanged, and there's a compatibility guard per provider asserting it stays there. Plenty of post-handler work has nothing to do with durability.

## Verified per provider, not once

Each of **Marten, Polecat, Fisher, EF Core, RavenDb and CosmosDb** gets a codegen test asserting the emitted `AfterCommit` call really does land after *that provider's* commit frame, plus the guard that `After` still lands before it. List membership alone would only prove the frame went into the right list; the guarantee is about where it ends up relative to a frame contributed from a completely different code path, and only the emitted source shows that.

These are pure codegen assertions, so **none needs a live server** — including RavenDb, whose `DocumentStore.Initialize()` doesn't open a connection.

One provider difference worth flagging: **CosmosDb has no commit postprocessor at all.** Its `ApplyTransactionSupport` puts a `TransactionalFrame` in the *middleware* (the outbox enlistment) and only `FlushOutgoingMessages` in the postprocessors. Its test asserts against the flush rather than a `SaveChangesAsync` it never emits.

## Core behaviour pinned

- the frame lands in `PostCommitPostprocessors` and **not** in `Postprocessors`
- the attribute works without the conventional name
- observed call order is `Handle → After → AfterCommit`
- an after-commit method does **not** run when an earlier postprocessor throws — frames are concatenated without a `try`/`finally`, so the exception unwinds past them. The issue asked for this to be pinned so it stays true.

## Docs

The issue's nice-to-have was that the `After` name "points the wrong way for anyone thinking about durability." `docs/guide/handlers/middleware.md` now shows the emitted order explicitly, with the two caveats: these don't run when the commit throws, and they run after the outbox flush, so a cascade from one isn't atomic with the write.

## Verification

- `CoreTests` middleware area: 33/33; new `after_commit_middleware` 4/4
- Marten / Polecat / Fisher / EF Core / RavenDb / CosmosDb: 2/2 each
- Full `wolverine.slnx` builds clean pinned to `net9.0`

### DIFF
--- a/docs/guide/handlers/middleware.md
+++ b/docs/guide/handlers/middleware.md
@@ -137,7 +137,66 @@ Here's the conventions:
 |----------------------------------------------------------|-----------------------------|
 | Before the Handler(s)                                    | `Before`, `BeforeAsync`, `Load`, `LoadAsync`, `Validate`, `ValidateAsync` |
 | After the Handler(s)                                     | `After`, `AfterAsync`, `PostProcess`, `PostProcessAsync` |
+| After the transactional commit                           | `AfterCommit`, `AfterCommitAsync` |
 | In `finally` blocks after the Handlers & "After" methods | `Finally`, `FinallyAsync`   |
+
+## "After" is before the commit {#after-vs-after-commit}
+
+::: warning
+`After` runs after the *handler*, not after the *transaction*. If your side effect must only happen once
+the write is durable, use `AfterCommit` instead.
+:::
+
+This trips people up because the name reads the other way around. When a chain is transactional, the commit
+itself is a postprocessor contributed by the persistence provider — Marten's `SaveChangesAsync`, EF Core's
+`SaveChangesAsync`, and so on. `After` methods are inserted at the *front* of the postprocessors, so they run
+**before** that commit:
+
+```
+handler()
+After()            // <-- the write is NOT durable yet, and the transaction may still fail
+SaveChangesAsync() // <-- the commit
+FlushOutgoingMessages()
+AfterCommit()      // <-- the write is durable
+```
+
+So an `After` method that records "the alert was raised" or writes back to a cache is recording something for
+a transaction that can still roll back. That is exactly the shape of bug this distinction exists to prevent.
+
+Use `AfterCommit` / `AfterCommitAsync`, or the `[WolverineAfterCommit]` attribute, when the side effect must
+follow a durable write:
+
+```csharp
+public static class RaiseAlertHandler
+{
+    public static void Handle(RaiseAlert command, IDocumentSession session)
+    {
+        session.Events.Append(command.Id, new AlertRaised(command.Reason));
+    }
+
+    // Only runs if the append above actually committed
+    public static void AfterCommit(AlertLatch latch, RaiseAlert command)
+    {
+        latch.MarkRaised(command.Id);
+    }
+}
+```
+
+These methods bind parameters exactly like `After` methods do, and they work on message handlers, sagas and
+HTTP endpoints alike.
+
+Two behaviours worth knowing:
+
+* **They do not run when the commit throws.** The frames are concatenated without a `try`/`finally`, so an
+  exception from the commit unwinds straight past them. If you need something to run either way, that is what
+  `Finally` is for.
+* **They run after the outbox flush as well as the commit.** A cascading message returned from an
+  `AfterCommit` method is therefore sent through the normal end-of-pipeline flush rather than the transaction
+  that just committed. If a message has to be atomic with the write, cascade it from the handler instead.
+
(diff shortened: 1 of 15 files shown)


jeremydmiller:
Full `CoreTests` regression on this branch: **2410 passed, 2 skipped, 0 failed** (2412 total).

Per-provider after-commit tests: Marten 2/2, Polecat 2/2, Fisher 2/2, EF Core 2/2, RavenDb 2/2, CosmosDb 2/2. Full `wolverine.slnx` builds clean pinned to `net9.0`.

PR #3985, opened 2026-08-20
GH-3983: make the MySQL envelope storage migration idempotent, and upgrade to Weasel 9.25.0

### LINKED ISSUE #3983: MySQL: envelope storage migration fails on every startup
**Describe the bug**

On MySQL, the Wolverine envelope storage migration fails on **every** startup with:

```
Error executing Wolverine Envelope Storage database migration:
DROP INDEX `fk_wolverine_node_assignments_node_id` ON <schema>.wolverine_node_assignments

MySqlConnector.MySqlException (0x80004005): Cannot drop index
'fk_wolverine_node_assignments_node_id': needed in a foreign key constraint
   at Weasel.MySql.MySqlMigrator.executeCommand(...) in /_/src/Weasel.MySql/MySqlMigrator.cs:line 116
```

`wolverine_node_assignments` is created with an index and a foreign key constraint that share the same name. In MySQL a `FOREIGN KEY` constraint automatically creates a backing index named after the constraint, so the explicitly declared index and the FK's implicit index are the same object. The schema diff then treats the index as extra and issues `DROP INDEX`, which InnoDB refuses because the constraint depends on it.

This is **not** an upgrade artefact — it reproduces against a brand-new empty database on first start, and then repeats on every subsequent start.

**To Reproduce**

1. Create an empty MySQL 8.4 database.
2. Run this minimal host (`net10.0`, `WolverineFx.MySql` 6.29.0, `WolverineFx.RuntimeCompilation` 6.29.0):

```csharp
const string database = "wolverine_repro";
const string cs = $"Database={database};DataSource=localhost;Uid=user;Password=pass;";

var builder = Host.CreateApplicationBuilder();
builder.UseWolverine(opts =>
{
    opts.Discovery.DisableConventionalDiscovery();
    opts.PersistMessagesWithMySql(cs, database);
    opts.Services.AddResourceSetupOnStartup();
});

using var host = builder.Build();
await host.StartAsync();   // error is logged here
```

3. Observe the error logged from `Wolverine.MySql.MySqlMessageStore` at `Error` level.
4. Stop and start again — the same error is logged every time.

The table Wolverine itself creates shows the name collision:

```sql
SHOW CREATE TABLE wolverine_node_assignments;
```

```sql
CREATE TABLE `wolverine_node_assignments` (
  `id` varchar(255) NOT NULL,
  `node_id` char(36) DEFAULT NULL,
  `started` datetime NOT NULL DEFAULT (utc_timestamp(6)),
  PRIMARY KEY (`id`),
  KEY `fk_wolverine_node_assignments_node_id` (`node_id`),
  CONSTRAINT `fk_wolverine_node_assignments_node_id` FOREIGN KEY (`node_id`)
      REFERENCES `wolverine_nodes` (`id`) ON DELETE CASCADE
) ENGINE=InnoDB
```

**Expected behavior**

Resource setup completes without error, and a subsequent start is a no-op. Either the index and the FK constraint should not share a name, or the MySQL schema diff should recognise an index that backs a foreign key and not attempt to drop it.

**Screenshots**

N/A — server-side library issue.

**Desktop (please complete the following information):**

- OS: Windows 11 (MySQL running in Docker, `mysql:8.4.3`)
- Browser: N/A
- Version:
  - `WolverineFx.MySql` **6.29.0**
  - `WolverineFx.RuntimeCompilation` **6.29.0**
  - `MySqlConnector` **2.4.0**
  - .NET SDK **10.0.400**
  - MySQL Server **8.4.3**, `sql_mode = NO_ENGINE_SUBSTITUTION`

**Smartphone (please complete the following information):**

N/A

**Additional context**

- **Non-fatal but persistent.** The tables are created and function correctly; message persistence, node registration and agent assignment all work. The cost is an `Error`-level log entry on every startup of every node, which trains people to ignore errors from `MySqlMessageStore`.
- Possibly the same root cause as #1855 ("Migration failure on PostgreSQL due to index ownership preventing expires column migration"), which is the same shape — a migration blocked by index ownership — on a different provider and table.
- Confirmed separately that **GH-2940** (`PersistNodeRecord` emitting ANSI double-quoted schema identifiers, which MySQL rejects) **is** fixed in 6.29.0 — `wolverine_node_records` now populates correctly on MySQL. This report is a distinct remaining MySQL issue.

jeremydmiller:
Closes #3983.

> ⛔ **Blocked on [JasperFx/marten#5271](https://github.com/JasperFx/marten/pull/5271).** Draft until that ships and the Marten pin moves off 9.23.0. Everything else is green. See *Merge order* below.
>
> The original blocker — weasel#445 — **is resolved**: it merged and shipped in **Weasel 9.25.0**, which this PR now pins.

@cyclr-adrian's report was exactly right, including the diagnosis. Reproducing it turned up **four** drift items on every startup, not one:

```
ALTER TABLE <schema>.wolverine_nodes MODIFY COLUMN `node_number` INT AUTO_INCREMENT NOT NULL;
DROP INDEX `fk_wolverine_node_assignments_node_id` ON <schema>.wolverine_node_assignments;
ALTER TABLE <schema>.wolverine_node_assignments DROP FOREIGN KEY `fk_wolverine_node_assignments_node_id`;
ALTER TABLE <schema>.wolverine_node_assignments ADD CONSTRAINT `fk_wolverine_node_assignments_node_id` ...;
ALTER TABLE <schema>.wolverine_node_records MODIFY COLUMN `id` INT AUTO_INCREMENT NOT NULL;
```

Only the `DROP INDEX` throws, which is why the rest went unnoticed — the migration runner logs the first failure and the remaining statements never run. The first migration against an empty database is a pure `CREATE`, so none of it is reachable until the second run.

## 1. The Wolverine bug: `AUTO_INCREMENT` in the column type

`wolverine_nodes.node_number` and `wolverine_node_records.id` folded `AUTO_INCREMENT` into the column **type** string. `information_schema` reports the type as plain `INT` and carries `auto_increment` separately in `EXTRA`, so neither column ever compared equal and both emitted a `MODIFY COLUMN` forever. `Weasel.MySql` has a first-class `AutoIncrement()` for this. The generated DDL is unchanged apart from where the keyword sits, so **existing databases need no migration**.

## 2. The Weasel bugs, now shipped

The other three were `Weasel.MySql` defects, fixed in [weasel#445](https://github.com/JasperFx/weasel/pull/445) and released in **9.25.0**: the foreign key's backing index being diffed as an extra index, `WriteUpdate` emitting index changes before foreign key changes (so *removing* a foreign key failed the same way), and `Table(DbObjectName)` not normalizing to `MySqlObjectName` — which is what made the constraint report as `Different` on every check.

## 3. Fallout from the 9.25 upgrade: NServiceBus PostgreSQL interop

9.25 is Weasel's identifier release, and it **stops case-folding and rewriting column names**. `NServiceBusQueueTable` declared PascalCase columns and the send/receive SQL referenced them unquoted — with a comment saying outright that it relied on Weasel folding them to lowercase. Under 9.25 the table gets a genuinely case-sensitive `"Seq"` and the destructive pop fails:

```
Npgsql.PostgresException : 42703: column "Seq" does not exist
```

**The PascalCase declaration was wrong all along.** A real NServiceBus-provisioned queue table has lowercase columns — verified against a live `NServiceBus.Transport.PostgreSql` host when this transport was built — so Wolverine was only ever interoperating because Weasel silently folded the difference away. The columns are now declared lowercase directly, which is what both sides actually use. No behavior change against an NServiceBus-owned table.

## Testing

New `Bug_3983_repeated_envelope_storage_migration`: migrate an already-current database and assert no drift. The existing `can_migrate_schema_directly` migrates **once**, against an empty database — none of this is reachable until the second run, which is precisely the shape of the bug report.

Full suites on Weasel 9.25.0, all green:

| Suite | Result |
| --- | --- |
| MySql | 125 / 125 |
| PostgreSQL | 481 / 482 (1 skipped) |
| SQL Server | 396 / 398 (2 skipped) |
| SQLite | 168 / 169 (1 skipped) |
| EF Core | 205 / 205 |
| `wolverine.slnx` Release `-f net9.0` | clean, 0 warnings |

`MartenTests` is **607 / 610** — see below. `FisherTests` has 9 failures that are **pre-existing on `main`** and unrelated to this PR (verified against a clean `origin/main` checkout); worth a separate issue.

## The remaining blocker: a Marten defect this bump exposes

`mt_natural_key_X` shortens every name it writes — foreign keys and its index — but left the primary key constraint on Weasel's `pkey_{table}_{columns}` default, which is 64 characters for a 16-character aggregate name under conjoined tenancy. PostgreSQL truncated it silently; Weasel 9.25 validates primary key constraint names ([weasel#448](https://github.com/JasperFx/weasel/issues/448)) and refuses to emit one that would be cut.

Baselined both ways: `tenant_partitioned_natural_key_aggregate` passes on Weasel 9.24.0 and fails on 9.25.0. Fixed upstream in [JasperFx/marten#5271](https://github.com/JasperFx/marten/pull/5271); verified end-to-end by pointing `MartenTests` at a locally-built Marten carrying that commit — 0/3 → **3/3**.

This is not test-only: it affects any `[NaturalKey]` aggregate with a type name of 16+ characters under conjoined tenancy.

## Merge order

1. Merge and release [JasperFx/marten#5271](https://github.com/JasperFx/marten/pull/5271).
2. Bump the `Marten` pins in `Directory.Packages.props` off 9.23.0.
3. Undraft this and confirm `CIMySql` and `CIMarten` are green.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,43 @@
 
 ## Unreleased
 
+### WolverineFx.MySql
+
+- **The envelope storage migration no longer fails on every startup.**
+  ([#3985](https://github.com/JasperFx/wolverine/pull/3985), closes
+  [#3983](https://github.com/JasperFx/wolverine/issues/3983)) On MySQL every node logged an `Error` from
+  `MySqlMessageStore` on every start: `Cannot drop index 'fk_wolverine_node_assignments_node_id': needed
+  in a foreign key constraint`. The first migration against an empty database is a pure `CREATE`, so this
+  was unreachable until the second run — it reproduced against a brand-new database and then repeated
+  forever.
+
+  There were four drift items, not one, and only the first threw — so the migration runner logged it and
+  the rest never ran. `wolverine_nodes.node_number` and `wolverine_node_records.id` folded `AUTO_INCREMENT`
+  into the column *type* string; the catalog reports the type as plain `INT` and carries `auto_increment`
+  separately, so neither column ever compared equal and both emitted a `MODIFY COLUMN` on every check.
+  They now use Weasel's `AutoIncrement()`. The generated DDL is unchanged apart from where the keyword
+  sits, so **existing databases need no migration**.
+
+  The other three were `Weasel.MySql` bugs, fixed in [weasel#445](https://github.com/JasperFx/weasel/pull/445)
+  and shipped in Weasel 9.25.0.
+
+### Dependencies
+
+- **Weasel bumped to 9.25.0**, the identifier release. If your schema names are all plain lowercase
+  identifiers the emitted DDL is byte-identical to 9.24 and there is nothing to do. Otherwise three
+  changes are worth knowing about, and Weasel's
+  [upgrade notes](https://weasel.jasperfx.net/release-9-25) carry the full list:
+
+  - **Oracle can now see index, foreign key and primary key drift**, which it previously could not detect
+    at all. Expect a first run that applies index and foreign key changes it had been silently ignoring —
+    that is the backlog being worked off, not new drift.
+  - **SQLite no longer drops the table to change a column.** A change that `ALTER TABLE` cannot express
+    used to be answered by dropping the object and recreating it, losing every row.
+  - **Column names are no longer case-folded or rewritten.** Wolverine's NServiceBus PostgreSQL queue
+    table declared PascalCase columns and relied on Weasel folding them to lowercase; a real
+    NServiceBus-provisioned table has lowercase columns, so the declaration was wrong all along and is now
+    stated as lowercase directly. No behavior change against an NServiceBus-owned table.
+
 ### WolverineFx (core)
 
 - **`AfterCommit` — a declarative way to run work after the transactional commit.**
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -137,13 +137,13 @@
     <PackageVersion Include="System.Diagnostics.DiagnosticSource" Version="9.0.5" />
     <PackageVersion Include="System.Net.NameResolution" Version="4.3.0" />
     <PackageVersion Include="System.Threading.Tasks.Dataflow" Version="9.0.5" />
-    <PackageVersion Include="Weasel.Core" Version="9.24.0" />
-    <PackageVersion Include="Weasel.EntityFrameworkCore" Version="9.24.0" />
-    <PackageVersion Include="Weasel.MySql" Version="9.24.0" />
-    <PackageVersion Include="Weasel.Oracle" Version="9.24.0" />
-    <PackageVersion Include="Weasel.Postgresql" Version="9.24.0" />
-    <PackageVersion Include="Weasel.SqlServer" Version="9.24.0" />
-    <PackageVersion Include="Weasel.Sqlite" Version="9.24.0" />
+    <PackageVersion Include="Weasel.Core" Version="9.25.0" />
+    <PackageVersion Include="Weasel.EntityFrameworkCore" Version="9.25.0" />
+    <PackageVersion Include="Weasel.MySql" Version="9.25.0" />
+    <PackageVersion Include="Weasel.Oracle" Version="9.25.0" />
+    <PackageVersion Include="Weasel.Postgresql" Version="9.25.0" />
+    <PackageVersion Include="Weasel.SqlServer" Version="9.25.0" />
+    <PackageVersion Include="Weasel.Sqlite" Version="9.25.0" />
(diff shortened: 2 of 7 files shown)


jeremydmiller:
Superseded by #3991.

This branch was cut before Weasel 9.25.1 existed and pins 9.25.0, whose identifier validation broke `tenant_partitioned_natural_key_aggregate`. The repo blocks force-push, so the stale pin cannot be rewritten out of this branch's history — #3991 is the same work off a clean base, on 9.25.1, plus the Oracle `ORA-03048` fix that CI surfaced here.

Issue #4026, opened 2026-08-23
Batch incoming durable-inbox persistence on the remaining one-at-a-time transports (Redis Streams, Kafka topic-group, Pulsar, NATS, MQTT)


jeremydmiller:
Follow-up to #3711, which is closing with the completion side done (#4025, O1b batched mark-as-handled) and the step-1 audit posted. This issue carries the audit's remaining **arrival-side** rows — the transports whose durable listeners still hand `DurableReceiver` one `Envelope` at a time, so every message pays its own inbox `INSERT` round trip while RabbitMQ/Kafka/SQS/ASB/Pub-Sub pay per batch.

Audit basis: `main` @ e235a1bef, every `IReceiver.ReceivedAsync(` call site in `src/Transports` (full table on [#3711](https://github.com/JasperFx/wolverine/issues/3711#issuecomment-5386857716)).

## Work, in the recommended order

1. **Redis Streams** — `RedisStreamListener` already reads `count: BatchSize` entries per `XREADGROUP` and then calls `ReceivedAsync(envelope)` once per entry in `ProcessMessage`. Map the read into an `Envelope[]` and hand it over once. Natural batches, no timer, no new knob — the cheapest win here. Keep the per-entry mapping try/catch semantics (a bad entry must not poison its neighbours).
2. **Kafka `KafkaTopicGroupListener`** — the topic-group path (`:97`, `:136`) never got the `KafkaListener` drain-up-to-`MaximumMessagesToReceive` treatment. Mirror it, same gate (`Durable && MaximumMessagesToReceive > 1`).
3. **Pulsar** (`PulsarListener` main + retry consumer loops, `PulsarReaderListener`), **NATS** (`CoreNatsSubscriber`, `JetStreamSubscriber`), **MQTT / MQTT5** (`MqttListener`) — per-message consume loops. Add the Rabbit pattern: `BatchingChannel<Envelope>` gated on `Mode == Durable && <batch size> > 1`, 5 ms max-age window, flush on dispose/latch like `WorkerQueueMessageConsumer`, and the same nack-the-batch fallback if `ReceivedAsync(Envelope[])` throws. Each of these needs a `MaximumMessagesToReceive`-style knob on its endpoint (Pulsar/NATS/MQTT don't have one today).

[...]

PR #4031, opened 2026-08-23
GH-4028: make UseDurableInbox() on a Redis stream a real durable inbox

### LINKED ISSUE #4028: Redis Streams: UseDurableInbox() acks on receipt and never writes the inbox — IDatabaseBackedEndpoint makes it at-most-once, not durable
Found while auditing arrival-side inbox batching for #4026 (the Redis row there is withdrawn — there is no INSERT to batch). Filing separately because this is a correctness/semantics problem, not a perf item.

## What the code does

`RedisStreamEndpoint` implements `IDatabaseBackedEndpoint` (`src/Transports/Redis/Wolverine.Redis/Internal/RedisStreamEndpoint.cs:14`, implementation at `:405` — `ScheduleRetryAsync` re-adds the serialized envelope to the stream for native scheduled retries). That marker has two effects inside `DurableReceiver`, and Redis only wanted one of them:

1. **Native scheduled retry** — `_scheduleExecution` routes to `db.ScheduleRetryAsync` instead of the inbox (`DurableReceiver.cs:201-205`). Intended.
2. **No inbox persistence, and complete-on-receipt** — `ShouldPersistBeforeProcessing = !(endpoint is IDatabaseBackedEndpoint)` (`DurableReceiver.cs:52`), so the arrival `INSERT` is skipped entirely, and the receiver then enqueues the envelope **and immediately posts it to `_completeBlock`** (`:644-655`), which on Redis is `RedisStreamListener.CompleteAsync` → `XACK` (`Internal/RedisStreamListener.cs:223-244`).

For the database transports that marker was written for, that's correct: `PostgresqlQueueListener` moves the row from the queue table *into the incoming table* in the same pop transaction (`PostgresqlQueueListener.cs:206-215`), so the message is durable before the delete, and "complete" just removes the queue row. Redis does none of that. Net effect of `ListenToRedisStream(...).UseDurableInbox()`:

- the message is **XACK'd before the handler runs** — an ungraceful crash mid-handler loses it (at-most-once), exactly the Buffered semantics the mode name says it avoids;
- **nothing is ever written to the inbox** — the Redis tests run `UseDurableInbox()` with **no message store configured at all** (`Wolverine.Redis.Tests/EndToEndRetryTests.cs:44`, `NativeSchedulingRetryTests.cs`) and pass only because of this;
- with a real store configured (`global_partitioned_sharded_processing.cs` uses Postgres), every completion still issues a mark-as-handled `UPDATE` against a row that was never inserted — a wasted round trip per message (coalesced by #4025, still wasted).

The docs present it as the durable option: `docs/guide/messaging/transports/redis.md:397` — "Use durable inbox so retry messages are persisted"; `:65`, `:165`. What the mode actually buys on Redis is "scheduled retries go back into the stream natively", which is a real feature but not durability.

## Options (decision needed)

A. **Make it honest**: keep `IDatabaseBackedEndpoint` (native retry) but stop calling the mode durable — e.g. rename/alias to something like `UseNativeScheduledRetries()` and make `UseDurableInbox()` on a Redis listener either throw with guidance or behave as option B. Docs updated to say Redis streams + consumer-group PEL is the at-least-once story only while the entry is un-ACK'd, and that Wolverine ACKs on receipt in every mode today except Inline.

B. **Make it durable**: drop `IDatabaseBackedEndpoint` from `RedisStreamEndpoint` so Durable mode does the real inbox `INSERT` before ACK (at-least-once via the inbox, dedup by envelope id), and keep native scheduled retries via a different seam — `DurableReceiver` already implements `ISupportNativeScheduling`/`MoveToScheduledUntilAsync` through the listener; `RedisStreamListener` could own the "re-add to stream at time T" behaviour there. Requires a message store to be configured for Durable (like every other transport) — the existing Redis "durable" tests would need one, which is the honest state anyway. This is the behaviour the docs already promise.

C. **Cheapest correct-ish**: keep everything but don't ACK on receipt for Redis — i.e. give `IDatabaseBackedEndpoint` a second axis (`PersistsIncomingItself`) so Redis gets native retry + ACK-after-handler through the PEL. Still no inbox, so no dedup and no outbox atomicity, but at-least-once on crash and no wasted `UPDATE`s.

I'd lean B — it's what `UseDurableInbox()` means everywhere else and what the docs say — with A's docs honesty as the fallback if the native-retry seam is a bigger refactor than it looks. Not starting it without a call on which.

## Repro sketch

Redis stream listener with `UseDurableInbox()`, a handler that blocks, send one message, kill the process: the entry is gone from the consumer group's PEL (`XPENDING` empty) and nothing is in any inbox table. Compare `ProcessInline()`, where `XPENDING` shows it until the handler completes.

jeremydmiller:
Closes #4028 (option B). Found while auditing arrival-side batching for #4026.

## The defect

`RedisStreamEndpoint` implemented `IDatabaseBackedEndpoint` to get Redis-native scheduled retries out of `DurableReceiver`. That marker does two things in `DurableReceiver`, and Redis only wanted one:

1. routes scheduled retries to `endpoint.ScheduleRetryAsync` (intended);
2. sets `ShouldPersistBeforeProcessing = false` and **completes the delivery on receipt** — which for Redis is `XACK` before the handler runs, with **no inbox write at all**.

So `ListenToRedisStream(...).UseDurableInbox()` was at-most-once on a crash — the exact semantics the mode name says it avoids — and the Redis test suite ran "durable" listeners with **no message store configured**, which only worked because of this. With a real store configured, every completion also issued a mark-as-handled `UPDATE` against a row that never existed.

## The fix

- **`RedisStreamEndpoint` is no longer `IDatabaseBackedEndpoint`.** A Durable Redis listener now behaves like every other durable transport: `INSERT` into the inbox **before** the stream entry is acknowledged, handled from the inbox, scheduled retries parked in the inbox. It requires a configured message store, as everywhere else.
- **Redis-native scheduled retries are kept for the non-durable modes** via `RedisStreamListener : ISupportNativeScheduling` (`NativeSchedulingEnabled => Mode != Durable`): park the copy in the stream's scheduled sorted set, then ACK the original (that order so a crash in between costs a duplicate, not a loss). This is strictly better than before for **Buffered** (was in-memory scheduling, lost on crash) and **Inline** (was the inbox, which needed a store). `RedisStreamEndpoint.ScheduleRetryAsync` stays as the method behind it, with docs explaining why.
- A Durable listener deliberately reports `NativeSchedulingEnabled == false`: `MessageContext.ReScheduleAsync` prefers a listener-level rescheduler and **never completes the envelope** in that path, so re-adding the same envelope id to the stream would leave the inbox row orphaned and the redelivered copy would be discarded as a duplicate. The inbox is the only correct place for a durable scheduled retry.

## Tests

| | Result |
|---|---|
| **new `durable_inbox_is_real_4028`** (Redis + Postgres inbox): (a) with the handler parked, the inbox shows the row `Incoming` **and** the consumer group's pending list is already empty — durable, then acked; (b) a `ScheduleRetry` lands in the inbox as `Scheduled`, the Redis scheduled sorted set stays empty, and the retry still runs; (c) the endpoint is not `IDatabaseBackedEndpoint` | 3/3 |
| `EndToEndRetryTests`, `rate_limiting_end_to_end` — both assert the Redis-native scheduled set, i.e. the non-durable path, and both used `UseDurableInbox()` with no store; switched to `BufferedInMemory()` (the code path they were really testing) | ✅ |
| `DatabaseBackedEndpointTests`, `NativeSchedulingRetryTests` — now assert the endpoint is **not** database-backed and the listener owns native scheduling; the `ScheduleRetryAsync` sorted-set tests are unchanged | ✅ |
| **full `Wolverine.Redis.Tests`** | **147 / 147** |
| `dotnet build wolverine.slnx -c Release -f net9.0` | clean |

## Docs / behaviour change

New "Durable Inbox" section in `docs/guide/messaging/transports/redis.md` with a warning box describing the pre-6.30 behaviour and the migration: if you were running `UseDurableInbox()` on Redis without a store you now need one; if you actually wanted Redis-native scheduled retries without a database, the default `BufferedInMemory()` (or `ProcessInline()`) listener is that. The scheduling sample no longer pairs `UseDurableInbox()` with "scheduled natively in Redis".

This is a behaviour change for anyone on Redis "durable" today — but the behaviour they had was not durable, so I'd argue it's a bug fix with a loud note rather than a major-version item. Flagging it explicitly so you can disagree.

### DIFF
--- a/docs/guide/messaging/transports/redis.md
+++ b/docs/guide/messaging/transports/redis.md
@@ -360,6 +360,24 @@ public class OurRedisJsonMapper<TMessage> : EnvelopeMapper<StreamEntry, List<Nam
 <sup><a href='https://github.com/JasperFx/wolverine/blob/main/src/Transports/Redis/Wolverine.Redis.Tests/DocumentationSamples.cs#L230-L291' title='Snippet source file'>snippet source</a> | <a href='#snippet-sample_ourredisjsonmapper' title='Start of snippet'>anchor</a></sup>
 <!-- endSnippet -->
 
+## Durable Inbox <Badge type="tip" text="6.30" />
+
+`UseDurableInbox()` on a Redis stream listener means exactly what it means on every other transport:
+each message is written to your configured message store (`PersistMessagesWithPostgresql()` etc.) **before**
+its stream entry is acknowledged, is handled from there, and scheduled retries are parked in the inbox. A
+process crash mid-handler therefore replays the message from the inbox instead of losing it. Because the
+entry is acknowledged as soon as it is durable, the consumer group's pending list stays short regardless of
+how long handlers take.
+
+::: warning
+Before 6.30 the Redis stream endpoint was marked `IDatabaseBackedEndpoint`, which made a "durable" listener
+skip the inbox write entirely and acknowledge the entry **on receipt** — effectively at-most-once on a crash,
+and scheduled retries went to Redis's own scheduled set rather than the inbox. If you were running
+`UseDurableInbox()` on Redis without a message store, that configuration now requires one, the same as for
+RabbitMQ, Kafka or SQS. If what you actually wanted was Redis-native scheduled retries without a database,
+use the default `BufferedInMemory()` (or `ProcessInline()`) listener: those schedule natively in Redis.
+:::
+
 ## Scheduled Messaging <Badge type="tip" text="5.10" />
 
 The Redis transport supports native Redis message scheduling for delayed or scheduled delivery. There's no configuration
--- a/src/Transports/Redis/Wolverine.Redis.Tests/DatabaseBackedEndpointTests.cs
+++ b/src/Transports/Redis/Wolverine.Redis.Tests/DatabaseBackedEndpointTests.cs
@@ -5,6 +5,7 @@ using Shouldly;
 using StackExchange.Redis;
 using Wolverine.Configuration;
 using Wolverine.Redis.Internal;
+using Wolverine.Transports;
 using Wolverine.Runtime;
 using Wolverine.Runtime.Serialization;
 using Wolverine.Util;
@@ -22,7 +23,7 @@ public class DatabaseBackedEndpointTests
     }
 
     [Fact]
-    public async Task redis_endpoint_implements_idatabase_backed_endpoint()
+    public async Task redis_endpoint_is_not_database_backed_and_its_listener_schedules_natively_when_not_durable()
     {
         var streamKey = $"dbe-test-{Guid.NewGuid():N}";
         using var host = await Host.CreateDefaultBuilder()
@@ -38,10 +39,13 @@ public class DatabaseBackedEndpointTests
         var transport = runtime.Options.Transports.GetOrCreate<RedisTransport>();
         var endpoint = transport.StreamEndpoint(streamKey);
 
-        // Verify it implements IDatabaseBackedEndpoint
-        endpoint.ShouldBeAssignableTo<IDatabaseBackedEndpoint>();
+        // GH-4028: IDatabaseBackedEndpoint made DurableReceiver skip the inbox INSERT and ACK on receipt,
+        // which is the opposite of what UseDurableInbox() promises. The Redis-native scheduled retry now
+        // lives on the listener, and only for the non-durable modes.
+        endpoint.ShouldNotBeAssignableTo<IDatabaseBackedEndpoint>();
+        typeof(ISupportNativeScheduling).IsAssignableFrom(typeof(RedisStreamListener)).ShouldBeTrue();
 
-        _output.WriteLine("✓ RedisStreamEndpoint implements IDatabaseBackedEndpoint");
+        _output.WriteLine("✓ RedisStreamEndpoint is no longer IDatabaseBackedEndpoint; RedisStreamListener is ISupportNativeScheduling");
     }
 
     [Fact]
@@ -59,7 +63,7 @@ public class DatabaseBackedEndpointTests
 
         var runtime = host.Services.GetRequiredService<IWolverineRuntime>();
         var transport = runtime.Options.Transports.GetOrCreate<RedisTransport>();
-        var endpoint = transport.StreamEndpoint(streamKey) as IDatabaseBackedEndpoint;
+        var endpoint = transport.StreamEndpoint(streamKey);
         var database = transport.GetDatabase(database: 0);
         var scheduledKey = transport.StreamEndpoint(streamKey).ScheduledMessagesKey;
 
@@ -123,7 +127,7 @@ public class DatabaseBackedEndpointTests
 
         var runtime = host.Services.GetRequiredService<IWolverineRuntime>();
         var transport = runtime.Options.Transports.GetOrCreate<RedisTransport>();
-        var endpoint = transport.StreamEndpoint(streamKey) as IDatabaseBackedEndpoint;
+        var endpoint = transport.StreamEndpoint(streamKey);
         var database = transport.GetDatabase(database: 0);
         var scheduledKey = transport.StreamEndpoint(streamKey).ScheduledMessagesKey;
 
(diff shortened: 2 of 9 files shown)


jeremydmiller:
Reviewed. The fix is right and the reasoning in the description holds up — one real defect, which CI will not catch.

## `ClearAllWolverineStorageAsync` silently stops resetting Redis streams

`StorageExtensions.cs:39-43` selects the queues it resets with exactly the marker this PR removes:

```csharp
var queues = transport.Endpoints()
    .OfType<IBrokerQueue>()
    .Where(x => x is IDatabaseBackedEndpoint)
    .ToArray();

if (queues.Length == 0) continue;
```

With `RedisStreamEndpoint` no longer `IDatabaseBackedEndpoint`, that array is empty for `RedisTransport`, the whole transport is `continue`d, and Redis stream endpoints never get their `SetupAsync`/`PurgeAsync`. `StorageExtensions.cs` isn't in the diff.

Three things make this worth fixing rather than just noting:

* **The comment immediately above it becomes false.** `StorageExtensions.cs:37` enumerates the covered transports as "PostgreSQL, SQL Server, MySQL, Oracle, SQLite, **and Redis streams**".
* **Nothing catches it.** There is no Redis implementation of `ClearAllWolverineStorageCompliance` — only Sqlite, Oracle, Postgresql, SqlServer and MySql — and no Redis test calls the helper. This can go green across all 37 checks with the regression in.
* **`docs/guide/testing.md:835` presents that helper as *the* integration-test reset**, so a Redis user following the docs silently starts carrying stream entries between runs.

The root cause is the same shape as the bug this PR fixes: `IDatabaseBackedEndpoint` carries **three** unrelated jobs on one marker — the native-retry route (`DurableReceiver.cs:201-205`), the skip-the-inbox/complete-on-receipt switch (`DurableReceiver.cs:52`), and this reset selector. The PR correctly separates the first two and inherits the third by accident. Fix is either a distinct marker meaning "has queue state to reset", or naming `RedisStreamEndpoint` explicitly in `StorageExtensions` — plus correcting that comment either way.

## What I checked and found correct

* **The new test genuinely fails without a store.** This was my main concern going in, since a Wolverine host with no store configured silently resolves `NullMessageStore` — which is precisely how the old "durable" Redis tests asserted durable behaviour while proving nothing. `durable_inbox_is_real_4028` configures `PersistMessagesWithPostgresql` and resolves `IMessageStore` to assert on inbox rows directly, so it cannot pass against `NullMessageStore`. That is the load-bearing part of this PR.
* **Connection string comes from `Servers.PostgresConnectionString`**, not a literal — no hardcoded 5432 to fail in CI.
* **The Buffered path is actually covered**, not just implied by the Durable one. `EndToEndRetryTests` and `rate_limiting_end_to_end` move to `BufferedInMemory()` and assert the Redis scheduled sorted set, exercising the new `ISupportNativeScheduling` seam on the listener. Worth calling out because `DurableReceiver` and `BufferedReceiver` implement that interface separately.
* **The ordering in `MoveToScheduledUntilAsync` is the right way round** — park in the sorted set, then `CompleteAsync`, so a crash between them costs a duplicate rather than a loss. The inline comment says so, which is the kind of thing that gets silently "tidied" later without one.
* **`NativeSchedulingEnabled => Mode != Durable` is justified** and the stated reason holds: a listener-level reschedule never completes the envelope, so re-adding the same envelope id to the stream would orphan the inbox row and the redelivery would be discarded as a duplicate.

## On the behaviour-change question

Agree with the framing in the description. What users had was not durable — it ACKed before the handler ran — so restoring the documented meaning is a bug fix, not a breaking change to a working feature. The loud docs warning is the right treatment. The one group who genuinely lose something are those running `UseDurableInbox()` on Redis with **no** store purely to get native scheduled retries; the migration note pointing them at `BufferedInMemory()` covers that, and they now get better crash behaviour than before.

CI was 37/37 pending when I looked, so this is a read of the diff rather than of a green run.

Issue #4012, opened 2026-08-22
Ack/nack retry budget and terminal-failure classification (shared AckAttempts, per-transport transient detection)


jeremydmiller:
## Context

Filing the residue of a field-report analysis (`ACK-RETRY-CLASSIFICATION-ANALYSIS.md`, repo root) that was never turned into an issue. The original incident — a `NullReferenceException` in `RabbitMqListener.CompleteAsync` looping ~4/min through eleven deploys alongside a steady "Duplicate incoming envelope detected" stream — **has been fixed**. What remains is the generalization.

Surfacing it now because #3708 (`EndpointMode.NativeAck`) makes settle reliability materially more load-bearing: with no inbox row, a settle that silently exhausts its retry budget is a redelivery and a duplicate execution, not a no-op.

## Already shipped — do NOT rebuild

Verified against `main` @ `a4b6f5bd9`:

- **The reported NRE.** `RabbitMqListener.CanSettle` (`:487-495`) guards on `ReferenceEquals(channel, envelope.DeliveredOn) && channel.IsOpen`, so an ack racing a channel teardown is skipped rather than dereferencing null.
- **The dead string match.** The needle carried a trailing apostrophe that never matched real broker text (`'PRECONDITION_FAILED - unknown delivery tag 1'` — the tag number sits *inside* the quotes), so the terminal-failure branch was unreachable and every unknown tag burned three pointless retries. Now a shared `UnknownDeliveryTag` constant matched without the closing quote (`RabbitMqChannelCallback.cs:21-26`), used at both `:42` and `:132`.
- **GH-3950** added `QuiesceAfterRejectedSettle` on both paths — once the broker has closed a channel, stop feeding it.

### 1. `Envelope.AckAttempts` — a retry budget shared across stacked blocks

`RetryBlock.MaximumAttempts` already bounds retries *within* one `PostAsync`. It cannot bound them **across** posts, and the durable path stacks two blocks:

`DurableReceiver._completeBlock` (`DurableReceiver.cs:87`) → `Listener.CompleteAsync` → `RabbitMqChannelCallback.Complete` (`RabbitMqChannelCallback.cs:31`)

That is 3 x 3 = **9** attempts, and neither block can see the other's count. `AckAttempts` gives them one budget per delivery. Grep confirms no such member exists today.

Shape mirrors `SendAttempts` (`Envelope.cs:246`), reset alongside it in `Envelope.Reset()` (`Envelope.Internals.cs:490`).

[...]

Issue #3711, opened 2026-07-29
Audit/extend batched incoming-envelope persistence for durable endpoints (remaining transports + batched mark-as-handled)


jeremydmiller:
## Context

Lower-priority companion to #3708. The original ask — "opt into batching the persisting of incoming envelopes instead of writing and acking one envelope at a time" — turned out to be **mostly shipped already** in the GH-3492 perf wave. This issue is the audit/extension of that work, not a new design.

What already exists:

- **RabbitMQ**: `WorkerQueueMessageConsumer` builds a `BatchingChannel<Envelope>` when `Mode == EndpointMode.Durable && MaximumMessagesToReceive > 1` (default **100**, 5ms max-age window — `src/Transports/RabbitMQ/Wolverine.RabbitMQ/Internal/WorkerQueueMessageConsumer.cs:37-42`), flushing into `DurableReceiver.ReceivedAsync(listener, Envelope[])` → one transaction / one round trip via `StoreIncomingAsync(IReadOnlyList<Envelope>)` (`src/Persistence/Wolverine.RDBMS/MessageDatabase.Incoming.cs:174-229`). Measured **1,086/s → 3,101/s (+186%)** max durable throughput. Broker acks remain per-message even on the batched path.
- **Kafka** gates the same way (`src/Transports/Kafka/Wolverine.Kafka/Internals/KafkaListener.cs:88-90`).
- Batch-receive transports (SQS, ASB, Pub/Sub) already flow arrays into the same `Envelope[]` overload.
- Duplicate handling on a failed batch insert falls back to one-at-a-time redrive (`DurableReceiver.cs:726-737`).

Explicit non-goal: broker **ack** coalescing. Prototyped in the GH-3492 wave and REJECTED at −10% throughput — `basic.ack` is a fire-and-forget frame, not an RPC (`RABBITMQ-PERF-DEEP-DIVE-PLAN.md:176-181`, ledger row 216). Do not revisit.

## Execution plan

[...]

PR #4039, opened 2026-08-23
GH-4026: stop the NATS durable redelivery storm (9.5x max durable throughput)

jeremydmiller:
Follow-up to #4029's NATS lead. The lead as written ("~1.3 ms of awaited ack per message; try `AckAll`") was **wrong** — NATS.Net 2.8 publishes acks fire-and-forget, and per-ack cost was never the ceiling. Instrumenting a run at 1 Hz against the JetStream consumer stats and correlating with the listener log showed what actually happens, repeatedly, on every back-pressure pause:

```
delivered frozen for 30s ┐         ack_pending pinned at 1,000 (= MaxAckPending)
                         ├─ AckWait expires → rd=793 redelivery burst → ~1,000 duplicate inserts
restarted listener idle ─┘
```

## Three defects

1. **Stop dropped the drain's acks.** `NatsListener.StopAsync` cancelled the listener's token *before* draining, and `RetryBlock` silently no-ops on a cancelled token — every ack issued while the stop flushed the durable receive window was discarded. Each one became a redelivery `AckWait` (30 s) later.
2. **1,000-message orphan window.** The client pre-fetches up to 1,000 messages (`ConsumeAsync` `MaxMsgs` default), all counting against the consumer's `MaxAckPending` (server default 1,000) until acked. After a stop, the orphans pinned that cap: the **restarted** listener received *nothing* for the full `AckWait`, then ate ~1,000 duplicate inserts in a burst — and a single duplicate fails the whole batched `INSERT` down to the one-at-a-time redrive.
3. **Persisted-but-unacked tail.** `DurableReceiver` acked each envelope only *after* `EnqueueAsync` onto the bounded (10,000) worker block. Once the queue filled, the tail of an already-persisted batch sat un-acked behind it. A persisted message needs nothing else to be safe — settle first, then enqueue. (Generic: applies to every durable transport.)

## The fixes

- `NatsListener.StopAsync`: dispose the subscriber first (ends deliveries, flushes the durable window into the inbox) **while the token is live so the acks go out**; cancel last. The subscriber gains its own consume-loop `CancellationTokenSource`, because cancelling the token is the *only* thing that ends an active `ConsumeAsync` — `INatsJSConsumer` is a metadata handle, and relying on "disposing" it left a stopped listener happily consuming another 7 million messages during one experiment. The consume-task wait at disposal is bounded (30 s) so teardown can never hang.
- `JetStreamSubscriber`: a **Durable** endpoint bounds the pull batch to `MaximumMessagesToReceive` (default 100), shrinking the maximum orphan window 10×. Buffered/Inline keep the client default (they ack on receipt; bounding them costs top-end throughput for nothing — measured 62 k → 30 k when bounded, unchanged at 64 k without).
- `DurableReceiver` (single + batch paths): settle the broker delivery **before** enqueueing for execution, with the reasoning in comments.

## Measured

Rig `n-max-durable` (NATS + PostgreSQL, uncapped inline publisher, 0 ms handler, 15 s warmup / 45 s measured), **both sides against a freshly dropped rig schema**:

| Build | Runs (msg/s) | Duplicates / run | Stop-starve cycles |
|---|---|---|---|
| `main` (post-#4029) | 521.2 / 429.1 — mean 475 | ~3,800 | 3 |
| **this PR** | **4,531.3 / 3,189.8 / 5,880.7 — mean 4,534** | **0** | pauses still occur (14–20) and are now harmless |

**+9.5× mean, no overlap.** `n-max-buffered` unchanged (64,184/s). This also retroactively explains the modest +35 % NATS number in #4029: the batching win there was real but capped by the storm; with the storm gone the durable inbox batching finally shows Kafka/Pulsar-class gains.

### Rig housekeeping that made the numbers honest
- The per-run JetStream stream is now deleted on consumer shutdown — **67 M leftover messages / 101 GB** had accumulated in the broker container across the day's cells and were degrading every later run.
- `rig.sh` drops the rig's Postgres schema per run — **6 M leftover inbox rows** made Wolverine's startup recovery (`update … set owner_id = 0`) time out before the consumer host even started.

## Tests

| | Result |
|---|---|
| full `Wolverine.Nats.Tests` (incl. `durable_jetstream_batching_4026`, JetStream compliance) | **153/153** |
| full `CoreTests` — includes the GH-3711 mark-as-handled and GH-4012 ack-budget suites over the reordered settle path | **2,523 / 0** |
| `dotnet build wolverine.slnx -c Release -f net9.0` | clean |

No ack-semantics changes: acks are still explicit and per message; only *when* they are issued (immediately after persistence) and how teardown treats the ones in flight changed.

### DIFF
--- a/src/Testing/KafkaPerfRig/WolverineNats.cs
+++ b/src/Testing/KafkaPerfRig/WolverineNats.cs
@@ -21,7 +21,16 @@ public static class WolverineNats
         RigHandlerSettings.SequenceByGame = cfg.Sequencing == "semaphore";
 
         var builder = Host.CreateDefaultBuilder()
-            .ConfigureLogging(logging => logging.SetMinimumLevel(LogLevel.Warning))
+            .ConfigureLogging(logging =>
+            {
+                logging.SetMinimumLevel(LogLevel.Warning);
+                // RIG_LOG_LISTENER=1: see back-pressure pauses / restarts (logged at Information)
+                if (Environment.GetEnvironmentVariable("RIG_LOG_LISTENER") == "1")
+                {
+                    logging.AddFilter("Wolverine.Transports.ListeningAgent", LogLevel.Information);
+                    logging.AddFilter("Wolverine.Runtime.BackPressureAgent", LogLevel.Information);
+                }
+            })
             .UseWolverine(opts =>
             {
                 opts.Durability.Mode = DurabilityMode.Solo;
@@ -50,6 +59,11 @@ public static class WolverineNats
 
         await host.WaitForShutdownAsync();
 
+        // The max-throughput cells leave millions of unconsumed messages in the per-run stream (the
+        // publisher is ~100x faster than a durable consumer); without this they accumulate across runs
+        // (67M messages / 101 GB after one afternoon) and slow the server down for every later cell.
+        await deleteRunStreamAsync(cfg);
+
         StageRecorder.Dump(cfg.OutDir, "nats-consumer", new
         {
             harness = "wolverine-nats",
@@ -62,6 +76,21 @@ public static class WolverineNats
             maxParallel = cfg.MaxParallel,
             maxReceive = cfg.MaxReceive
         });
+    }
+
+    private static async Task deleteRunStreamAsync(RigConfig cfg)
+    {
+        try
+        {
+            await using var connection = new NATS.Client.Core.NatsConnection(new NATS.Client.Core.NatsOpts { Url = cfg.NatsUrl });
+            var js = new NATS.Client.JetStream.NatsJSContext(connection);
+            await js.DeleteStreamAsync(cfg.NatsStream);
+            Console.WriteLine($"[rig] deleted JetStream stream {cfg.NatsStream}");
+        }
+        catch (Exception e)
+        {
+            Console.WriteLine($"[rig] could not delete JetStream stream {cfg.NatsStream}: {e.Message}");
+        }
     }
 
     private static void configureListener(NatsListenerConfiguration listener, RigConfig cfg)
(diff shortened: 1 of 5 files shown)


jeremydmiller:
## Settle-before-enqueue re-measured on the other durable transports

Since the `DurableReceiver` reorder in this PR applies to every durable transport, here is the A/B on the Rabbit, Kafka and Pulsar `*-max-durable` cells. Methodology: **two separate worktrees** — `origin/main` (post-#4029) vs this branch — each built independently (no stashing), two interleaved runs per side, the rig's Postgres schema dropped before every run, zero duplicate inserts on either side of any cell.

| Cell | `main` (msg/s) | this PR (msg/s) | Verdict |
|---|---|---|---|
| `r-max-durable` (RabbitMQ) | 3,972.5 / 2,707.8 — mean 3,340 | 4,002.9 / 3,858.0 — mean 3,930 | neutral → slightly positive |
| `max-durable` (Kafka) | 3,754.5 / 4,022.3 — mean 3,888 | 3,578.4 / 4,118.3 — mean 3,848 | flat (overlapping) |
| `p-max-durable` (Pulsar) | 6,201.6 / 5,801.2 — mean 6,001 | 6,366.3 / 6,146.1 — mean 6,256 | flat (overlapping) |

As expected: the reorder is transformative only where a flooded worker queue was blocking the ack path behind `EnqueueAsync` — which on this box only NATS's delivery pattern provokes (the +9.5× in the PR body). Rabbit acks are fire-and-forget frames, Kafka commits through the offset watermark, and Pulsar acks through its batched ack handler, so none of them were paying the NATS failure mode; for them the change is a semantic hardening (a persisted message can no longer sit un-acked behind a full queue during a back-pressure stop) at no measurable cost.

Two footnotes for the ledger:
- The Pulsar mean here (~6,000 on `main`) is above the 4,172 reported in #4029 — that table's cells ran while the rig's Postgres schema was accumulating rows across the day (6M by evening), which this PR's `rig.sh` now drops per run. #4029's *relative* +447% stands (both sides interleaved under the same pollution); its absolute numbers were depressed.
- The Rabbit `main`-side spread (3,973 vs 2,708) is the box's real noise band; single-run deltas smaller than that band are not findings.

Issue #3712, opened 2026-07-29
Validate or warn on silently-ignored listener configuration combos (Inline + partitioning/parallelism)


jeremydmiller:
## Context

Configuration-hygiene companion to #3708. Auditing listener configuration for the native-ack mode design surfaced several combos that are accepted silently but do nothing — or worse, do different things depending on call order. #3708 makes some of these combos *meaningful* for the new mode, which is the right moment to make the rest loudly invalid.

## The findings (verified against source)

1. **`Inline` + `PartitionProcessingByGroupId(...)` silently no-ops.** `GroupShardingSlotNumber` is read only by `BufferedReceiver` (`src/Wolverine/Runtime/WorkerQueues/BufferedReceiver.cs:62-64`) and `DurableReceiver` (`DurableReceiver.cs:81-83`); `InlineReceiver` ignores it entirely. A user calling `.ProcessInline().PartitionProcessingByGroupId(PartitionSlots.Five)` gets single-threaded processing with no grouping semantics and no warning. Only a doc comment (`src/Wolverine/Configuration/Endpoint.cs:214-218`) states the limitation.
2. **Order-dependent `ProcessInline()` clamp.** `ProcessInline()` sets `MaxDegreeOfParallelism = 1` (`src/Wolverine/Configuration/ListenerConfiguration.cs:382-390`). Consequence: `.MaximumParallelMessages(20).ProcessInline()` ends with `MaxDegreeOfParallelism == 1`, while `.ProcessInline().MaximumParallelMessages(20)` ends with `20` — which nothing reads, but the diagnostics table (`src/Wolverine/WolverineSystemPart.cs:174`) displays as if live. Same two calls, different endpoint state depending on sequence.
3. **`Inline` + `Sequential()` / `MaximumParallelMessages(n)` silently no-op** (only the block constructors read `MaxDegreeOfParallelism`). The only place stating this is a Kafka doc line (`docs/guide/messaging/transports/kafka.md:1309-1310`).
4. **`Inline` + `ExclusiveNodeWithParallelism(20)`** sets `MaxDegreeOfParallelism = 20`, `ListenerCount = 1` — for an Inline endpoint that means fully sequential, the opposite of the method's stated intent.
5. **`Inline` + `BufferingLimits`** is accepted but inert (`Endpoint.ShouldEnforceBackPressure()` at `Endpoint.cs:662-665` prevents the `BackPressureAgent` from being created — `ListeningAgent.cs:99-103`).

[...]

Issue #3186, opened 2026-06-22
Pulsar: finish DLQ sender + resolve transport-vs-endpoint default TODOs


jeremydmiller:
Part of the **Re-Evaluate Pulsar Integration** epic #3176.

## Problem
Two loose ends in the Pulsar transport leave the dead-letter path half-wired:
- `src/Transports/Pulsar/Wolverine.Pulsar/PulsarEndpoint.cs:118` — the dead-letter **sender** is a stub TODO.
- `src/Transports/Pulsar/Wolverine.Pulsar/PulsarTransport.cs:31` — open TODO on whether DLQ/retry defaults should live at the transport level or only per-endpoint.

## Proposed work
- Implement a real DLQ sender so `DeadLetterTopic` in `Native` mode produces failed envelopes to the configured (or `{topic}-DLQ` default) topic with the existing exception metadata stamping (`PulsarEnvelopeConstants.Exception`, `DeadLetterQueueConstants.StampFailureMetadata`).
- Decide and implement transport-level DLQ/retry defaults vs per-endpoint override precedence (mirror the Kafka `DeadLetterQueueTopicName(...)` transport default + per-endpoint override shape).
- Consider an `ExternallyOwned()` analogue (Kafka #3104) so externally-managed DLQ topics skip create/delete.

## Acceptance criteria
- Native-mode DLQ delivers a failed message to the DLQ topic end-to-end, with exception headers, in an integration test.
- Transport-level default + per-endpoint override precedence is covered by a test.
- No remaining `TODO` at the two referenced lines.

**Sequencing:** foundational, do first. Independent of other children.

PR #4074, opened 2026-08-24
GH-4060: warn when a Pulsar hot-tail listener carries requeue policy

### LINKED ISSUE #4060: Pulsar hot-tail listeners silently drop deferred messages in every mode
## Summary

A Pulsar hot-tail listener (`TailFromLatest()`) reads through a non-durable `Reader` with no cursor, so **both** of its settlement operations are deliberate no-ops (`src/Transports/Pulsar/Wolverine.Pulsar/PulsarReaderListener.cs:74-78`):

```csharp
// A Reader cursor is non-durable and unacknowledged: there is nothing to complete or defer.
public ValueTask CompleteAsync(Envelope envelope) => ValueTask.CompletedTask;

public ValueTask DeferAsync(Envelope envelope) => ValueTask.CompletedTask;
```

`CompleteAsync` being a no-op is correct and harmless — there is genuinely nothing to acknowledge. **`DeferAsync` being a no-op is not harmless.** Defer is the "hand this back, try it again" terminal, and on a hot-tail listener it does nothing at all: the message is simply gone. No redelivery, no retry, no trace beyond whatever the error policy logged on its way past.

## Scope, and what is already handled

#4047 / PR #4057 refuses `TailFromLatest()` together with `ProcessInParallelWithNativeAcks()`, because `EndpointMode.NativeAck`'s entire premise is that an unsettled delivery comes back and hot-tail cannot deliver that. That rejection is right and is not what this issue is about.

**The hole is still open for `Inline` and `BufferedInMemory` on a hot-tail listener**, which are the modes people actually use with it today. A handler that throws, exhausts its retries, or hits a `RequeueContinuation` routes through `DeferAsync` and the message evaporates. There is no mode in which a hot-tail listener redelivers anything.

## Why this is defensible-but-surprising

Hot-tail is explicitly a "read the live tail, do not care about history" feature, and someone choosing it has already accepted a non-durable cursor. Losing a message on a *broker restart or a missed window* is inherent to that choice and fine.

Losing a message because **its handler failed** is a different thing, and nothing in the API says so. `TailFromLatest()` reads as a statement about where the cursor starts, not as an opt-out of error handling.

## Options

1. **Say it out loud.** Warn at bootstrap when a hot-tail listener has any error-handling policy that can defer or requeue, since none of it can work. Cheapest, and consistent with GH-3712's principle that a silently-ignored setting should either work, throw, or warn.
2. **Document it** on the Pulsar transport page and in the `TailFromLatest()` xml-docs — "failures on a hot-tail listener are dropped, not retried" — which is currently stated nowhere.
3. **In-lane retry only.** Retries that never leave the process (`RetryNow`, in-memory backoff) are still meaningful on a hot-tail listener; it is only broker-side redelivery that cannot work. Making the distinction explicit in the docs would let people configure something that actually functions.
4. Consider whether the native dead-letter path still works here — a DLQ send is a separate producer rather than a listener settle operation, so it may well function. Worth confirming rather than assuming, because if it does, dead-lettering is the one recovery route hot-tail can actually offer.

Found while implementing #4047. Related: #3708, #3712.

jeremydmiller:
Closes #4060

## What this does

A Pulsar hot-tail listener (`TailFromLatest()`) reads through a non-durable `Reader` that commits no cursor, so `PulsarReaderListener.DeferAsync` is necessarily a no-op. Losing a message to a broker restart or a missed window is inherent to that choice and fine; losing one because **its handler failed** is a different thing, and nothing in the API said so.

Per the issue's options 1 and 2, this is deliberately conservative — no attempt to give hot-tail listeners real redelivery.

**Warn (option 1).** `Endpoint` gains a `supportsRedelivery` hook (default `true`); `PulsarEndpoint` answers `false` when `IsHotTail`. `ListenerConfigurationValidator` warns at bootstrap when such a listener is combined with any requeue-shaped policy — `Requeue()`, `RequeueIndefinitely()`, `PauseThenRequeue()`, `MaximumAttempts()`. Deliberately outside the Inline-only gate: this listener ignores requeue in **every** mode.

Failure policies live on the handler graph, not on an endpoint, so `WolverineRuntime` computes the flag once after `Handlers.Compile()` and passes it in (`FailureRuleCollection.AnyRequeuePolicies()`). This mirrors GH-3712's principle without duplicating #4057's fatal-only `validateModeConfiguration` hook — the two are independent and should merge cleanly.

**Document (option 2).** New "Handler failures on a hot-tail listener are dropped, not retried" section on the Pulsar page with a what-works/what-doesn't table, plus expanded `TailFromLatest()` xml-docs.

## Investigation results (options 3 and 4)

All verified against a live broker, not assumed.

| | Hot-tail listener |
| --- | --- |
| `RetryTimes()`, `RetryWithCooldown()`, `ScheduleRetry()` | ✅ **Work** — the retry never leaves the process |
| `Requeue()` and friends | ❌ Handler runs exactly once, message gone |
| Pulsar native dead letter topic | ❌ **Never reached** |
| Wolverine durable dead letter table | ✅ **Works** with a message store configured |

**Option 3 — in-lane retries work.** Proven: three attempts against the same delivery, ending in a success. This is the error handling to reach for on a hot-tail listener and the docs now say so.

**Option 4 — the native DLQ does *not* work.** A native dead-letter move is a listener operation (`ISupportDeadLetterQueue`), and `PulsarReaderListener` implements neither that nor `ISupportNativeScheduling`; `PulsarEndpoint.TryBuildDeadLetterSender` deliberately returns false (#3186), so `BufferedReceiver` reports no native DLQ either. `MoveToErrorQueue` falls back to `Storage.Inbox.MoveToDeadLetterStorageAsync` — a **silent no-op on `NullMessageStore`**. The test asserting zero DLQ messages ships with a positive control on an ordinary subscription, so the zero means something.

The durable dead letter table *does* catch it, verified with a real Postgres store. That is the only recovery route hot-tail can offer, and the docs say so.

## Separate bug found along the way — not fixed here

`UsePulsar()` registers a **global** `AlwaysMatches` failure rule (`PulsarNativeResiliencyPolicy`) ahead of anything the user configures later. `PulsarNativeContinuationSource.Build` returns `null` for a non-`PulsarListener` envelope, commented "let the next continuation source handle it" — but `FailureRule.TryCreateContinuation` returns `true` regardless, so the null falls through to `?? new MoveToErrorQueue(ex)` and the rule is terminal.

Consequence: on **any** Pulsar-enabled host, `opts.Policies.OnException<T>()` rules never run for an envelope whose listener is not a `PulsarListener`. Measured:

- Hot-tail listener + global `RetryTimes(3)` → 1 attempt, straight to the error queue.
- **A plain local queue in a Pulsar host** + global `RetryTimes(3)` → 1 attempt. Same host with `UsePulsar()` removed → 3 attempts.

Chain-level rules (`chain.OnException<T>()`) still work, because `CombineRules` sorts them ahead of the global ones — which is why the tests here configure retries on the chain. Worth its own issue; the fix has real blast radius for every Pulsar user and does not belong in a warn-and-document change.

## Verification

- `dotnet build wolverine.slnx -c Release -f net9.0` — 0 errors, 0 warnings
- Full CoreTests: **2591 total, 0 failed**
- Full `Wolverine.Pulsar.Tests`: **233 total, 0 failed**
- Red-baselined both ways: disabling the validator check fails 3 of the 12 new CoreTests (and only the warning-asserting ones); flipping `PulsarEndpoint.supportsRedelivery` to `true` fails both Pulsar-side warning tests.

Every new broker test gets its own topic and subscription; no shared static tracking state.

### DIFF
--- a/docs/guide/messaging/transports/pulsar.md
+++ b/docs/guide/messaging/transports/pulsar.md
@@ -111,6 +111,41 @@ A few things to know:
 - Reach for `TailFromLatest()` when you want **all** nodes to process each message; use a normal `Shared` /
   `KeyShared` subscription when you want each message processed **once** across the cluster.
 
+#### Handler failures on a hot-tail listener are dropped, not retried
+
+Losing a message because a node was down or had not attached its reader yet is inherent to choosing a
+non-durable cursor, and is the trade you accept when you call `TailFromLatest()`. Losing one because its
+**handler threw** is easier to be surprised by, so it is worth being explicit about what does and does not
+work here.
+
+A hot-tail listener commits no cursor, so there is nothing for the broker to redeliver from. Every
+requeue-shaped error policy therefore resolves to a no-op and the message is simply gone:
+
+| Error handling | On a hot-tail listener |
+| --- | --- |
+| `RetryTimes()`, `RetryWithCooldown()` | ✅ Works — the retry never leaves the process |
+| `ScheduleRetry()` | ✅ Works — rescheduled in memory, or in the durable inbox if one is configured |
+| `Requeue()`, `RequeueIndefinitely()`, `PauseThenRequeue()`, `MaximumAttempts()` | ❌ **Silently drops the message** |
+| Pulsar native dead letter topic (`DeadLetterQueueing()`) | ❌ Never reached — a native dead-letter move is a listener operation, and this listener is a `Reader` |
+| Wolverine's durable dead letter table | ✅ Works, if a message store is configured |
+
+Wolverine warns at bootstrap if a hot-tail listener is combined with any requeue-shaped policy, since none of
+it can run.
+
+In-process retries are the error handling to reach for on a hot-tail listener, and a configured message store
+is the only recovery route it can offer:
+
+```csharp
+opts.ListenToPulsarTopic("persistent://public/default/live-events").TailFromLatest();
+
+// Retried inside this process; nothing is handed back to the broker.
+opts.Policies.OnException<TransientFailureException>().RetryWithCooldown(
+    50.Milliseconds(), 100.Milliseconds(), 250.Milliseconds());
+```
+
+If these are messages you cannot afford to lose on a handler failure, use an ordinary `Shared` / `KeyShared`
+subscription instead — the cursor is what makes redelivery possible.
+
 ## Replaying a Topic <Badge type="tip" text="6.8" />
 
 When you need to **reprocess** a window of a topic's history — error recovery, rebuilding downstream state,
--- a/src/Testing/CoreTests/Configuration/hot_tail_requeue_policy_validation.cs
+++ b/src/Testing/CoreTests/Configuration/hot_tail_requeue_policy_validation.cs
@@ -1,0 +1,165 @@
+using JasperFx.Core;
+using Shouldly;
+using Wolverine;
+using Wolverine.Configuration;
+using Wolverine.ErrorHandling;
+using Wolverine.Runtime;
+using Wolverine.Transports;
+using Wolverine.Transports.Sending;
+using Xunit;
+
+namespace CoreTests.Configuration;
+
+/// <summary>
+/// GH-4060. A listener whose cursor is ephemeral -- Pulsar's TailFromLatest() is the one transport feature that
+/// does this today -- has nowhere to redeliver a deferred message from, so its DeferAsync is necessarily a no-op.
+/// Requeue-shaped error handling configured against it does nothing at all, which nothing in the API says, so it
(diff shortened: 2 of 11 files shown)


jeremydmiller:
Filed the separate bug found here as #4079, fixed in #4080 (continuation sources can now genuinely decline, and the Pulsar native rule only claims failures it can act on). The two are complementary — #4080 makes error policies run again on a hot-tail listener, and the bootstrap warning here is what catches the requeue-shaped ones that still cannot redeliver. Disjoint files, should merge cleanly either order.

Issue #4100, opened 2026-08-24
CIPulsar wedges in PulsarNativeReliabilityTests — DotPulsar producers stall in Setup, and the test's 1000s timeout outlives the 20-minute job cap


jeremydmiller:
`CIPulsar` wedged twice on 2026-08-24, on unrelated branches, and both times was cancelled at the
20 minute cap rather than failing. The new CI instrumentation (#4084, #4090, #4096) caught the second
one with full evidence, so for once this is not a guess.

There are **two separate problems** here. The second one is a one-character fix and guarantees the
worst possible failure mode for the first.

| | |
|---|---|
| occurrence A | run 32719650836, `gh-4065/pubsub-listener-lifecycle`, 12:30:44 → 12:51:02 = **1218s** |
| occurrence B | run 32754218016, `gh-4083/bobcat-070-test-progress`, 17:15:32 → 17:35:53 = **1221s** |
| conclusion | **`cancelled`** both times — not `failure` |
| `CIPulsar` norm | avg 378s, max 390s over 28 green runs |

A re-run of B went green, so it is intermittent. It is not caused by either PR it appeared on —
occurrence A predates them by five hours, on a branch that touches neither Pulsar nor the build.

## 1. Producers and consumers never finish `Setup`

The stall watchdog fired at 782s and captured the host: **`Threads: 18`, `thread states: 18 S`** —
every thread parked — 3 GB RSS, 442 open fds, and a live socket to the broker on `:6650`.

`dumpasync --coalesce` says what they are parked on. Seven producers:

```
STACKS 3
[14] System.Threading.Tasks.Task<DotPulsar.ProducerState>
  [7] DotPulsar.Extensions.ProducerExtensions+<StateChangedTo>d__1
  [7] TaskFactory+CompleteOnInvokePromise<Task<DotPulsar.ProducerState>>
    [7] DotPulsar.Internal.Producer<ReadOnlySequence<Byte>>+<Monitor>d__31
      [7] DotPulsar.Internal.Executor+<TryExecuteOnce>d__11
        [7] DotPulsar.Internal.Executor+<Execute>d__5
          [7] DotPulsar.Internal.Producer<ReadOnlySequence<Byte>>+<Setup>d__30
```

and three consumers in the mirror image (`Consumer.Setup` → `Executor.Execute` → `Monitor` →
`StateChangedFrom`/`StateChangedTo`). Alongside them, `SubProducer.MessageDispatcher` blocked on
[...]

Issue #4098, opened 2026-08-24
dotnet.yml's build job wedged and was cancelled at its cap — and the logs were discarded, so there is no evidence


jeremydmiller:
The `build` job in `.github/workflows/dotnet.yml` wedged once on 2026-08-24, ran to its cap, and was
cancelled — and because of how a cancelled job is handled, **every piece of evidence about it is
gone.** This is the same class of failure as #4083, in the one workflow that has no instrumentation.

Filing it because the occurrence is real and the next one will be just as unreadable.

## What happened

Attempt 1 of [run 32737221654](https://github.com/JasperFx/wolverine/actions/runs/32737221654), on
PR #4090:

| | |
|---|---|
| job | `build` (97462820888) |
| conclusion | **`cancelled`** — not `failure`, and no step marked failed |
| ran | 14:48:43 → 15:13:43 = **25m00s** |
| `Test` step | started 14:49:01, **never completed** |
| logs | **gone** — `BlobNotFound` |

The 25m00s is worth a note: the job's cap is `timeout-minutes: 20`, and GitHub allows a job up to a
further 5 minutes to wind down after cancellation before force-terminating it. 20 + 5 = exactly the
25m00s observed, which suggests the process did not merely overrun the cap — **it also failed to die
when asked**, using the entire grace window. That is a wedge, not a slow run.

## It is far outside normal

Across the last 25 green runs of this workflow, the `build` job takes:

```
n=25   min 458s   median 616s   max 738s
slowest five: 700, 706, 709, 720, 738
```

So the norm is ~10 minutes and the worst on record is 12m18s, against a 20 minute cap. The wedged
attempt was still going at **24m42s in the Test step alone** — more than double the slowest healthy
run, and it never produced a result.

Not caused by the PR it appeared on: #4090's diff is `.github/workflows/tests.yml` and
`build/ci-memory-sampler.sh`, and this workflow runs `./build.sh ci`, which references neither. A
re-run went green and #4090 merged on 38/38.

## Why this is worth an issue rather than a shrug

It is rare — **0 cancelled runs in the last 100** — but the current setup loses the evidence twice
over:

1. **Cancellation discards the logs.** Not truncates: `BlobNotFound`. Nothing survives except the
   per-step timestamps in the API.
2. **Re-running overwrites the run's conclusion.** That occurrence now reports `success`, because I
   re-ran it. Scanning history for `conclusion == "cancelled"` finds nothing — the only reason this
[...]

PR #4101, opened 2026-08-24
GH-4100: bring the Pulsar reliability test's timeout back inside the CI job cap

jeremydmiller:
One character. Closes the first half of #4100.

`PulsarNativeReliabilityTests.run_setup_with_simulated_exception_in_handler` waited on its tracked
session for `TimeSpan.FromSeconds(1000)` — **16m40s, against this workflow's 20 minute
`timeout-minutes` cap.**

## That timeout can essentially never fire in CI

* the job starts, so the cap lands 20 minutes later
* `CIPulsar` spends ~6.5 minutes building and standing up the broker before the first test runs
* a test starting then has ~13 minutes of budget, and 1000s does not fit

On [run 32754218016](https://github.com/JasperFx/wolverine/actions/runs/32754218016) the arithmetic
was exact:

| | |
|---|---|
| job started | 17:15:32 → cap at **17:35:32** |
| test started | 17:22:11 |
| its timeout would fire | **17:38:51** — nearly 3 minutes *after the job was already dead* |

## Why that matters

The consequence is not a slow test. It is that when the underlying Pulsar flakiness hits — #4100:
seven producers and three consumers stalled in `DotPulsar.Internal.Producer/Consumer.Setup`, all 18
threads parked — **the test cannot fail.** The job runs to the cap and is `cancelled` instead, and a
cancelled job's logs are discarded outright (`BlobNotFound`, #4098), so the failure leaves no
evidence at all.

That has now happened twice, on unrelated branches, three seconds apart in duration (1218s / 1221s).

**This PR does not fix the wedge.** It converts a 20 minute silent cancellation into a 100 second
failure with a tracking dump — the difference between a bug that can be diagnosed and one that
cannot. The stall itself is the other half of #4100.

## Why 100

* every other test in the same file uses `TimeSpan.FromSeconds(100)` — lines 129, 187, 216 — for the
  same shape of assertion
* the retry schedule under test is `4 + 2 + 3` seconds plus a dead-letter hop
* it was the **only** `TrackActivity(TimeSpan.FromSeconds(1000))` in the repository

## Measured, not assumed

The risk in this change is that the test genuinely needs more than 100s. It does not — run against a
local broker on `:6650`:

```
run 1: passed (12s 874ms)
run 2: passed (12s 821ms)
run 3: passed (12s 837ms)
```

~12.8s, very tightly clustered, so **100s leaves roughly 8× headroom** — and 1000s was ~78×.

## Worth considering separately

Should *any* tracked-session timeout be allowed to exceed the job cap? One that does can only ever
produce a cancellation instead of a failure. This was the only 1000s case, but the general rule may
be worth enforcing. Raised in #4100, not addressed here.

Refs #4100. Refs #4098.

### DIFF
--- a/src/Transports/Pulsar/Wolverine.Pulsar.Tests/PulsarNativeReliabilityTests.cs
+++ b/src/Transports/Pulsar/Wolverine.Pulsar.Tests/PulsarNativeReliabilityTests.cs
@@ -69,7 +69,7 @@ public class PulsarNativeReliabilityTests : /*TransportComplianceFixture,*/ IAsy
     [Fact]
     public async Task run_setup_with_simulated_exception_in_handler()
     {
-        var session =  await WolverineHost.TrackActivity(TimeSpan.FromSeconds(1000))
+        var session =  await WolverineHost.TrackActivity(TimeSpan.FromSeconds(100))
             //.WaitForMessageToBeReceivedAt<SRMessage1>(WolverineHost)
             .DoNotAssertOnExceptionsDetected()
             .IncludeExternalTransports()


jeremydmiller:
All five runs finished after the PR was opened — adding the last two for completeness. Passes are unanimous and the spread is 66ms across the set:

```
run 1: passed (12s 874ms)
run 2: passed (12s 821ms)
run 3: passed (12s 837ms)
run 4: passed (12s 808ms)
run 5: passed (12s 808ms)
```

12.81–12.87s against a 100s budget — ~8x headroom, and none of the five came close to the flakiness described in #4100, so the local runs say nothing about how often the stall hits.

Issue #4125, opened 2026-08-25
DoNotAssertOnExceptionsDetected() also suppresses the tracked-session timeout assertion, so a resiliency test passes on a session that achieved nothing


jeremydmiller:
Splitting this out of JasperFx/ai-skills#118, where it surfaced while documenting the tracked-session completion model. That issue only asked the skills to describe what the code does today; this one asks whether the behaviour is what you want.

## What happens

`TrackedSession.Execution.cs:84-97` — the same `AssertNoExceptions` flag guards two different assertions:

```csharp
if (AssertNoExceptions)               AssertNoExceptionsWereThrown();
if (AssertAnyFailureAcknowledgements) AssertNoFailureAcksWereSent();
if (AssertNoExceptions)               AssertNotTimedOut();   // <-- same flag
```

So `DoNotAssertOnExceptionsDetected()` turns off the timeout check as well. A session that times out having completed none of its work returns normally with `Status == TrackingStatus.TimedOut` instead of throwing.

Verified with a throwaway test: an unsatisfiable condition plus a 1-second timeout plus `DoNotAssertOnExceptionsDetected()` returns rather than throwing.

## Why it matters

The flag's own XML doc scopes it to exceptions only:

> Do not assert or fail if **exceptions** where thrown during the message activity. This is useful for testing **resiliency features** and exception handling with message failures.

That is exactly the case that gets hurt. A resiliency test reaches for this flag for its documented reason — handlers are *expected* to throw — and silently also opts out of the only assertion that was going to catch the session timing out. The result is a green test over a session that did nothing, and the symptom is a test that never fails rather than one that fails oddly, so it can sit in a suite indefinitely.

Note the interaction with `IsCompleted()`: because registered conditions *replace* the quiescence check rather than adding to it, a session can already report success while tracked work is still running. Losing the timeout assertion on top of that removes the last thing standing between a misconfigured wait and a permanently green test.

## The wrinkle — the coupling is load-bearing for a second use case

This is not simply a bug to invert, which is why I am filing it as a question rather than a PR.

The suppression is genuinely useful for **firehose debugging**: on a timeout the wait method throws instead of returning, so you never get the `ITrackedSession` and cannot call `AllRecordsInOrder()`. `DoNotAssertOnExceptionsDetected()` is currently the only way to get the session back and inspect what actually happened. That behaviour is relied upon deliberately — our own skills documentation teaches it.

So one flag is serving two unrelated intents:

[...]

PR #4137, opened 2026-08-26
GH-4136: a failed message now reaches a terminal tracking state

### LINKED ISSUE #4136: A message that fails never reaches a terminal tracking state: MessageFailed records MessageEventType.Sent, and the pipeline's last line of defense records nothing
Two independent defects on the same path, found while working #4125.

## 1. `WolverineRuntime.MessageFailed` records the wrong event type

`WolverineRuntime.Tracking.cs:184` and `:200` both do:

```csharp
ActiveSession?.Record(MessageEventType.Sent, envelope, _serviceName, _uniqueNodeId, ex);
```

`Sent` is **not a terminal** `MessageEventType`. `EnvelopeHistory` only completes a `Sent` record once a matching `Received` arrives (`EnvelopeHistory.cs:156-161`), whereas `MessageFailed` completes every record sharing the envelope's `UniqueNodeId` (`:177-190`). Compare the sibling `MessageSucceeded`, which correctly records `MessageEventType.MessageSucceeded`.

Consequence: a message that fails never reaches a terminal state in a tracked session, so any session containing one can only end by **timing out**.

Every caller of `IMessageTracker.MessageFailed` is on a terminal path — `MoveToErrorQueue.cs:64`, the `MessageSucceededContinuation` catch (`:33`), `ApplyItemContinuation`, `ProbeIndividuallyContinuation` — so `Sent` is wrong at all of them. It went unnoticed because `MoveToErrorQueue` emits its own terminal `MovedToErrorQueue` record immediately afterwards, which covers the one path anybody had tests for.

## 2. `HandlerPipeline.RecoverFromFailedProcessingAsync` emits no message event at all

The pipeline's last line of defense (`HandlerPipeline.cs:127-153`) acks the envelope away and calls `tracker.LogException(...)`. `LogException` is free-form diagnostic text, not a message event — no `MessageFailed`, no `Discarded`, no `MovedToErrorQueue`.

So an envelope that dies here — a handler chain that fails to compile at runtime, a sticky-handler misconfiguration where `HandlerGraph.HandlerFor(type, endpoint)` throws before any executor exists — is invisible to every `IMessageTracker`:

* **Failure metrics under-count it.** This is a production observability hole, not just a testing one: the message is gone and nothing incremented.
* **A tracked session never sees it complete**, so the session times out.

## How this was masked

`DoNotAssertOnExceptionsDetected()` also suppressed the tracked-session timeout assertion (#4125). With the timeout assertion suppressed, both defects were invisible: sessions that could only ever time out returned green instead.

Ungating that assertion turned up exactly 3 failures across CoreTests' 2644 tests, and all three traced back here. With both defects fixed, all 2644 pass.

## Fix

1. Record `MessageEventType.MessageFailed` in `WolverineRuntime.MessageFailed`.
2. Call `tracker.MessageFailed(envelope, exception)` in `RecoverFromFailedProcessingAsync` alongside the existing `LogException`.

Note (1) is a behaviour change beyond tests: wire taps and any custom `IMessageTracker` will now see a `MessageFailed` event where they previously saw a `Sent`.

jeremydmiller:
Closes #4136. First of three stacked PRs; the other two are #4125's flag change and an unrelated codegen collision both found by this thread.

## What was wrong

**`WolverineRuntime.MessageFailed` recorded `MessageEventType.Sent`.** `Sent` is not terminal — `EnvelopeHistory` only completes a `Sent` record once a matching `Received` arrives, whereas `MessageFailed` completes every record sharing the envelope's `UniqueNodeId`. So a message that failed never reached a terminal state in a tracked session, and any session containing one could only end by **timing out**.

Every caller is on a terminal path — `MoveToErrorQueue`, the `MessageSucceededContinuation` catch, both batching continuations — so `Sent` was wrong at all of them. It went unnoticed because `MoveToErrorQueue` emits its own terminal `MovedToErrorQueue` record immediately afterwards, and that is the one path with coverage. The sibling `MessageSucceeded` already records the right type.

**`HandlerPipeline.RecoverFromFailedProcessingAsync` emitted no message event at all.** The pipeline's last line of defense acks the envelope away and calls `LogException`, which is free-form diagnostic text rather than a message event. An envelope that dies there — a handler chain that fails to compile at runtime, a sticky-handler misconfiguration where `HandlerGraph.HandlerFor` throws before any executor exists — was invisible to every `IMessageTracker`. **That is a production observability hole, not just a testing one:** failure metrics under-counted every such envelope.

## How it stayed hidden

`DoNotAssertOnExceptionsDetected()` also suppressed the tracked-session timeout assertion (#4125). With that assertion off, sessions that could only ever time out returned green. Ungating it surfaced exactly **3 failures across CoreTests' 2644 tests**, and all three traced back here.

## Behaviour change worth flagging

The first fix is visible beyond tests: wire taps and custom `IMessageTracker` implementations now see a `MessageFailed` event where they previously saw a `Sent`. That is the correct event for the situation, but it is a semantic change on a shipped line.

## Verification

- `dotnet build wolverine.slnx -c Release -f net9.0` — clean, 0 warnings.
- Full CoreTests: **2642 passed, 2 skipped, 0 failed.**
- New `terminal_event_on_message_failure` pins that a message dying in the pipeline's last line of defense records a terminal `MessageFailed` and the session completes instead of timing out.

### DIFF
--- a/src/Testing/CoreTests/Bugs/Bug_3413_unknown_destination_on_scheduled_recovery.cs
+++ b/src/Testing/CoreTests/Bugs/Bug_3413_unknown_destination_on_scheduled_recovery.cs
@@ -87,8 +87,15 @@ public class Bug_3413_unknown_destination_on_scheduled_recovery : IAsyncLifetime
             Destination = TheLocalDestination
         };
 
-        var session = await _host.ExecuteAndWaitAsync(
-            () => _host.GetRuntime().EnqueueDirectlyAsync([orphan, good]).AsTask());
+        // The orphan is dead-lettered with UnknownTransportException by design, so the exception
+        // assertion has to be opted out of. It used to pass without this only because the
+        // exception-carrying tracking record lost a race with session completion -- GH-4136 made
+        // MovedToErrorQueue the first terminal record and gave it the exception, so it is now
+        // deterministically present.
+        var session = await _host.TrackActivity()
+            .DoNotAssertOnExceptionsDetected()
+            .ExecuteAndWaitAsync(
+                _ => _host.GetRuntime().EnqueueDirectlyAsync([orphan, good]).AsTask());
 
         session.Received.SingleMessage<OrphanedMessage>()
             .Name.ShouldBe("deliver me");
--- a/src/Testing/CoreTests/Runtime/MoveToErrorQueueTester.cs
+++ b/src/Testing/CoreTests/Runtime/MoveToErrorQueueTester.cs
@@ -67,8 +67,15 @@ public class MoveToErrorQueueTester
     {
         await theContinuation.ExecuteAsync(theLifecycle, theRuntime, DateTimeOffset.Now, null);
 
-        theRuntime.MessageTracking.Received().MessageFailed(theEnvelope, theException);
         theRuntime.MessageTracking.Received().MovedToErrorQueue(theEnvelope, theException);
+
+        // GH-4136: exactly ONE terminal tracking event on this path. A terminal record sweeps every
+        // earlier record for the envelope complete, and the durable transports' dead-letter move emits
+        // a trailing Sent that nothing else completes -- so a second terminal record here cannot be
+        // ordered to satisfy both that sweep and the guarantee that the ending record is present.
+        // MessageFailed's dead-letter counter, effective time and failure wire tap moved into
+        // MovedToErrorQueue rather than being dropped.
+        theRuntime.MessageTracking.DidNotReceive().MessageFailed(theEnvelope, theException);
     }
 
     [Fact]
--- a/src/Testing/CoreTests/Tracking/terminal_event_on_message_failure.cs
+++ b/src/Testing/CoreTests/Tracking/terminal_event_on_message_failure.cs
@@ -1,0 +1,67 @@
+using Microsoft.Extensions.Hosting;
+using Shouldly;
+using Wolverine;
+using Wolverine.Attributes;
+using Wolverine.Tracking;
+using Xunit;
+
+namespace CoreTests.Tracking;
+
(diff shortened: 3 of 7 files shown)


jeremydmiller:
CI caught a real problem with my first pass at this — pushed a fix.

Reporting **both** `MessageFailed` and `MovedToErrorQueue` from `MoveToErrorQueue` was harmless while `MessageFailed` recorded the non-terminal `Sent`: `MovedToErrorQueue` came last and, being terminal, swept every earlier record for the envelope complete. Once `MessageFailed` became terminal there were two terminal records on one path, and **neither ordering works**:

| Order | What breaks |
| --- | --- |
| `MessageFailed` first | It completes the envelope, so the session can resume before `MovedToErrorQueue` lands — `TransportCompliance.shouldMoveToErrorQueueOnAttempt` fails "No ending activity detected" (CIRabbitMQ) |
| `MovedToErrorQueue` first | It stops sweeping the trailing `Sent` the durable transports' own dead-letter move emits, which nothing else completes — the session never completes at all. SqliteTests' dead-letter compliance went **625ms → 30s timeout** (CISqlite) |

So the path now reports `MovedToErrorQueue` and nothing else, last, carrying the exception. `MessageFailed`'s dead-letter counter, effective-time recording and failure wire tap are folded into `MovedToErrorQueue`, so no metric or subscriber loses anything. `MessageFailed` stays terminal for its other callers, none of which emit a second terminal event.

Worth recording: those Sqlite dead-letter tests were **previously green over a session that timed out for 30 seconds** — the timeout assertion was suppressed. Exactly the class of vacuous pass this branch exists to make visible.

Verified locally: CoreTests 2643/0, SqliteTests 169/0, CircuitBreakingTests 24/0, RabbitMQ dead-letter compliance 12/0, and `dotnet build wolverine.slnx -c Release -f net9.0` clean.

PR #4138, opened 2026-08-26
GH-4125: DoNotAssertOnExceptionsDetected no longer suppresses the tracked-session timeout assertion

### LINKED ISSUE #4125: DoNotAssertOnExceptionsDetected() also suppresses the tracked-session timeout assertion, so a resiliency test passes on a session that achieved nothing

jeremydmiller:
Closes #4125. **Stacked on #4137** — merge that one first; this PR retargets to `main` afterwards.

## The change

`AssertNotTimedOut()` shared the `AssertNoExceptions` flag with `AssertNoExceptionsWereThrown()`:

```csharp
if (AssertNoExceptions)               AssertNoExceptionsWereThrown();
if (AssertAnyFailureAcknowledgements) AssertNoFailureAcksWereSent();
if (AssertNoExceptions)               AssertNotTimedOut();   // <-- same flag
```

A resiliency test reaches for `DoNotAssertOnExceptionsDetected()` for its documented reason — handlers are *expected* to throw — and silently also opted out of the only assertion that would catch the session completing none of its work. The failure mode is a test that never fails rather than one that fails oddly, so it can sit in a suite indefinitely.

The session's own timeout is not one of the "exceptions detected during the message activity" the flag opts out of; it is the harness reporting that the activity never finished. So the assertion is **ungated** rather than moved behind a second flag — no new API, no migration story.

## On the competing use case

The issue raised a real second intent: `DoNotAssertOnExceptionsDetected()` was the only way to get an `ITrackedSession` back after a timeout, for `AllRecordsInOrder()` inspection when debugging a firehose.

Nothing is lost. `AssertNotTimedOut` builds its `TimeoutException` through `BuildActivityMessage`, which renders the full envelope-record grid, the registered conditions with their satisfied state, and every captured exception. **The message already is the inspection** — that is exactly what the three failures this change surfaced were diagnosed from.

## Why it is stacked

Failure tracking had to reach a terminal state first (#4136), or this assertion reports a five-second timeout in place of the real failure. Landing this alone turns 3 CoreTests red.

## Verification

- `dotnet build wolverine.slnx -c Release -f net9.0` — clean, 0 warnings.
- Full CoreTests on the stack: **2642 passed, 2 skipped, 0 failed.**
- New `timeout_assertion_is_not_suppressed_4125` covers both halves: a timed-out session still throws with the flag set, and handler exceptions are still suppressed by it.

### DIFF
--- a/docs/guide/messaging/listeners.md
+++ b/docs/guide/messaging/listeners.md
@@ -192,13 +192,20 @@ Two things follow, and both matter more than the headline percentage:
   already running, so nothing runs twice. The prefetch window is still redelivered — it simply had not been
   executed yet, so those are first executions.
 * **A hard kill costs about one duplicate per busy lane, not one per unacked message.** Only handlers that
-  were *mid-flight* when the connection died can run twice, and that population is the partition slot count,
-  not the prefetch depth. In the runs above that was three or four per killed node against an unacked window
-  of 180 — a factor of roughly 45, and it barely moved between runs.
+  were *mid-flight* when the connection died can run twice, and that population is the number of lanes the
+  endpoint runs, not the prefetch depth. On the group-partitioned endpoint measured here that is the partition
+  slot count: three or four duplicates per killed node against an unacked window of 180 — a factor of roughly
+  45, and it barely moved between runs. On an **unpartitioned** `NativeAck` endpoint there are no slots, and
+  the ceiling is `MaximumParallelMessages()` instead.
 
-None of this weakens the guarantee. Every duplicate still enters its group's sequential lane, so a duplicate
-never runs concurrently with the original, and the intra-group concurrency invariant held across every kill
-and every handoff in all of the runs above. Handlers must still be idempotent — at-least-once is the contract,
+On a **group-partitioned** endpoint, none of this weakens the guarantee: every duplicate re-enters its group's
+sequential lane, so a duplicate never runs concurrently with the original, and the intra-group concurrency
+invariant held across every kill and every handoff in all of the runs above. That is a property of group
+partitioning rather than of `NativeAck` itself — an **unpartitioned** `NativeAck` endpoint has no such lanes,
+and a redelivery can and will run alongside the original. Group partitioning is opt-in via
+[`PartitionProcessingByGroupId()`](/guide/messaging/partitioning); it is never on by default.
+
+Handlers must still be idempotent either way — at-least-once is the contract,
 and 0.1% of a flood is not a small number of messages — but size that work against in-flight lanes rather than
 against prefetch. The [in-memory idempotency guard](#in-memory-idempotency-guard) below removes the remainder
 within a *running* process; it cannot help across a node's death, because the guard dies with it.
--- a/docs/tutorials/idempotency.md
+++ b/docs/tutorials/idempotency.md
@@ -92,14 +92,16 @@ Idempotency checking within message handlers executing within `Buffered` or more
 you to "opt in." First though, the idempotency check in this case can be done in one of two modes:
 
 1. `Eager` -- just means that Wolverine will apply some middleware around the handler such that it will make an early database call to try to insert a skeleton placeholder in the transactional inbox storage
-2. `Optmistic` -- Wolverine will try to insert the skeleton message information as part of the message handling transaction to try to avoid extra database round trips
+2. `Optimistic` -- Wolverine would try to insert the skeleton message information as part of the message handling transaction to try to avoid extra database round trips
 
-To be honest, the EF Core integration will always use the `Eager` approach no matter what. Marten supports both modes, and the `Optimistic`
-approach may be valuable if all the activity of your message handler is in changes to that same database so everything can still be 
-rolled back by the idempotency check failing. 
+`Optimistic` was the more optimal of the two on paper, but as the warning above says, it is **not currently reachable**. Every
+provider -- EF Core, Marten, Polecat, and Fisher alike -- emits the `Eager` check for any handler configured with either style,
+so `IdempotencyStyle.Optimistic` behaves exactly like `IdempotencyStyle.Eager` at runtime. Treat the two as one option until
+that changes; nothing in your code has to change if it does.
 
-For another example, if your message handler involves a web service call to an external system or really any kind of action
-that potentially makes state changes outside of the current transaction, you have to use the `Eager` mode.
+`Eager` is also the mode you actually want in most cases: if your message handler involves a web service call to an external
+system or really any kind of action that potentially makes state changes outside of the current transaction, the check has to
+happen before the handler runs.
 
 With all of that being said, you can either opt into the idempotency checks one at a time with an overload of the `[Transactional]`
 attribute like this:
--- a/src/Persistence/EfCoreTests/eager_idempotency_emitted_once_4128.cs
+++ b/src/Persistence/EfCoreTests/eager_idempotency_emitted_once_4128.cs
@@ -1,0 +1,95 @@
+using System.Reflection;
+using JasperFx.CodeGeneration;
+using JasperFx.CodeGeneration.Model;
+using Microsoft.EntityFrameworkCore;
+using Shouldly;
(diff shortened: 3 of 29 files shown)


jeremydmiller:
Retargeted from `gh-tracking/terminal-event-on-message-failure` to `main` so this actually gets CI — `.github/workflows/dotnet.yml` only triggers on `pull_request: branches: [main]`, so while it was stacked on a feature branch it was getting **zero checks**.

Consequence: this PR's diff now shows #4137's commits as well, since they are its ancestors. The logical split is unchanged — two issues (#4136 and #4125), two commits, and #4137 is still the one to review and merge first. The combined stack is the meaningful thing to run CI against anyway: #4138's change alone turns 3 CoreTests red by design, because the timeout assertion cannot be trusted until failed messages reach a terminal state.

Merge #4137 first and this collapses to just the GH-4125 commit.

jeremydmiller:
Rebased onto the updated #4137, which fixes the CICircuitBreaking and CISqlite failures seen here. See the detail on #4137 — the dead-letter path now reports exactly one terminal tracking event. Local: CoreTests 2643/0, SqliteTests 169/0, CircuitBreakingTests 24/0.

jeremydmiller:
CI found four more sessions that were green over a timeout. Two different causes, and only one is a test problem.

**A real gap: `WaitForMessage` did not accept `MovedToErrorQueue`.** The condition behind `WaitForMessageToBeReceivedAt` accepted `MessageSucceeded` and `MessageFailed` as terminal outcomes, but not a dead-letter. So a message that fails into the DLQ **could never satisfy it**, and any session waiting on one ran to its full timeout every single time — `Kafka.DeadLetterQueueTests.dead_letter_message_has_exception_headers` was burning 60 seconds and passing anyway. Being dead-lettered *is* terminal: nothing further will ever happen to that envelope. Fixed in the condition, not in the test.

**Three genuinely expect the activity not to happen**, which is a legitimate thing for a test to want:
- ASB `ignores_message_not_matching_the_filter` — a subscription rule filters the message out, so it never arrives. Timing out is the assertion.
- Two circuit-breaker `pause_*_listener_on_matching_error_condition` tests — the message exists to latch the listener, so its envelope never reaches a terminal state; a separate listener-status waiter is the real synchronisation.

Those get a new **`DoNotAssertOnTimeout()`**, deliberately separate from `DoNotAssertOnExceptionsDetected()` and named for exactly what it does.

That is the issue's option 1, and I think the evidence now supports it over the alternatives: **both intents the issue described are real.** The defect was never that suppressing the timeout is wrong — it is that one flag served both intents, so the dangerous one happened *by accident*, to tests that only wanted exception suppression. Now it has to be asked for by name.

Verified locally: CoreTests 2648/0, SqliteTests 169/0, CircuitBreakingTests 24/0, Kafka DeadLetterQueueTests 6/0, and `dotnet build wolverine.slnx -c Release -f net9.0` clean.

Running tally for this branch: **8 tests** across CoreTests, SqliteTests, CircuitBreaking, ASB, and Kafka were passing over work that did not happen.

jeremydmiller:
CIMarten caught a genuine regression I introduced, and it invalidates the `WaitForMessage` change from the previous commit — reverted.

Letting `MovedToErrorQueue` satisfy `WaitForMessageToBeReceivedAt` was wrong: **a dead letter is not the end of the story.** It can be marked replayable and redelivered, and a test waiting on that message is usually waiting for exactly the second, successful pass. Completing on the dead-letter returned the session on the *failing* delivery and skipped the replay — `Bug_971_replay_dead_letter_queue_of_event_wrapper` failed with "No messages of type `IEvent<ErrorCausingEvent>` were received".

The Kafka test that motivated the change is now fixed where the problem actually was. `dead_letter_message_has_exception_headers` waited on `WaitForMessageToBeReceivedAt<DlqTestMessage>`, but that message fails into Kafka's **native** dead-letter topic and so produces neither `MessageSucceeded` nor `MessageFailed`. The condition could never be satisfied, the session ran its full 60 seconds every time, and it passed only because the timeout assertion was suppressed. It now waits on quiescence, exactly as its sibling `failed_message_moves_to_dead_letter_queue` already did — a per-test fix rather than a semantic change to a shared condition.

Verified locally: Kafka `DeadLetterQueueTests` 6/0, Marten `Bug_971` green, CoreTests 2648/0, and `dotnet build wolverine.slnx -c Release -f net9.0` clean.

Issue #4160, opened 2026-08-27
New WolverineFx.AmazonS3 package: IPersistenceFrameProvider for S3-backed entities and sagas, absorbing the claim check store


erdtsieck:
Follow-up to the discussion on #4082.

#4082 proposed `[Entity(Loader = typeof(...))]` — a per-parameter seam in core Wolverine for loading an entity out of anything Wolverine has no persistence provider for. The preference from that discussion was to **not** add a new core seam, and instead ship a full `IPersistenceFrameProvider` in a new package:

> I'm more or less voting to just have a full implementation of IPersistenceFrameProvider in a new AmazonS3 package. Then add a claim check implementation too.

The same user-facing capability lands either way — a plain `[Entity]` parameter resolving out of a bucket, keeping `Required` / `OnMissing` / `MissingMessage` — but the mapping logic lives in the package rather than in `Wolverine`. Nothing is added to `Wolverine` at all.

## Status

| PR | What | Base | State |
| --- | --- | --- | --- |
| #4161 | `AddStopConditionIfNull` accepts the null identity its signature declares, plus a `MissingMessage` containing a quote no longer emits code that will not compile | `main` | open |
| #4164 | One shared emulator probe; the S3 claim check suite silently skipped against a running LocalStack | `main` | open |
| #4165 | **`WolverineFx.AmazonS3`** — the `IPersistenceFrameProvider` | #4164 | open |
| — | S3 saga storage | #4165 | not started |
| — | Claim check key prefix + doc fixes | `main` | not started |

#4082 is closed as superseded. #4163 is closed — see below.

## Packaging: two packages, not one

The "claim checks, your thing, sagas" sketch reads as one S3 package, and #4163 tried that: fold the already-shipped `WolverineFx.ClaimCheck.AmazonS3` into `WolverineFx.AmazonS3`. **That was wrong and has been reverted.**

`WolverineFx.ClaimCheck.*` is a deliberate seven-package family — one solution folder, one `CIClaimCheck` target that string-formats `Wolverine.ClaimCheck.{name}.Tests` across all of them — and pulling S3 out meant special-casing S3 out of that helper. More decisively, the repo already answers the question the other way: **`WolverineFx.Marten` and `WolverineFx.ClaimCheck.Marten` ship side by side.** Azure Blob and GCS are object stores in the same position S3 is, so if S3 leaves the family the family stops meaning anything.

[...]

PR #4161, opened 2026-08-27
Let AddStopConditionIfNull accept the null identity its signature declares

erdtsieck:
First in the stack for #4160. Independent of the rest of it — this is two latent defects in the missing-data guard frames, both found while writing #4082.

## 1. The nullable identity was dereferenced anyway

```csharp
Frame[] AddStopConditionIfNull(Variable data, Variable? identity, IDataRequirement requirement);
```

Both implementations dereferenced it as `identity!` straight into the guard frames — `HandlerChain.cs`, `HttpChain.cs`. Passing the null the signature invites therefore crashed code generation instead of producing a guard, so the nullability on the parameter was decorative.

`ThrowRequiredDataMissingExceptionFrame` and `WriteProblemDetailsIfNull` now take a `Variable?`. With no identity they emit the message as it stands instead of substituting into it, and the stock message becomes `Required {Type} was not found` rather than `Unknown {Type} with identity {Id}` — there is no id, so the message must not ask for one. A supplied `MissingMessage` is used verbatim either way, `{Id}` placeholder included: nothing to substitute, and silently deleting the placeholder would hide the mistake.

`HttpHandler.WriteProblems` already took an `object? identity`, so the HTTP side needed no change beyond passing `null`.

## 2. The missing-data message was interpolated into generated source raw

```csharp
writer.Write($"throw new {...}(\"{Message}\".Replace(...));");
```

`Message` can come straight from `[Entity(MissingMessage = "...")]`, so a quote, a backslash or a newline in it emitted source that does not compile — from a string an application author is invited to write. Both frames now escape it through a new `string.ToStringLiteral()`.

That extension exists rather than reusing JasperFx's `Constant.For(string)` / `Constant.ForString(string)`, which only wrap the value in quotes. Anything reaching generated code from application configuration has to be escaped properly, and it handles quotes, backslashes, the C# escape characters, other control characters, and NEL / LINE SEPARATOR / PARAGRAPH SEPARATOR — which the C# lexer also treats as line breaks.

## Verification

- `CoreTests.Util.SourceCodeExtensionsTests` — 13 cases over `ToStringLiteral`, including that braces and non-ASCII characters are deliberately left alone.
- `CoreTests.Persistence.stop_condition_with_no_identity` and `Wolverine.Http.Tests.Persistence.stop_condition_with_no_identity` — 12 tests pinning the seam in both chain implementations: the stock message with and without an identity, a verbatim `MissingMessage`, a `{Id}` placeholder with no id to put in it, and that an identity is still carried through when there is one.
- `Wolverine.Http.Tests.Persistence.reacting_to_entity_attributes` gains two end-to-end cases on endpoints whose `MissingMessage` contains `"` and `\`, through `ProblemDetails` (JSON, so the quotes are visible in the assertion) and through `ThrowException`. Before the escaping fix these two endpoints do not compile at all.
- Regression: `CoreTests.Persistence.*` 249/249, `Wolverine.Http.Tests.Persistence.*` 36/36.
- `dotnet build wolverine.slnx -c Release -f net9.0` — 0 errors, 0 code warnings.

Nothing in the tree passes a null identity today, so part 1 removes a latent crash rather than fixing a visible one; the tests pin the contract at the seam instead of through a caller. The escaping half is reachable now, from any `MissingMessage` containing a quote.

### DIFF
--- a/docs/guide/handlers/persistence.md
+++ b/docs/guide/handlers/persistence.md
@@ -96,6 +96,19 @@ You can choose a different answer with the `OnMissing` property:
 | `EmptyContentWith204` <Badge type="tip" text="6.28" /> | Log it and stop | Empty **204** |
 | `ThrowException` | Throws `RequiredDataMissingException` | Throws `RequiredDataMissingException` |
 
+`MissingMessage` sets the text of that answer -- the `ProblemDetails` detail, or the
+`RequiredDataMissingException` message. Two placeholders are substituted into it:
+
+| Placeholder | Replaced with |
+|---|---|
+| `{0}` | the identity value, through `string.Format` |
+| `{Id}` | the identity value |
+
+The message is escaped on its way into generated code, so a quote, a backslash or a newline in it is
+safe. Where an entity was not addressed by a single identity value there is nothing to substitute, so
+the message is used exactly as written and the stock message names the entity type rather than an id
+it does not have.
+
 If you need or want any other kind of failure handling on the entity not being found, you'll need to
 use explicit code instead, maybe with a `LoadAsync()` "before" method to still keep your main
 handler or endpoint method a *pure function*.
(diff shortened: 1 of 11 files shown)


erdtsieck:
Pushed a docs addition I should have had in the first place: `MissingMessage` substitution was not documented anywhere, and it is exactly the code this PR changes. `docs/guide/handlers/persistence.md` now carries the `{0}` / `{Id}` placeholders, that the message is escaped on its way into generated code, and what happens when there is no identity to substitute.

Issue #4007, opened 2026-08-22
Claim check: the object-store backend suites never run in CI, and Azurite is missing from docker-compose


jeremydmiller:
Follow-up surfaced while closing epic #3511.

#4005 added `CIClaimCheck`, which runs the three database-backed claim-check suites (PostgreSQL, SQL Server, Marten). Before that, **no CI target ran any claim-check backend tests at all** — all five shipped backends had test projects that only ever executed on a developer machine, and were merely *compiled* by the full-solution build.

Four suites are still uncovered:

- `Wolverine.ClaimCheck.AmazonS3.Tests`
- `Wolverine.ClaimCheck.AzureBlobStorage.Tests`
- `Wolverine.ClaimCheck.GoogleCloudStorage.Tests`
- `Wolverine.ClaimCheck.Nats.Tests`

This is how a backend regression ships unnoticed: the packages are published from the Pack list, but nothing verifies their behavior on the way in.

## Azurite is the sharper half

Three of the four emulators are already in `docker-compose.yml` — `localstack` (:4566), `fake-gcs-server` (:4443), and `nats` (:4222). **Azurite is not**, even though `AzuriteFactAttribute` exists and guards every test in the Azure Blob suite.

The practical consequence: those tests skip **everywhere**, including on a dev machine running `docker compose up -d`. The Azure Blob claim-check backend's assertions have, as far as I can tell, never actually executed in CI. It is the one backend with no executed test coverage at all.

## Proposal

- Add an `azurite` service to `docker-compose.yml`.
- Extend `CIClaimCheck` (or add a sibling target) to build and run the four object-store suites, starting `localstack`, `fake-gcs-server`, `nats`, and `azurite`.
- Keep the existing per-suite skip guards so a developer without a given emulator running still gets a clean local run — the point is that **CI** has them all.

The suites are small and the emulators are lightweight, so this should stay well inside the 20-minute job cap; `CIClaimCheck` currently runs in about 2 minutes.

PR #4163, opened 2026-08-27
Add WolverineFx.AmazonS3 and move the S3 claim check store into it

erdtsieck:
Second in the stack for #4160, and the one to review first if you only review one — it is where the package identity decision lives. Reviewable on its own; PR #4161 is independent of it.

## Why a new package rather than a new folder in the old one

Amazon S3 is about to carry more than one Wolverine concern: an `IPersistenceFrameProvider` for `[Entity]`-loaded documents and saga storage are the next two PRs in this stack. `WolverineFx.ClaimCheck.AmazonS3` is named for one of them. So this is one package per *service* rather than one per *concern against the same service* — `WolverineFx.AmazonS3`, in `src/Persistence/Wolverine.AmazonS3/`, next to `Wolverine.CosmosDb`.

It lands with the claim check store already in it so it is never an empty shell, which also means this PR is worth nothing more than a rename if the rest of the stack is rejected — and still leaves the tree tidier than it found it.

## What happens to WolverineFx.ClaimCheck.AmazonS3

It has shipped continuously since 5.34.0, so it is **not** deleted. It stays as an assembly containing nothing but:

```csharp
[assembly: TypeForwardedTo(typeof(AmazonS3ClaimCheckStore))]
[assembly: TypeForwardedTo(typeof(AmazonS3ClaimCheckExtensions))]
```

An assembly that simply vanishes breaks anything already compiled against it at *load* time with a `TypeLoadException`, not at build time with something a developer can act on. The forwards make that a non-event.

The two types **keep their `Wolverine.ClaimCheck.AmazonS3` namespace** for the same reason — source reading `using Wolverine.ClaimCheck.AmazonS3;` has to keep compiling, and the docs sample says exactly that. The namespace not matching the assembly is the deliberate price; renaming it would make this a source-breaking change for no benefit.

Both packages ship at the same version out of `Directory.Build.props`, so the shim's dependency is always exact. It stays on the Nuke pack list. **Deprecating it on nuget.org is a manual step that is not in this PR** — say the word and I'll do it once this merges, or leave it undeprecated if you'd rather.

## An actual bug found on the way

The suite's LocalStack probe was one `ConnectAsync("localhost", 4566)` with a two second budget for the whole call. docker-compose publishes the gateway on `127.0.0.1:4566` only, `localhost` resolves to `::1` first on Windows, and the IPv6 attempt does not always fail inside that budget — so **all four claim check store tests skipped against a LocalStack that was up, and the suite reported green.** That is the same silent-skip failure mode as #4007, one layer down. The probe now tries every address the host resolves to.

The sibling `AzuriteFact`, `FakeGcsFact` and `NatsFact` probes are three more copies of the same code with the same bug. They are not touched here — a later PR in this stack folds all four onto one shared probe rather than fixing the same thing four times.

## Everything else

- `CIClaimCheck` only ever ran this suite for its claim check half, so the suite moves to a `CIAmazonS3` target of its own and out of `CIClaimCheck`'s LocalStack dependency. Added to the `tests.yml` matrix — `VerifyCITargetCoverage` requires it.
- Registered in `build/build.cs`'s `NugetProjects` — `ValidatePackList` requires it.
- The SQS oversized-message exception messages and the two docs pages that told users to `dotnet add package WolverineFx.ClaimCheck.AmazonS3` now name the new package. Those exception strings are user-facing, so sending people to a retired package id would be the wrong kind of accurate.

## Verification

- `Wolverine.AmazonS3.Tests` — 9/9, against LocalStack. Four of them are the moved claim check store tests, which **actually execute for the first time on this machine** thanks to the probe fix. Five are new, pinning the shim: the types resolve out of `Wolverine.AmazonS3`, the retired assembly declares no types of its own, an old-style `Type.GetType("…, Wolverine.ClaimCheck.AmazonS3")` still resolves and is listed in `GetForwardedTypes()`, and the namespace did not move with the assembly.
- `dotnet build wolverine.slnx -c Release -f net9.0` — 0 errors, 0 code warnings.
- `ValidatePackList` — "All 51 projects declaring a &lt;PackageId&gt; are in the Pack list".
- `VerifyCITargetCoverage` — "Every CI* target is reachable from a workflow", 41 CI targets.


erdtsieck:
Closing this — the packaging move in it was wrong.

`WolverineFx.ClaimCheck.*` is a deliberate seven-package family (one solution folder, one `CIClaimCheck` target that string-formats `Wolverine.ClaimCheck.{name}.Tests` over all of them), and pulling S3 out of it meant special-casing S3 out of that helper. More to the point, the repo already answers the question the other way: **`WolverineFx.Marten` and `WolverineFx.ClaimCheck.Marten` ship side by side.** Azure Blob and GCS are object stores in the same position S3 is, so if S3 leaves the family the family stops meaning anything.

So `WolverineFx.ClaimCheck.AmazonS3` stays exactly where it is, and the new `WolverineFx.AmazonS3` will contain only the persistence provider — the Marten shape.

Two things from this PR are real and survive on their own, in their own PRs off `main`:

- The LocalStack probe bug. `ConnectAsync("localhost", 4566)` with a two second budget for the whole call, against a gateway published on `127.0.0.1` only, where `localhost` resolves to `::1` first — all four S3 claim check store tests skipped against a LocalStack that was up, and the suite reported green. `AzuriteFact`, `FakeGcsFact` and `NatsFact` are three more copies with the same bug; all four are getting one shared probe.
- The provider itself, in a new `WolverineFx.AmazonS3`.

Tracked in #4160.

PR #4165, opened 2026-08-27
Add WolverineFx.AmazonS3: an IPersistenceFrameProvider for S3-backed documents

erdtsieck:
The main piece of #4160, and the replacement for #4082. Stacked on #4164 (shared emulator probe) because the new suite reuses that helper — review that one first, it is nine lines of real change.

This is the `IPersistenceFrameProvider`-in-a-package route rather than #4082's `[Entity(Loader = ...)]` seam in core. Nothing is added to `Wolverine` at all.

## The gap

Plenty of real entities live in a bucket rather than a database — an invoice's rendered content, a generated report, a document scan. Without a persistence provider for that, a handler drops out of Wolverine's declarative persistence entirely: inject a client, `await` it, and re-implement the "what if it is not there" half by hand in every handler that reads it.

Now a registered type resolves through a plain `[Entity]` parameter, keeping `Required`, `OnMissing` and `MissingMessage`, and the declarative `Storage.Store()` / `Delete()` return values write it.

## Why registration is explicit

```csharp
opts.UseAmazonS3Persistence(s3 =>
{
    s3.Store<InvoiceContent>(x =>
    {
        x.BucketName = "invoice-content";
        x.KeyFor = ctx => $"invoices/v7/{ctx.TenantId}/{Sanitize(ctx.Id)}.json.br";
    });
});
```

Both the bucket and the key function are **required** and there is no default key layout, for two reasons.

The first is that the identity-to-key mapping is the part only the application knows. The real key behind that example is a generation prefix, a tenant, a separator substitution and brotli; no attribute-level template survives contact with it, so Wolverine would either be useless against a bucket that already exists or grow a key-template language.

The second is the one #4082's write-up called *"a landmine, not an extension point"*: `TryFindPersistenceFrameProvider` takes the **first** provider whose `CanPersist` is true, after the `IsCatchAll` sort. A provider claiming anything an object store could theoretically hold would compete with Marten and EF Core for their own documents, with the winner depending on registration order. Explicit registration makes this provider **selective by construction** — `IsCatchAll => false`, `CanPersist` true only for types the application named. `s3_persistence_frame_provider_tests` reads `IsCatchAll` through the interface deliberately, so a lost override reads the `false` default and fails a test instead of quietly shadowing Marten.

The key function receives an `S3KeyContext` — entity type, identity, tenant — which is what makes a multi-part key expressible at all. `[Entity]`'s identity convention finds one id, which is all `session.LoadAsync<T>(id)` needs, but an object key is routinely built from a tenant **and** an id.

## Generated code

Constructor-injected client and configuration, session built inline, no service location:

```csharp
var s3DocumentSession = new S3DocumentSession(_amazonS3, _amazonS3Configuration);
var readInvoice = (ReadInvoice)context.Envelope.Message;
// Load the InvoiceContent from S3, or null if it is not there
var content = await S3StorageActionApplier.LoadAsync<InvoiceContent>(
    s3DocumentSession, ((ReadInvoice)context.Envelope.Message).Id, context.TenantId, cancellation)
    .ConfigureAwait(false);
var result_of_Assert1 = EntityIsNotNullGuard<InvoiceContent>.Assert(content, _loggerForMessage, "content", context.Envelope);
if (result_of_Assert1 == HandlerContinuation.Stop) return;
var outgoing1 = InvoiceHandler.Handle(readInvoice, content);
```

Making `S3DocumentSession` public was not cosmetic — as an internal type it forced service location, which `ServiceLocationPolicy.NotAllowed` refuses. The suite caught that.

The tenant is an argument rather than something the session digs out of the container, resolved in the same order the Marten, Polecat and Fisher session frames use: the chain's own `tenantId` variable, then the active `IMessageContext`. So an S3 document and a Marten document in one handler resolve the same tenant. `StorageConstants.DefaultTenantId` is normalised to null before a key function sees it — without that, non-tenanted writes landed under a literal `invoices/v7/*DEFAULT*/` prefix, which is how I found it.

## What it deliberately does not do

- **No `IMessageStore`.** S3 is a poor transactional inbox and outbox. Durability stays with Postgres, SQL Server or Marten; S3 documents sit alongside it.
- **No unit of work.** `CommitUnitOfWorkFrame` emits a comment saying so. Every write takes effect immediately, so a handler that writes two documents and then throws has written one of them. `Insert`, `Update` and `Store` are one `PutObject` — S3 has no insert-versus-update, so these are last-write-wins by design.
- **No transaction support.** `CanApply` claims no chains. Claiming one would also make `AutoApplyTransactions` ambiguous in the normal case of S3 documents alongside a transactional store.
- **No querying.** `TryBuildAllFrame` / `FirstOrDefault` / `Queryable` stay at their interface defaults, so `[All]` and `[Queryable]` fail at bootstrapping naming this provider. `ListObjectsV2` over a key prefix is a paged scan, and a scan that looks like a query is worse than an honest refusal.
- **No soft delete.** Only a serializer knows what deleted means for its payload.

## Packaging

`WolverineFx.AmazonS3` sits **alongside** `WolverineFx.ClaimCheck.AmazonS3`, which is untouched — the same split `WolverineFx.Marten` and `WolverineFx.ClaimCheck.Marten` already have. (I first tried folding the claim check store into this package, per the "claim checks, your thing, sagas" sketch; that broke the seven-package `ClaimCheck.*` family and its `CIClaimCheck` helper, so it is out. See the closing comment on #4163.)

Sagas are the next PR in the stack, using S3 conditional writes (`If-Match` / `If-None-Match`) for optimistic concurrency.

## Verification

`Wolverine.AmazonS3.Tests` — **30/30, 0 skipped**, against LocalStack:

- `loading_entities_from_s3` (6) — loads an object written *outside* Wolverine, a missing required object stops the handler, a present one lets it run, an optional one arrives as null, the key function sees the tenant (same id under two tenants, different bodies), and a tenanted key is invisible without the tenant. **No other persistence is registered in that host at all**, which is the point.
- `storage_actions_against_s3` (5) — store writes, store overwrites, store honours the tenant key, delete removes, and a `UnitOfWork<T>` applies every action.
- `s3_persistence_frame_provider_tests` (9) — the precedence and identity-type contract.
- `s3_document_mapping_validation` (6) and `s3_document_serializer_tests` (5) — bootstrap diagnostics and the brotli/gzip round trip.

Also: `Wolverine.ClaimCheck.AmazonS3.Tests` 4/4 after the shared `LocalStack.cs` move, `dotnet build wolverine.slnx -c Release -f net9.0` 0 errors / 0 warnings, `ValidatePackList` and `VerifyCITargetCoverage` both green.

Docs: `docs/guide/durability/amazon-s3.md`, badged 6.31, next to the CosmosDB page, opening with what this is *not* so nobody confuses it with claim checks.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -50,6 +50,7 @@ jobs:
           - CIAWSSqs
           - CIAWSSqsCompliance
           - CIAWSSns
+          - CIAmazonS3
           # Wolverine.AzureServiceBus.Tests is one project sharded three ways by CLASS, balanced on
           # MEASURED per-class durations. Namespaces cannot split it — 264 of its 306 tests are in the
           # root namespace. CIAzureServiceBus is the "everything else" shard, so a newly added class
--- a/build/CITargets.cs
+++ b/build/CITargets.cs
@@ -626,6 +626,23 @@ partial class Build
             }
         });
 
+    /// <summary>
+    /// GH-4160. WolverineFx.AmazonS3 stores documents and sagas as S3 objects. The suite keeps its own
+    /// LocalStack skip guard so a developer without the emulator still gets a clean local run.
+    /// </summary>
+    Target CIAmazonS3 => _ => _
+        .ProceedAfterFailure()
+        .Executes(() =>
+        {
+            var project = RootDirectory / "src" / "Persistence" / "Wolverine.AmazonS3.Tests" /
+                "Wolverine.AmazonS3.Tests.csproj";
+
+            BuildTestProjects(project);
+            StartDockerServices("localstack");
+
+            RunTestProject(project);
+        });
+
     Target CISqlite => _ => _
         .ProceedAfterFailure()
         .Executes(() =>
--- a/build/build.cs
+++ b/build/build.cs
@@ -318,6 +318,7 @@ partial class Build : NukeBuild
                 Solution.Persistence.Oracle.Wolverine_Oracle,
                 Solution.Persistence.Sqlite.Wolverine_Sqlite,
                 Solution.Persistence.CosmosDb.Wolverine_CosmosDb,
+                Solution.Persistence.AmazonS3.Wolverine_AmazonS3,
                 Solution.Persistence.ClaimCheck.Wolverine_ClaimCheck_AmazonS3,
                 Solution.Persistence.ClaimCheck.Wolverine_ClaimCheck_AzureBlobStorage,
                 Solution.Persistence.ClaimCheck.Wolverine_ClaimCheck_GoogleCloudStorage,
--- a/docs/.vitepress/config.mts
+++ b/docs/.vitepress/config.mts
@@ -352,6 +352,7 @@ const config: UserConfig<DefaultTheme.Config> = {
                         {text: 'Oracle Integration', link: '/guide/durability/oracle'},
                         {text: 'RavenDb Integration', link: '/guide/durability/ravendb'},
                         {text: 'CosmosDB Integration', link: '/guide/durability/cosmosdb'},
+                        {text: 'Amazon S3 Integration', link: '/guide/durability/amazon-s3'},
                         {text: 'Entity Framework Core Integration', collapsed: false, link: '/guide/durability/efcore', items: [
                                 {text: 'Transactional Middleware', link: '/guide/durability/efcore/transactional-middleware'},
                                 {text: 'Transactional Inbox and Outbox', link: '/guide/durability/efcore/outbox-and-inbox'},
--- a/docs/guide/durability/amazon-s3.md
+++ b/docs/guide/durability/amazon-s3.md
@@ -1,0 +1,157 @@
+# Amazon S3 Integration <Badge type="tip" text="6.31" />
+
+::: tip
+`WolverineFx.AmazonS3` stores **documents and sagas** as S3 objects. It is not the same thing as
+[claim checks](/guide/durability/claim-checks), which off-load a large *message payload* to shared
+storage and ship a token through the broker. Claim checks live in
+`WolverineFx.ClaimCheck.AmazonS3`, and the two are independent — install either or both.
+:::
+
+Plenty of real entities are not in any database. An invoice's rendered content, a generated report, a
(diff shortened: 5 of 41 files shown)


erdtsieck:
Pushed one more docs change: `docs/guide/handlers/persistence.md` named Marten, RavenDb and EF Core as the stores behind `[Entity]` and not S3, so a reader of the `[Entity]` docs would never learn this exists. It now points at the new page and says that the identity maps to an object key through the registered key function rather than through the store.

erdtsieck:
Question came up: can one app use `[Entity]` against both Marten and S3? Yes, and I had asserted it in the design without testing it here — now closed.

`mixed_persistence_precedence` (5 tests) pins it: a type registered with `Store<T>()` resolves to S3 and everything else falls through to the catch-all store, in **both** registration orders; an S3 registration with no document types claims nothing at all; and saga chains still resolve to the transactional store, because `CanApply` claims none.

The catch-all is a stub standing in for Marten rather than Marten itself — the question is purely consultation order, and the real thing would put a Postgres dependency on a suite whose CI target only starts LocalStack. Real-store permutations of the same rule already live in `PersistenceTests/persistence_provider_precedence_permutations.cs`.

Docs gained an "Alongside another store" section, including the part that is not automatic: **there is no atomicity across the two.** The Marten write commits in its transaction, the S3 write already happened, so a handler doing both and then throwing has written the object and not the document. The section says to move the S3 write to a projection or a follow-on message if that matters.

jeremydmiller:
Superseded by #4206, which is this exact commit rebased onto `main` after #4164 merged — squashing #4164 gave its commit a new sha, and this branch could not be rebased in place from our side. Merged as 2a3337c85bb5b20a7942587655947b6f6f2460c9 with your authorship intact.

Thank you for this one — the `IsCatchAll => false` reasoning and the refusal to fake querying or a unit of work are what make it safe to sit next to Marten.

The follow-ups from the review are landing immediately in a stacked PR: the shared `StorageActionCompliance` suite (17/17 against LocalStack), the S3 claim check store moving into `WolverineFx.AmazonS3` per Jeremy's call, a refusal for `Saga`-derived registrations, and the `NoSuchBucket`-read-as-404 narrowing.

PR #4183, opened 2026-08-28
GH-4180: logical message deduplication, opt-in, on its own table

### LINKED ISSUE #4180: Promote DeduplicationId to a first-class logical message id for handler-level idempotency (opt-in column on RDBMS message stores)
## Summary

Promote `DeduplicationId` from its current broker-specific role into a **first-class logical message id**, and use it for handler-level idempotency checks. Storage of the id as a real column would be **opt-in** on the RDBMS-backed message stores, verified by a new compliance test suite.

The transport half of this already exists and works on every transport. What is missing is storage, enforcement, and a retention policy.

## Why not `Envelope.Id`

`Envelope.Id` identifies a **delivery**, not an intent, and it cannot express "the same logical operation, published twice".

It is also unreachable. `Envelope.Id` is `{ get; set; }` (`Envelope.cs:368`), but every public send path is `(T message, DeliveryOptions?)` and `DeliveryOptions` has no `Id` — `DeliveryOptions.Override(envelope)` never touches it. So an application cannot supply one today.

And it should not: the default is `IdGenerator = NewId.NextSequentialGuid` (`Envelope.cs:363`), deliberately sequential for insert locality on the envelope tables' primary key. Any deterministic, content-derived id is random-ordered and would fragment that index.

`DeduplicationId` has none of these problems: it is a **string**, it is application-settable today, and it carries no ordering contract.

## What already exists

Verified against `main` at `2ee114f8e`:

| Piece | Where |
|---|---|
| `Envelope.DeduplicationId` — public settable | `Envelope.cs:401` |
| `DeliveryOptions.DeduplicationId` — public settable | `DeliveryOptions.cs:197` |
| Wire header constant `deduplication-id` | `EnvelopeConstants.cs:26` |
| Written to the wire | `EnvelopeSerializer.cs:383` |
| Read from the wire | `EnvelopeSerializer.cs:120` |
| Carried across envelope clones | `Envelope.Internals.cs:498` |
| Cleared on object-pool recycle | `Envelope.Internals.cs:573` (`Reset()`) |

So the value already round-trips on **every** transport, not only the two that consume it today (SQS FIFO via `AmazonSqsQueue.DetermineDeduplicationId`, and Pub/Sub via `PubsubEnvelopeMapper`).

The idempotency machinery also already exists: `IEnvelopeTransaction.TryMakeEagerIdempotencyCheckAsync` (`IEnvelopeTransaction.cs:13`) is implemented by RDBMS, EF Core, RavenDB, Polecat and Fisher, and `MessageContext` calls it on the way in (`MessageContext.cs:132`), throwing `DuplicateIncomingEnvelopeException`. It just keys on `Envelope.Id`.

## What is missing

`deduplication_id` is not a column anywhere — `DatabaseConstants` has no such constant. The value survives persistence inside the serialized envelope body, so it is stored but **not queryable and cannot carry a constraint**.

## Proposed design

**1. Opt-in column on the RDBMS message stores.** Off by default; existing deployments get no schema change. When enabled, `wolverine_incoming_envelopes` gains a nullable `deduplication_id` column with a **partial unique index** — unique only where not null — so traffic that sets no logical id is entirely unaffected.

**2. The column is write-only on the hot path.** `DatabaseConstants.IncomingFields` (`DatabaseConstants.cs:53`) is a single `static readonly` field list baked into every incoming SELECT, so making it conditional per store would be invasive. It does not need to be: nothing reads the dedup id back into an `Envelope` — the id is already in the serialized body. The column exists solely to carry the unique constraint and to answer an explicit existence check. That keeps the read path and `DatabasePersistence` untouched.

**3. Enforcement in the existing check.** `TryMakeEagerIdempotencyCheckAsync` prefers the logical id when one is present, and falls back to `Envelope.Id` when it is not — so the semantics are additive and the existing `DuplicateIncomingEnvelopeException` path is reused rather than duplicated.

## Retention is the first decision, not a detail

Incoming rows are deleted once handled, so a dedup marker has to **outlive the row it came from**. That means either a tombstone window on the incoming row, or a separate processed-ids table with a TTL.

Whichever it is, the guarantee is only as long as that window — and an unstated window turns "idempotent" into "idempotent for a while", which is worse than no guarantee because it will hold in testing and fail in production. The window should be explicit, configurable, and enforced by a reaper.

Worth settling before implementation, because it determines whether this is one column or a second table.

## Semantics to pin down

- **What does the second delivery do?** For a create-style command the right answer is a benign discard with an ack, not a dead-letter. Whatever it is, it should be observable — a duplicate that vanishes silently is indistinguishable from a message that was lost.
- **Scope the first cut to fire-and-forget.** Replaying the *original response* for a deduplicated `InvokeAsync<T>` is Stripe-style idempotency-key machinery — storing and returning the prior result — and is a much larger feature. Sending and publishing first; request/reply later or never.
- **Multi-tenancy:** is the logical id unique per tenant or globally? Per tenant seems right (it matches how the envelope tables are already partitioned by tenant in database-per-tenant setups), but it changes the index definition.

## Compliance tests

This wants its own suite rather than additions to `MessageStoreCompliance`, matching the existing per-concern pattern (`DeadLetterAdminCompliance`, `NodePersistenceCompliance`, `LeadershipElectionCompliance`, `StorageActionCompliance`, `ClearAllWolverineStorageCompliance`).

A `MessageDeduplicationCompliance` abstract suite would pin, per store:

- a second envelope with the same logical id is rejected, and the handler runs exactly once
- two envelopes with *different* logical ids both run
- an envelope with **no** logical id is unaffected — the `Envelope.Id` fallback still applies, and no constraint fires
- the id survives a round trip through the transport under test
- the guarantee holds across a process restart (this is the one that distinguishes a real durable check from an in-memory set)
- with the opt-in **disabled**, none of the above applies and nothing breaks

That last case matters: the opt-in means there are two supported configurations, and only a compliance suite that exercises both will keep the disabled path honest.

## Motivating use case

CritterWatch is adding operator-scheduled work — "rebuild this projection at 03:00 tonight", and later cron-style recurring jobs. Both are **create** commands delivered at-least-once, so an operator double-click, a console republish, or an agent that re-publishes upcoming occurrences ahead of time all produce duplicate work.

A logical id of the form `"{scheduleId}|{occurrenceUtc:O}"` makes every one of those a no-op, with no bespoke dedupe table on the consumer side — and being a string rather than a Guid, it is legible in the database when someone is working out why a job did not fire.

This is a general need for any command that creates something, not specific to CritterWatch; the scheduling case is just where it became unavoidable.

## Notes

Code-area references above are verified line reads at `2ee114f8e`, offered as pointers into the design — the retention and semantics sections are proposals, not findings.

jeremydmiller:
Closes #4180.

Promotes `DeduplicationId` from its broker-specific role into a first-class **logical message id**, and uses it for handler-level idempotency. Storage is opt-in (`Durability.EnableMessageDeduplication`), so an upgrade is a no-op for anyone who does not want it.

## The distinction this rests on

`Envelope.Id` identifies **one delivery**. That is the right identity for "the broker handed me this twice", and the wrong one for "the operator clicked *Rebuild* twice", "the console republished after a timeout", "the agent pre-published tonight's occurrence yesterday". Those are different deliveries of the *same intent* — different `Envelope.Id`s, and every one gets through today.

`Envelope.DeduplicationId` already round-tripped on every transport. What was missing was storage, enforcement, and a retention policy.

## Design decisions worth reviewing

**A separate `wolverine_deduplication` table, not a nullable column on the inbox.** The issue left this open; the column turned out not to work:

- Under `EnableInboxPartitioning` the inbox is `PARTITION BY LIST (status)`, and PostgreSQL refuses a unique index on a partitioned table unless the index carries the partition key. But marking an envelope handled `UPDATE`s `status`, which *moves the row between partitions* — so a `(deduplication_id, status)` index would let the same logical id exist once as `Incoming` and once as `Handled`. Structurally incapable of enforcing the guarantee, silently, and only for users who enabled partitioning.
- A marker has to outlive the inbox row it came from, and the inbox is reaped on `KeepAfterMessageHandling` — five minutes by default. `DeduplicationWindow` defaults to 24h and is stated rather than inherited.
- Per your own comment on the issue: the inbox is small, hot, and written by every node. Its own table means that schema never moves for this feature at all.

**Claiming is an `INSERT`, never a `SELECT`-then-`INSERT`.** The duplicates this exists to stop are concurrent. A check-then-act passes every single-threaded test written against it and fails under the operator double-click it was built for. There is a test that only a real claim survives (`concurrent_claims_of_the_same_id_produce_exactly_one_winner`).

**A failed non-transactional execution releases its claim.** Without this the first failure permanently claims the id, every retry is discarded as a duplicate of its own failed attempt, and the work silently never happens *while the logs report successful deduplication* — strictly worse than not having the feature.

## `IChain` and the per-chain-type divergence

`IChain` gains `Deduplication`, `RequiresDeduplication()`, `ResolveDeduplicationId(...)` and `BuildDeduplicationStopCondition(...)`. Resolving the id and claiming it are shared frames; only the refusal differs — the same seam as `AddStopConditionIfNull` / `CreateSimpleValidationFrame`.

| | Duplicate | Missing required id |
|---|---|---|
| **HandlerChain** | discard + ack. Deliberately **not** `DuplicateIncomingEnvelopeException` — that is the storage-level signal and `DurableReceiver.handleDuplicateIncomingEnvelope` routes it through a redelivery check that can dead-letter. Returning early also keeps exactly one terminal tracking event on the path. | throw → error policies → DLQ |
| **HttpChain** | `409` + `ProblemDetails`, configurable to a 2xx where a replay is benign | `400` + `ProblemDetails` |
| **GrpcChain** | `AlreadyExists` (AIP-193) | `InvalidArgument` |

Generated handler code:

```csharp
var deduplicationId = context.Envelope?.DeduplicationId;
var missingDeduplicationId = string.IsNullOrWhiteSpace(deduplicationId);
if (missingDeduplicationId) throw new MissingDeduplicationIdException(...);

var isDuplicateMessage = false;
if (!string.IsNullOrWhiteSpace(deduplicationId))
{
    isDuplicateMessage = !(await _messageDeduplicator.TryClaimAsync(deduplicationId, null, cancellation));
}
if (isDuplicateMessage) return;
try { /* handler */ }
catch { if (...) await _messageDeduplicator.ReleaseAsync(...); throw; }
```

Note the claim is guarded by a null check, so unkeyed traffic on a `Required = false` chain pays no database round trip.

## Two things to flag

**gRPC is only half-wired, on purpose.** The refusal semantics are implemented on all three gRPC chain types. The id-resolution half is not: gRPC generates each RPC method inside `AssembleTypes` with a method-local `ServerCallContext` expression rather than through the `Middleware` list that `ApplyDeduplication` weaves into. Rather than let `[Deduplicated]` silently do nothing there, `GrpcGraph` throws at bootstrap naming the limitation. Happy to do the plumbing in a follow-up if you want it in this release.

**A `[Deduplicated]` handler in an assembly shared with hosts that have not enabled deduplication is a hard bootstrap failure.** This is by design — the alternative is a host reporting itself as protected while every duplicate runs — but it bit me in the test suite and it will bite users in a modular monolith. Worth a docs callout beyond what is already there if you think the sharp edge is too sharp.

## Codegen phase gotcha

The frames are woven from `HandlerChain.applyCustomizations` / `HttpChain.AssembleTypes`, **not** from an `IHandlerPolicy`. `[Deduplicated]` is a `ModifyChainAttribute` and is applied well after `HandlerGraph.Compile()` runs the policies, and `IsTransactional` is not final until the stores' policies have run. Weaving from a policy produced a chain with the attribute set and no frames at all — which is how the first test run failed.

## Verification

- `dotnet build wolverine.slnx -c Release -f net9.0` — clean, 0 warnings
- CoreTests: 2666 passed
- Wolverine.Http.Tests: 1002 passed
- PostgresqlTests: 508 passed (includes 8 new deduplication tests, and 5 new HTTP ones)
- Baseline confirmed against a clean `origin/main` worktree

Storage: PostgreSQL, SQL Server, MySQL, SQLite. Other stores report unsupported and fail loudly rather than answering "yes, that's new" to every id. Scope is fire-and-forget; replaying the original response of a deduplicated `InvokeAsync<T>` is separate, larger machinery.

### DIFF
--- a/docs/.vitepress/config.mts
+++ b/docs/.vitepress/config.mts
@@ -289,6 +289,7 @@ const config: UserConfig<DefaultTheme.Config> = {
                         {text: 'Validation', link: '/guide/http/validation'},
                         {text: 'Fluent Validation', link: '/guide/http/fluentvalidation'},
                         {text: 'Problem Details', link: '/guide/http/problemdetails'},
+                        {text: 'Idempotency Keys', link: '/guide/http/deduplication'},
                         {text: 'Caching', link: '/guide/http/caching'},
                         {text: 'Output Caching', link: '/guide/http/output-caching'},
                         {text: 'Rate Limiting', link: '/guide/http/rate-limiting'},
--- a/docs/.vitepress/config.mts
+++ b/docs/.vitepress/config.mts
@@ -300,6 +300,7 @@ const config: UserConfig<DefaultTheme.Config> = {
                                 {text: 'How gRPC Handlers Work', link: '/guide/grpc/handlers'},
                                 {text: 'Code-First and Proto-First Contracts', link: '/guide/grpc/contracts'},
                                 {text: 'Error Handling', link: '/guide/grpc/errors'},
+                        {text: 'Idempotency Keys', link: '/guide/grpc/deduplication'},
                                 {text: 'Streaming', link: '/guide/grpc/streaming'},
                                 {text: 'Typed gRPC Clients', link: '/guide/grpc/client'},
                                 {text: 'Multi-Tenancy', link: '/guide/grpc/multi-tenancy'},
--- a/docs/guide/durability/idempotency.md
+++ b/docs/guide/durability/idempotency.md
@@ -27,3 +27,146 @@ using var host = await Host.CreateDefaultBuilder()
 ```
 <sup><a href='https://github.com/JasperFx/wolverine/blob/main/src/Persistence/PersistenceTests/Samples/DocumentationSamples.cs#L188-L198' title='Snippet source file'>snippet source</a> | <a href='#snippet-sample_configuring_keepaftermessagehandling' title='Start of snippet'>anchor</a></sup>
 <!-- endSnippet -->
+
+## Logical Message Deduplication <Badge type="tip" text="6.31" />
+
+::: warning
+This is opt-in, and turning it on provisions a new `wolverine_deduplication` table. Leaving it off
+means no schema change at all on upgrade.
+:::
+
+Everything above keys on `Envelope.Id`, which identifies **one delivery**. That is the right identity
+for "the broker handed me this same message twice", and it is the wrong identity for a different
+question:
+
+> The operator clicked *Rebuild* twice. The console republished the command after a timeout. A
+> scheduling agent pre-published tonight's 03:00 occurrence yesterday, and the scheduler published it
+> again on the night. Should the projection rebuild four times?
+
+Each of those is a *different* delivery of the *same intent*, so each carries a different
+`Envelope.Id` and every one of them gets through. What is needed is an identity for the intent, and
+that is what `DeduplicationId` is.
+
+`Envelope.DeduplicationId` is a `string`, it is set by your application, and it already round-trips on
+every transport Wolverine supports — it is written to and read from the `deduplication-id` wire header
+by Wolverine's own serialization, so it survives Rabbit MQ, Kafka, Azure Service Bus, SQS, and the
+rest without any transport-specific configuration. (Two transports also consume it natively: SQS/SNS
+FIFO queues and topics, and GCP Pub/Sub.)
+
+### Turning it on
+
+<!-- snippet: sample_enabling_message_deduplication -->
+<!-- endSnippet -->
+
+### Deduplicating a message handler
+
+<!-- snippet: sample_deduplicated_message_handler -->
+<!-- endSnippet -->
+
+and on the publishing side:
+
+<!-- snippet: sample_publishing_with_a_deduplication_id -->
+<!-- endSnippet -->
+
+A logical id is a `string` rather than a `Guid` on purpose: it is meant to be legible in the database
+when someone is working out why a job did not fire.
+
+The second message with that id never reaches your handler. It is discarded, acknowledged to the
+broker, and logged at `Information` — a duplicate that vanished without a trace would be
+indistinguishable from a message that was lost.
+
+### Where the id comes from
+
+By default a message handler reads `Envelope.DeduplicationId`. You can point at a member of the
+message itself instead, so publishers do not have to set `DeliveryOptions`:
+
+<!-- snippet: sample_deduplicated_from_the_message_body -->
+<!-- endSnippet -->
+
+`ValueSource.Header` reads an envelope header, and `ValueSource.Anything` uses the chain type's
+natural default.
+
+### Required or optional?
+
+`Required` defaults to `true`, and a message that arrives at a `[Deduplicated]` handler with no
+logical id is dead-lettered with a `MissingDeduplicationIdException`.
+
+That default is deliberately the strict one. The lenient reading is the dangerous one: a handler that
+asked for deduplication and then quietly processed every unkeyed message would report itself as
+protected while providing nothing, and the failure is invisible — the traffic succeeds, the
+duplicates run, and no log line distinguishes it from a working configuration.
+
+Set `Required = false` when a mixed stream is genuinely expected:
+
+<!-- snippet: sample_deduplication_with_optional_id -->
+<!-- endSnippet -->
+
+Unkeyed messages on such a handler pay no database round trip at all.
+
+### Applying it across many handlers
+
+<!-- snippet: sample_requiring_deduplication_ids_by_policy -->
+<!-- endSnippet -->
+
+The filter is required rather than optional. Applying logical deduplication to *every* handler would
+demand an id on traffic that has no business carrying one, and with `Required` defaulting to `true`
+that turns into a dead-lettered message per unkeyed send.
+
+### The deduplication window
+
+`DeduplicationWindow` defaults to 24 hours, and it is the whole guarantee: past that window the same
+logical id is accepted again.
+
+It is deliberately **not** `KeepAfterMessageHandling`. That setting exists to absorb a broker
+redelivery and defaults to five minutes, which would turn "idempotent" into "idempotent for a while" —
+worse than no guarantee, because it holds in testing and fails in production. Size the window against
+how long a duplicate could plausibly arrive: an operator double-click is seconds, a console republish
+is minutes, an agent pre-publishing tomorrow's occurrences is a day.
+
+A background reaper deletes expired claims on its own timer
+(`Durability.DeduplicationCleanupPollingTime`, default 5 minutes), in bounded batches, in its own
+transaction. It logs how many claims it removed each cycle — a count that keeps climbing is the signal
+that the window is too long, the cadence too slow, or the volume higher than the settings assume.
+
+### Failed handlers do not poison the id
+
+If your handler throws, the claim is released and a retry gets through. Where the handler is
+transactional the claim was written inside that transaction and the rollback removes it; where it is
+not, Wolverine issues a compensating release.
+
+This matters more than it sounds. Without it, the first failed attempt would permanently claim that
+logical id, every retry would be discarded as a duplicate of its own failed attempt, and the work
+would silently never happen while the logs reported successful deduplication.
+
+### Why a separate table
+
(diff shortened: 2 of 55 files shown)


jeremydmiller:
## gRPC is now wired end to end, and the sharp edge is gone

Two follow-up changes since the initial push.

### 1. gRPC deduplication is real, not just its refusal half

`ReadGrpcDeduplicationIdFrame` reads the logical id out of `ServerCallContext.RequestHeaders`, and all three chain flavours weave the full sequence in `AssembleTypes`.

Generated code (code-first, unary), sitting after tenant detection:

```csharp
// GH-4180: logical deduplication id from the 'idempotency-key' request metadata
var deduplicationId = context.ServerCallContext.RequestHeaders?.GetValue("idempotency-key");
var missingDeduplicationId = string.IsNullOrWhiteSpace(deduplicationId);
if (missingDeduplicationId)
    throw new RpcException(new Status(StatusCode.InvalidArgument, "..."));

var isDuplicateMessage = false;
if (!string.IsNullOrWhiteSpace(deduplicationId))
    isDuplicateMessage = !(await _deduplicator.TryClaimAsync(deduplicationId, null, context.ServerCallContext.CancellationToken));

if (isDuplicateMessage)
    throw new RpcException(new Status(StatusCode.AlreadyExists, "..."));

try { return await _bus.InvokeAsync<...>(request, context.CancellationToken); }
catch { await _deduplicator.ReleaseAsync(...); throw; }
```

Three things worth knowing:

**Frames are built per RPC method, not through `Middleware`.** gRPC's `CloneFrames` only deep-copies `ConstructorFrame` and `MethodCall` — everything else passes by reference — and frames carry per-method mutable state. A shared instance would corrupt every method but the last. Same rule `DetectGrpcTenantIdFrame` already documents.

**The requirement is resolved per RPC method too.** A gRPC chain is a whole *service*, so a chain-level requirement would force "all of this service's calls or none." A method attribute beats a service one; there's a test asserting the claim appears exactly twice across a three-RPC contract.

**I was wrong about why the old guard existed.** I said `[Deduplicated]` "happily sets the requirement on a gRPC chain and then nothing consumes it." It never got applied at all — the gRPC chains never run `applyAttributesAndConfigureMethods`, so `chain.Deduplication` was always null and the guard could not have fired. The attribute is now read directly off the type/method instead.

Supported shapes: proto-first all four; code-first unary and client-streaming; hand-written unary. The limits are the ones tenant detection already has — a `CallContext` parameter for metadata access, and an awaitable return shape. Asking for it elsewhere is a bootstrap failure naming the method.

### 2. Asking for deduplication without storage now warns instead of throwing

I flagged this as "possibly too sharp" in the description. It is: it broke three test assemblies before it broke anything else, **including CIPersistence on the first push** — where two documentation samples stopped every unrelated host in `PersistenceTests` from starting.

`[Deduplicated]` lives on a handler *type*, and handler types are discovered by every host that scans their assembly. A modular monolith where one module enables deduplication and a sibling does not is entirely legitimate, and the hard failure made the attribute unusable in any shared assembly — the sibling could not start through no fault of its own configuration.

The property that matters is untouched, because it was never enforced there: `MessageDeduplicator.TryClaimAsync` still throws on a store that cannot deliver rather than answering "yes, that's new" to every id. A misconfigured host cannot process one duplicate quietly. It just gets to start, and to be told why at startup rather than at first message.

The `[WolverineIgnore]` workaround the Postgres tests needed is removed, which is the proof: 508 pass with those handlers discovered by every host in the assembly.

### Verification

Full `wolverine.slnx -c Release -f net9.0` clean, 0 warnings. CoreTests 2666, Wolverine.Grpc.Tests 312, PersistenceTests 105 (the red job), PostgresqlTests 508, Wolverine.Http.Tests 1002.

New docs page: [Idempotency Keys](docs/guide/grpc/deduplication.md) under the gRPC guide, including the lower-case metadata trap.

Issue #4216, opened 2026-08-31
Follow-ups to #4209: cross-store coverage for scheduled promotion and identity matching, and the three sibling statements it left alone


jeremydmiller:
Follow-ups from #4209 (thanks @outofrange-consulting), which fixed the PostgreSQL partitioned-inbox promotion wedge described in #4202. That PR was deliberately scoped to the incident and named its own leftovers; this is those leftovers, plus the coverage question they raise.

Two independent workstreams. Happy to split if you'd rather.

---

# 1. Test coverage — the part that worries me most

`MessageStoreCompliance` is 44 tests and **every** store answers it: PostgreSQL, SQL Server, Oracle, MySQL, SQLite, RavenDb and CosmosDb. That is a good foundation, and it makes the gaps below cheap to close.

### 1a. Only three stores run the suite under `MessageIdentity.IdAndDestination`

| Store | default identity | `IdAndDestination` |
|---|---|---|
| PostgreSQL | ✅ | ✅ `PostgresqlMessageStore_with_IdAndDestination_Identity` |
| SQL Server | ✅ | ✅ `SqlServerMessageStore_with_IdAndDestination_Identity` |
| RavenDb | ✅ | ✅ `message_store_compliance_with_message_identity_using_id_and_destination` |
| **Oracle** | ✅ | ❌ |
| **MySQL** | ✅ | ❌ |
| **SQLite** | ✅ | ❌ |
| **CosmosDb** | ✅ | ❌ |

The #4209 defect was *identity-shape specific* — the whole fix is about matching `id` alone versus `(id, received_at)`. Four stores never run the shared suite under the shape that exposes that class of bug. Adding a subclass per store is a handful of lines each and reuses everything.

Oracle deserves saying out loud: it diverges from the other RDBMS providers more than any of them (`NUMBER` reads back as `Int64`, `MySqlObjectName`-style quoting differences, the URI-casing bug in #3820), and it is exactly the provider where a shared-base assumption is most likely to be wrong in a way nobody notices.

### 1b. `PollForScheduledMessagesAsync` has one shared contract, and it skips two stores

`MessageStoreCompliance` touches promotion in exactly one place — `scheduled_poll_stamps_envelope_with_originating_store` (the #2576 contract) — and it early-returns for RavenDb and CosmosDb:

```csharp
if (thePersistence is not IMessageDatabase database)
{
    // Non-database stores (e.g. RavenDb, CosmosDb) wire scheduled
[...]

PR #4219, opened 2026-09-01
GH-4216: scheduled promotion must match the whole identity on every provider

jeremydmiller:
Addresses workstream 1 of #4216 (coverage) and, because the coverage found live defects, item 2c and more besides. Workstream 2's other statements are untouched and stay open on the issue.

## What the coverage found

The issue predicted one provider and one symptom: SQL Server's `uspMarkIncomingOwnership` reassigning `owner_id` for a row at another destination. It is **four providers and two symptoms**.

#4209 fixed PostgreSQL's promotion to match `(id, received_at)` and to constrain the update to `Scheduled` rows. Every other RDBMS provider has its own `PollForScheduledMessagesAsync`, and all four still matched on the id alone with no status predicate:

| Provider | promotion predicate before |
|---|---|
| PostgreSQL | ids **and** uris — fixed by #4209 |
| SQLite | `where lower(id) IN (…)` |
| MySQL | `WHERE id IN (…)` |
| Oracle | `WHERE id IN (…)` |
| SQL Server | `uspMarkIncomingOwnership` → `WHERE id IN (SELECT ID FROM @IDLIST)` |

Under `MessageIdentity.IdAndDestination` that produces two defects, **neither needing partitioning**:

- a row sharing the id at another destination has its `owner_id` reassigned by a poll that never selected it — the ownership clobber 2c names; and
- a scheduled sibling at another destination that is **not due yet** is promoted to `Incoming`, so a message scheduled an hour out executes now.

The second is the one worth pausing on: it is a delivery-time correctness bug, not an ownership smell, and it was not previously suspected.

## The fix

Each provider builds an OR of `(id, received_at)` clauses, mirroring what Postgres does through `unnest(@ids, @uris)`.

The pairs have to be matched **as pairs**. An `id IN (…) AND received_at IN (…)` pair of lists matches the cross product, which is the same bug wearing a longer statement — that reasoning is in a comment on each provider so the next person does not "simplify" it back.

Each also gains the `status = 'Scheduled'` predicate. The load side of the poll already selects only `Scheduled` rows, so constraining the update to match costs nothing and keeps promotion off rows it never read.

**SQL Server is deliberately different.** Its `@IDLIST` is an `EnvelopeIdList` TVP of ids alone, shared with three other procedures, so widening it was the wrong trade. The procedure keeps its signature, gains the status predicate, and still runs under `IdOnly` — where an id genuinely is the whole identity. Only `IdAndDestination` takes the new inline pair-matching statement.

## Coverage (how this was found)

- **1a** — the whole `MessageStoreCompliance` suite now runs under `IdAndDestination` on SQLite, MySQL, Oracle and CosmosDb, not just PostgreSQL, SQL Server and RavenDb. Each gets its own schema/database/file so the two key shapes never share DDL.
- **1c** — #4209's two non-partitioning-specific cases move into the shared suite so every store answers them, expressed store-agnostically through `Admin.AllIncomingAsync()` and skipped under `IdOnly` and on non-database stores.

Both contracts failed on SQLite, SQL Server, MySQL and Oracle before this change and pass after, with **PostgreSQL passing throughout as the control**.

## Also

Grants `SELECT ANY SEQUENCE` to the Oracle test user. An identity column creates an implicit sequence (`ISEQ$$_nnnnn`) owned by the schema its table lives in, and `CREATE/DROP ANY SEQUENCE` do not cover *reading* one, so any suite owning a schema other than `WOLVERINE` failed with `ORA-41900` on insert. This is why the Oracle evidence lagged the other three.

## Verification

Full project runs, unfiltered:

- SQLite **217/218** (1 pre-existing skip)
- MySQL **174/174**
- Oracle **164/164**
- SQL Server **409/411** (2 skips)
- PostgreSQL **534/536** — the one failure is `Bug_1942 … inline_rabbitmq_listener_replay_does_not_loop`, which fails identically on a stashed baseline because RabbitMQ is not running locally. 536 vs the baseline's 528 is exactly the +8 the two new contracts add across four Postgres compliance subclasses.
- `dotnet build wolverine.slnx -c Release -f net9.0` clean, 0 warnings.

CosmosDb's new subclass is **not verified locally** — no emulator on this machine — so `CICosmosDb` is the first real run of it.

## Still open on #4216

Workstream 2's other statements (`ScheduleExecutionAsync`, `_markEnvelopeAsHandledById`, `RescheduleExistingEnvelopeForRetryAsync`), the `Incoming`/`Handled` partitioning pair, and the `MaximumBrokerRedeliveries` doc contradiction. Those are design calls rather than incident fixes and are better made deliberately.


jeremydmiller:
## CICosmosDb was red; the CosmosDb suite is pulled

First CI run: **40 pass, 1 fail** — `CICosmosDb`, which is exactly the lane I flagged as unverified locally.

It is not a test failure. The job was **killed at 20m15s** (`##[error]The operation was canceled`) with the `CosmosDbTests` process still alive and terminated as an orphan, after roughly **18 minutes of no output**. That lane runs in ~5–6 minutes on `main`, consistently green, and this branch was the only delta — adding a second full `MessageStoreCompliance` suite roughly doubles its work against the emulator.

I have pulled the CosmosDb subclass (912c4b582) rather than guess at CI, for a reason worth stating plainly: **it contributed nothing to the fix.** CosmosDb is not an `IMessageDatabase`, so both new promotion contracts early-return there. It was 1a coverage only, on the one lane I cannot reproduce locally, and 1a-for-CosmosDb needs a CI budget conversation that does not belong in a defect fix.

The four RDBMS providers whose promotion statements this PR actually fixes — SQLite, MySQL, Oracle, SQL Server — keep their `IdAndDestination` suites, and all four passed. Everything else in the first run was green, including `CIPersistence`, `CIMySql`, `CIOracle` and `CISqlServer`.

1a-for-CosmosDb stays open as a follow-up on #4216.

PR #4223, opened 2026-09-01
Improve RabbitMQ documentation for AddResourceSetupOnStartup and Auto…

saithis:
…Provision


saithis:
Changes are based on your comment in discord:
> AutoProvision() tells Wolverine to check for rabbit mq objects on demand as needed. AddResourceSetupOnStartup() adds an IHostedService that tries to force every single JasperFx resource to set itself up on app startup. Which includes all Marten/Wolverine database objects and Wolverine message brokers too.

And my wondering why it works without `AutoProvision`, even though the docs clearly state I would require it.

jeremydmiller:
@saithis Thank you, looks good.

PR #4241, opened 2026-09-02
Concurrent IHost.StopAsync no longer tears the agents down twice

lahma:
## The bug

`WolverineRuntime.StopAsync` guards itself with a "have I already stopped" flag, but sets that flag only **after** awaiting the agent cancellation:

```csharp
if (_hasStopped) return;
if (DynamicCodeBuilder.WithinCodegenCommand) return;

await _agentCancellation.CancelAsync();   // <- yields here

_hasStopped = true;
```

Both `IHostedService.StopAsync` and `IAsyncDisposable.DisposeAsync` route into this method, so two callers can each read the flag as unset, pass the guard, and then run the entire shutdown concurrently.

`teardownAgentsAsync` is not written for that. It clears every field it disposes, so the second pass sees its own null check pass and then dereferences a field the first pass has already cleared:

```csharp
if (NodeController?.DeferredWork != null)
{
    await NodeController.DeferredWork.DisposeAsync();   // <- Agents.cs:436
    NodeController.DeferredWork = null;                 // <- the other pass clears it here
}
```

```
System.NullReferenceException : Object reference not set to an instance of an object.
   at Wolverine.Runtime.WolverineRuntime.teardownAgentsAsync() in /_/src/Wolverine/Runtime/WolverineRuntime.Agents.cs:line 436
   at Wolverine.Runtime.WolverineRuntime.StopAsync(CancellationToken cancellationToken) in /_/src/Wolverine/Runtime/WolverineRuntime.HostService.cs:line 462
   at Microsoft.Extensions.Hosting.Internal.Host.ForeachService[T](...)
   at Microsoft.Extensions.Hosting.Internal.Host.StopAsync(CancellationToken cancellationToken)
   at Microsoft.AspNetCore.Mvc.Testing.WebApplicationFactory`1.DisposeAsync()
```

That is a real trace from a downstream CI run — roughly 1 in 80 runs of one integration fixture, thrown out of `WebApplicationFactory.DisposeAsync()` after the test's assertions had already passed. `_dispatcher` a few lines above has the same shape and can fail the same way.

The crash is only the visible half. The quiet half is that the node gets torn down **twice**: `NodeAgentController.StopAsync` is the tail of the teardown and is not itself guarded, so it deletes this node's row and reports `NodeStopped()` once per pass.

Only a **Balanced** node can reach any of this, which is why it shows up under integration tests with a real message store rather than in Solo hosts:

- Balanced is what builds the `DeferredWork` runner (`Agents.cs:281`).
- Balanced is what puts a linked token source on `_agentCancellation`, which is what makes `CancelAsync()` genuinely yield and open the window. A storeless Solo host has no registrations, so `CancelAsync()` returns `Task.CompletedTask` and the window never opens.

The code is unchanged on `main` as of 6.32.0.

## The fix

**1. The shutdown claim becomes a joinable `Task`, taken before the first await.** The caller that wins the `CompareExchange` drives the shutdown; every later caller gets a task that completes when *that* shutdown has finished:

```csharp
var inFlight = Volatile.Read(ref _stopped);
if (inFlight != null) return inFlight;

var completion = new TaskCompletionSource(TaskCreationOptions.RunContinuationsAsynchronously);
inFlight = Interlocked.CompareExchange(ref _stopped, completion.Task, null);
if (inFlight != null) return inFlight;

return runShutdownAsync(completion);
```

`IAsyncDisposable.DisposeAsync` then awaits it unconditionally rather than deciding whether to skip, so it can never dispose endpoints and transports out from under a shutdown that is still running. That hazard predates this PR — the old flag was also set near the top of `StopAsync`, long before the endpoint drain, store drain, ownership release and agent teardown — and this is where it gets closed rather than moved. (Raised by the Copilot review; see [the follow-up comment](https://github.com/JasperFx/wolverine/pull/4241#issuecomment-5509073717).)

The joinable task always completes successfully (`TrySetResult` in a `finally`). Joiners want to know the shutdown is over; propagating a stop failure that was already reported to whoever drove it would turn one failed stop into a failed `Dispose()` as well.

`_hasStopped` is gone in favour of `_stopped`, so there is one piece of shutdown state rather than a flag plus a task. The `WithinCodegenCommand` early-out sits above the claim, so that path still leaves the runtime unclaimed for a later real shutdown, exactly as before.

**2. Single-read discipline in the two teardown paths**, so neither depends on (1): `teardownAgentsAsync` and its `DisableAgentsAsync` twin read `NodeController` once into a local, and claim the dispatcher and the deferred runner with `Interlocked.Exchange` (`NodeAgentController.TakeDeferredWork()`) so only one pass owns each and the loser gets `null` rather than a field cleared out from under it.

Collapsing the three `if (NodeController != null)` blocks in each method onto one early return is a straight refactor: all three were guarded on the same condition.

## Regression test

`src/Persistence/PostgresqlTests/Bugs/Bug_concurrent_stop_tears_agents_down_twice.cs`. It starts a Balanced host on Postgres and calls `StopAsync` twice from the same thread — the second call can only run a shutdown of its own if the first yielded before claiming it. Because `NodeAgentController.StopAsync` is unguarded, the observer's `NodeStopped()` count is a direct count of teardown passes:

```csharp
var first = runtime.StopAsync(CancellationToken.None);
var second = runtime.StopAsync(CancellationToken.None);

await Task.WhenAll(first, second);

await observer.Received(1).NodeStopped();
```

The controller captures `runtime.Observer` in its constructor during `StartAsync`, so the substitute is swapped in on a built-but-not-started host.

It lives under `PostgresqlTests` rather than `CoreTests` because it needs a real message store to be a Balanced node at all.

Verified in both directions against the same test, forcing a recompile each way:

- original ordering: **FAIL** — `Expected to receive exactly 1 call matching NodeStopped() … Actually received 2 matching calls`
- with the fix: **PASS** — 1

The `NullReferenceException` itself needs a specific interleaving inside that window, so the test pins the double teardown, which is deterministic.

## Verification

- `dotnet build wolverine.slnx -c Release -f net9.0` — 0 warnings, 0 errors.
- `Wolverine.Http.Tests` (net9.0 via `DOTNET_ROLL_FORWARD`) — 1017 passed, 10 skipped. This is the suite that leans hardest on host stop and dispose, and it covers the `Bug_using_host_stop` probe change.
- `PostgresqlTests.LeaderElection` (net10.0) — 15/15. This is what exercises the restructured `DisableAgentsAsync`.
- `PersistenceTests` agents (net10.0) — 10/10.
- `PostgresqlTests` (net10.0, full project) — 539 passed, 1 skipped. The single failure is `Bug_1942_replay_dlq_to_buffered_or_inline.inline_rabbitmq_listener_replay_does_not_loop`, `BrokerUnreachableException` on 127.0.0.1:5672 — I only had the `postgresql` compose service up.
- `CoreTests` (net10.0, full project) — 2752 passed, 2 skipped, 4 failed: `result_types_end_to_end` ×3 and `Bug_2896_mixed_lifetime_enumerable_dependency`, all `Cannot resolve scoped service … from root provider`. Confirmed pre-existing — the same 4 fail identically in a clean worktree at `upstream/main`.

No .NET 9 runtime on this machine, so test runs are net10.0 except the roll-forward one; the net9.0 build gate from `CLAUDE.md` was run as specified.

### DIFF
--- a/src/Http/Wolverine.Http.Tests/Bugs/Bug_using_host_stop.cs
+++ b/src/Http/Wolverine.Http.Tests/Bugs/Bug_using_host_stop.cs
@@ -58,7 +58,7 @@ public class Bug_using_host_stop
     {
         var field = typeof(WolverineRuntime).GetField("_hasStopped",
             BindingFlags.NonPublic | BindingFlags.Instance);
-        return (bool?)field?.GetValue(runtime) == false;
+        return (int?)field?.GetValue(runtime) == 0;
     }
 
     private static async Task<IHost> CreateHostAsync(HostType hostType) =>
--- a/src/Http/Wolverine.Http.Tests/Bugs/Bug_using_host_stop.cs
+++ b/src/Http/Wolverine.Http.Tests/Bugs/Bug_using_host_stop.cs
@@ -56,9 +56,10 @@ public class Bug_using_host_stop
 
     static bool IsRunning(WolverineRuntime runtime)
     {
-        var field = typeof(WolverineRuntime).GetField("_hasStopped",
+        var field = typeof(WolverineRuntime).GetField("_stopped",
             BindingFlags.NonPublic | BindingFlags.Instance);
-        return (int?)field?.GetValue(runtime) == 0;
+        field.ShouldNotBeNull("WolverineRuntime._stopped was renamed; this probe needs updating");
+        return field.GetValue(runtime) is null;
     }
 
     private static async Task<IHost> CreateHostAsync(HostType hostType) =>
--- a/src/Persistence/PostgresqlTests/Bugs/Bug_concurrent_stop_tears_agents_down_twice.cs
+++ b/src/Persistence/PostgresqlTests/Bugs/Bug_concurrent_stop_tears_agents_down_twice.cs
@@ -1,19 +1,22 @@
 using IntegrationTests;
+using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.Hosting;
+using NSubstitute;
 using Shouldly;
 using Wolverine;
 using Wolverine.Postgresql;
-using Wolverine.Tracking;
+using Wolverine.Runtime;
+using Wolverine.Runtime.Agents;
 using Xunit;
 
 namespace PostgresqlTests.Bugs;
 
 /// <summary>
-/// WolverineRuntime.StopAsync guards itself with a "have I already stopped" latch, but it used to set
-/// that latch only AFTER awaiting the agent cancellation. Both IHostedService.StopAsync and
-/// IAsyncDisposable.DisposeAsync route into it, so two callers could each read the latch as unset, pass
-/// the guard, and then run the entire shutdown -- including teardownAgentsAsync -- concurrently. That
-/// method clears every field it disposes, so the second pass saw its own
+/// WolverineRuntime.StopAsync guards itself against running twice, but it used to read that guard,
+/// await the agent cancellation, and only then set it. Both IHostedService.StopAsync and
+/// IAsyncDisposable.DisposeAsync route into it, so two callers could each see "not stopped", pass the
+/// guard, and run the entire shutdown concurrently. teardownAgentsAsync is not written for that: it
+/// clears every field it disposes, so the second pass saw its own
 /// <c>NodeController?.DeferredWork != null</c> check pass and then dereferenced a field the first pass
 /// had already cleared, throwing NullReferenceException out of IHost.StopAsync:
 ///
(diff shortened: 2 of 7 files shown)


lahma:
Thanks — good catch on the direction, though the framing needs one correction: that overlap isn't introduced here. Before this branch `_hasStopped` was already set near the top of `StopAsync`, right after the agent cancellation and *long* before the endpoint drain, the store drain, the ownership release and the agent teardown. So `DisposeAsync` could already observe "stopped", skip, and start disposing endpoints and transports under a shutdown that had barely begun. Latching before the first await moved that boundary by exactly one statement.

It is a real hazard either way, and moving a boundary isn't a fix, so I've closed it properly in e1692671c rather than leaving it where it was.

**The claim is now a `Task` instead of a flag.** The caller that wins the `CompareExchange` drives the shutdown; every later caller gets a task that completes when *that* shutdown has finished:

```csharp
var inFlight = Volatile.Read(ref _stopped);
if (inFlight != null) return inFlight;

var completion = new TaskCompletionSource(TaskCreationOptions.RunContinuationsAsynchronously);
inFlight = Interlocked.CompareExchange(ref _stopped, completion.Task, null);
if (inFlight != null) return inFlight;

return runShutdownAsync(completion);
```

and `DisposeAsync` awaits it unconditionally instead of deciding whether to skip:

```csharp
// Unconditional, and this is the point of StopAsync being joinable rather than skippable: a
// shutdown another caller is still running has to FINISH before the endpoints and transports
// below are torn out from under it.
await StopAsync(CancellationToken.None);
```

Two deliberate details:

- The joinable task **always completes successfully** (`TrySetResult` in a `finally`). Joiners want to know the shutdown is over; propagating a stop failure that was already reported to whoever drove it would turn one failed stop into a failed `Dispose()` as well.
- `_hasStopped` is gone in favour of `_stopped`, so there's one piece of shutdown state rather than a flag plus a task. The reflection probe in `Bug_using_host_stop` reads the new field and now asserts it still exists, instead of silently reporting "running" if it's ever renamed again.

It also let me sharpen the regression test considerably. `NodeAgentController.StopAsync` is the tail of the agent teardown and isn't itself guarded — it deletes the node's row and reports `NodeStopped()` every time it runs — so the observer's `NodeStopped()` count is a direct count of teardown passes. Two concurrent `StopAsync` calls, verified both ways with a forced recompile:

- original ordering: `Expected to receive exactly 1 call matching NodeStopped() … Actually received 2 matching calls`
- with the fix: 1

Re-verified after the change: `dotnet build wolverine.slnx -c Release -f net9.0` clean; `Wolverine.Http.Tests` 1017/1027 (10 skipped), which is the suite that leans hardest on host stop/dispose; `PostgresqlTests.LeaderElection` 15/15; `PersistenceTests` agents 10/10; `PostgresqlTests` 539 passed with one RabbitMQ-unreachable failure from only having the `postgresql` compose service up. Full `CoreTests` has 4 pre-existing failures (`result_types_end_to_end` ×3 and `Bug_2896`, all `Cannot resolve scoped service … from root provider`) — I confirmed those fail identically in a clean worktree at `upstream/main`.

jeremydmiller:
@lahma Thank you, I wouldn't be surprised if this helps test automation stability downstream for our work too.

PR #4264, opened 2026-09-03
Models for leader election

mlh758:
I also updated the tutorial file on leader election since it seems like it had drifted.

P is a model checker like TLA+ but built around state machines instead of set math. I was curious how leader election would behave with a partial outage and AI makes it much cheaper to write models like this now so I thought it would be nice to just offer these up as artifacts instead of reading and then forgetting.

The nix flake makes it easier to use the models since P is behind on .NET versions. I could gitignore it instead if you don't want that coming along.


jeremydmiller:
@mlh758 Do you have a doc website for the "P" thing? Not anything I'm aware of. And man, there's a *ton* of testing in there as is for the leadership election.

mlh758:
https://github.com/p-org/P it's a similar idea to TLA+, it's more about testing a design than testing the implementation. If you want to make changes you can change the model and it will run the model against the new design and see if it validates any properties you specified needed to be held (like single live leader).

There's always a model <> implementation gap so they're never perfect. It can be a nice way to ask "do we need this part of the implementation?" and the quickly see if a simplification or new feature is sound. It also serves as runnable documentaion.

Here's a paper that goes over some real world benefit: https://lamport.azurewebsites.net/tla/formal-methods-amazon.pdf

It's the kind of thing I used to not bother with on smaller projects because maintaining the models is time consuming but it's a lot easier with AI available to do it automatically.

I also won't be offended if you're just not interested.

jeremydmiller:
@mlh758 Sorry man, "I've got 99 problems and this isn't one"

PR #4265, opened 2026-09-03
GH-4262: synchronize an endpoint's delayed configuration against Compile

### LINKED ISSUE #4262: Data race on Endpoint.DelayedConfiguration NREs in Endpoint.Compile (line 685) and can leave a projection shard paused indefinitely
## Summary

`Endpoint.Compile` snapshots `DelayedConfiguration` with `ToArray()` under `EndpointCollection._channelLock`, while `DelayedEndpointConfiguration.Apply()` removes itself from that same `List<T>` under a *different* lock (the routing convention's `_senderRegistrationLock`). One unsynchronized `List<T>`, two locks. When the two overlap, `ToArray()` hands back an array containing a **null element** and the very next statement dereferences it:

```csharp
// Endpoint.cs:685
foreach (var configuration in DelayedConfiguration.ToArray()) configuration.Apply();
```

Reproduced deterministically-enough below (three runs, hit on rounds 6 / 65 / 15).

## Where it was seen in the wild

A Marten async projection raised a side effect. The publish missed the route cache, so the conventional routing built a sending agent for that message type for the first time, and `Compile` threw:

```
System.NullReferenceException: Object reference not set to an instance of an object.
 at Wolverine.Configuration.Endpoint.Compile(IWolverineRuntime runtime) Endpoint.cs:line 685
 at Wolverine.Configuration.EndpointCollection.buildSendingAgent(Uri uri, Action`1 configureNewEndpoint) EndpointCollection.cs:562
 at Wolverine.Configuration.EndpointCollection.GetOrBuildSendingAgent(Uri address, Action`1 configureNewEndpoint) EndpointCollection.cs:288
 at Wolverine.Transports.MessageRoutingConvention`4...DiscoverSenders(Type messageType, IWolverineRuntime runtime)+MoveNext() MessageRoutingConvention.cs:227
 at Wolverine.Runtime.WolverineRuntime.findRoutes(Type messageType) WolverineRuntime.Routing.cs:293
 at Wolverine.Marten.Publishing.MartenToWolverineMessageBatch.PublishAsync[T](T message, MessageMetadata metadata) MartenToWolverineMessageBatch.cs:28
 at Marten.Events.Daemon.Internals.ProjectionBatch.PublishMessageAsync(Object message, MessageMetadata metadata) ProjectionBatch.cs:87
 at JasperFx.Events.Daemon.AggregationRunner`4.processPossibleSideEffects(...) AggregationRunner.cs:834
 at JasperFx.Events.Daemon.AggregationRunner`4.ApplyChangesAsync(...) AggregationRunner.cs:436
```

The consequence is the reason this is worth reporting rather than just fixing quietly: **the projection shard was left `Paused` indefinitely**, `failure_category=Other`, with this Wolverine-internal stack as its `pause_reason`. Four sibling shards on the same node and the same `IWolverineRuntime` kept running and stayed ~250 events ahead, so nothing about the host was unhealthy — one shard simply stopped, and an operator sees only a projection that quit.

Seen on 6.32.0/6.33.0, but this is **not a regression in either** — `Endpoint.cs`, `EndpointCollection.cs` and `MessageRoutingConvention.cs` are unchanged between `V6.32.0` and `V6.33.0`, and the race is as old as the two locks.

## The race

`DelayedConfiguration` is `internal readonly List<IDelayedEndpointConfiguration> DelayedConfiguration = new();` (`Endpoint.cs:212`), so the *list* cannot be null. The null is an element.

**Writer** — `DelayedEndpointConfiguration.Apply()` ends with:

```csharp
// IDelayedEndpointConfiguration.cs:61
_endpoint.DelayedConfiguration.Remove(this);
```

reached from `MessageRoutingConvention.tryRegisterSenderConfiguration`:

```csharp
// MessageRoutingConvention.cs:324
if (_configuredSenders.Add(messageType))
{
    _configureSending(configuration, new MessageRoutingContext(messageType, runtime));
    configuration.As<IDelayedEndpointConfiguration>().Apply();
}
```

which holds `_senderRegistrationLock`.

**Reader** — `Endpoint.Compile` at line 685, reached from `EndpointCollection.buildSendingAgent`, which holds `_channelLock`.

The two locks do not exclude each other. `List<T>.RemoveAt` decrements `_size`, shifts the tail down, then assigns `_items[_size] = default` — so a `ToArray()` that read the pre-decrement `_size` copies a slot that has already been nulled.

Both sides are reachable together on the paths in the stack above. `tryRegisterSenderConfiguration` runs *both* during startup (`PreregisterSenders`) and lazily on the first publish (`DiscoverSenders`), and the code says so at `MessageRoutingConvention.cs:291-294`:

> *"a publish racing the tail of startup would otherwise corrupt the non-thread-safe HashSet. See GH-2874."*

GH-2874 guarded `_configuredSenders`. The endpoint's `DelayedConfiguration` list is reachable from the same race and was left unguarded.

Separately, `Endpoint.Compile` is not thread-safe on its own account: `_hasCompiled` is a plain `bool` read at `Endpoint.cs:676` and assigned at `723`, so two threads can both execute the body.

## Reproduction

Console app, `net9.0`, `ProjectReference` to `Wolverine.csproj` and `Wolverine.RuntimeCompilation.csproj`. `Apply()` on one thread against `Compile()` on another, over one endpoint:

```csharp
using Microsoft.Extensions.DependencyInjection;
using Microsoft.Extensions.Hosting;
using Wolverine;
using Wolverine.Configuration;
using Wolverine.Runtime;
using Wolverine.Transports;
using Wolverine.Transports.Sending;

var host = await Host.CreateDefaultBuilder().UseWolverine(opts =>
{
    opts.ServiceName = "compilerace";
    opts.Durability.Mode = DurabilityMode.Solo;
}).StartAsync();

var runtime = host.Services.GetRequiredService<IWolverineRuntime>();
var nres = 0;
string trace = null;

for (var i = 0; i < 20_000 && nres == 0; i++)
{
    // A freshly created conventional sender endpoint, in the state the
    // findRoutes cache-miss path finds it in: configuration registered,
    // nothing compiled yet.
    var endpoint = new RaceEndpoint(new Uri($"race://q{i}"));
    var configs = Enumerable.Range(0, 12).Select(_ => new RaceConfiguration(endpoint)).ToArray();

    var gate = new ManualResetEventSlim();

    var registering = Task.Run(() =>            // the convention
    {
        gate.Wait();
        foreach (IDelayedEndpointConfiguration c in configs) c.Apply();
    });

    var compiling = Task.Run(() =>              // the first publish
    {
        gate.Wait();
        try { endpoint.Compile(runtime); }
        catch (NullReferenceException e)
        {
            if (Interlocked.Increment(ref nres) == 1) trace = e.ToString();
        }
        catch { }
    });

    gate.Set();
    await Task.WhenAll(registering, compiling);
}

Console.WriteLine(nres > 0 ? trace : "not reproduced");
await host.StopAsync();

public class RaceEndpoint : Endpoint
{
    public RaceEndpoint(Uri uri) : base(uri, EndpointRole.Application) { }
    public override ValueTask<IListener> BuildListenerAsync(IWolverineRuntime runtime, IReceiver receiver)
        => throw new NotSupportedException();
    protected override ISender CreateSender(IWolverineRuntime runtime) => throw new NotSupportedException();
}

public class RaceConfiguration : DelayedEndpointConfiguration<RaceEndpoint>
{
    public RaceConfiguration(RaceEndpoint endpoint) : base(endpoint)
        => add(e => e.TelemetryEnabled = true);
}
```

Three runs on `main` @ `51c62fa17` hit on rounds **6**, **65** and **15**, each with the top frame from the field capture:

```
System.NullReferenceException: Object reference not set to an instance of an object.
   at Wolverine.Configuration.Endpoint.Compile(IWolverineRuntime runtime) in .../Wolverine/Configuration/Endpoint.cs:line 685
```

The underlying shape confirms on its own too: an unsynchronized `List<T>.ToArray()` racing `Remove` yields a null element within ~11 rounds of a two-thread loop.

## Suggested fix

`Apply()` is already idempotent — it returns early on `_haveApplied` under its own `_locker`, and `Compile` calling it a second time is a no-op. So **the `Remove` at `IDelayedEndpointConfiguration.cs:61` is not load-bearing for correctness**; it is list hygiene, and it is already wrapped in a `try/catch` that swallows everything. Two options:

1. Drop the `Remove` entirely. Smallest change, removes the writer.
2. Keep it, but put both it and `Compile`'s snapshot behind a lock owned by the **endpoint** rather than by the configuration object (the current `_locker` is per-configuration, so it cannot serialize two configurations mutating one endpoint's list).

Worth closing `Compile`'s `_hasCompiled` check-then-act in the same pass, since concurrent `Compile` on one endpoint is reachable — `EndpointCollection.EndpointFor` (`:305`) and `ExclusiveListeners` (`:342`) both call it with no lock at all.

Happy to put up a PR if you'd like it in this shape.

jeremydmiller:
Fixes #4262.

## What was wrong

`Endpoint.Compile` snapshotted `DelayedConfiguration` with `ToArray()` under `EndpointCollection._channelLock`, while that same `List<T>` was mutated from two places holding a **different** lock — `RegisterDelayedConfiguration` (called from the `DelayedEndpointConfiguration` constructor, which a routing convention invokes under its own `_senderRegistrationLock`) and the `Remove` at the end of `Apply()`. Neither excluded the other.

A `List<T>` racing its own `ToArray()` hands back an array with a **null element**, in both directions: `RemoveAt` nulls the vacated slot after decrementing `_size`, and `Add` publishes the incremented `_size` *before* writing the element. `Compile` then dereferenced it.

## ⚠️ Correction to what I suggested in the issue

The issue proposed simply **deleting** the `Remove`, on the grounds that `Apply()` is already idempotent via `_haveApplied` so the removal is only list hygiene. That reasoning is right but **the fix is not sufficient**, and I only found out by testing it: `List<T>.Add` sets `_size` before writing the element, so `ToArray()` racing a *registration* produces a null element just as readily. Deleting the `Remove` would have closed one of two writers and left a rarer version of the same crash. Demonstrated in isolation — a two-thread `Add`/`ToArray()` loop yields a null element within ~48 rounds.

So this synchronizes instead.

## The fix, which is two parts

**1. The endpoint owns a lock over its own list.** `RegisterDelayedConfiguration`, a new `RemoveDelayedConfiguration` (which `Apply()` now calls instead of reaching into the list), and a `snapshotDelayedConfiguration()` helper all take it. The lock has to belong to the **endpoint** — the existing per-configuration `_locker` cannot serialize two *different* configurations mutating one endpoint's list.

The snapshot helper deliberately returns rather than applying: `Apply()` runs user-supplied callbacks and takes its own lock, and holding the endpoint's lock across either invites a deadlock.

**2. `add()` takes `_locker`.** This is a second instance of the same bug one level down, and it is why part 1 alone still crashed. `DelayedEndpointConfiguration`'s **base constructor publishes `this` into the endpoint's list before the derived constructor body has run its `add(...)` calls**, so a concurrent `Compile` can legitimately pick up a half-constructed configuration and `Apply()` it while those adds are still landing. `Apply()` already iterates `_configurations` under `_locker`; `add()` did not take it. The torn read surfaced as an NRE on `action(endpoint)` at `IDelayedEndpointConfiguration.cs:53`.

## Verification

Two stress tests in `CoreTests/Configuration/delayed_configuration_is_not_raced_by_compile.cs`, one per writer.

**Negative controls run, and they matter here** — a stress test that cannot fail is a vacuous guard:

| | apply/`Remove` side | register/`Add` side |
|---|---|---|
| stock `main`, 3 runs | **failed every run** (NRE, ~11 ms) | **failed every run** (NRE, ~110 ms) |
| with this PR, 5 runs | passed | passed |

The register test needed **20,000 rounds** to reproduce reliably where the apply test needs 750 — `Compile` snapshots once per endpoint, so the `Add` window is much narrower. At 750 rounds that test passed against stock code, i.e. it was worthless; the round counts are deliberately generous for that reason. Both tests together run in ~0.5 s.

Also confirmed against a real host: the standalone repro from the issue, and an `Add`-side variant, both clean at 20,000 rounds with this branch (the repro reproduced on rounds 6 / 65 / 15 before it).

Full `CoreTests` green.

## ⚠️ One thing this does NOT fix, deliberately

Because the base constructor publishes `this` before the derived constructor finishes, a concurrent `Compile` can still `Apply()` a configuration whose `add(...)` calls have not all landed. With this PR that is safe — no crash, and the iteration sees a consistent list — but such a configuration sets `_haveApplied = true`, so **actions added afterwards are silently never applied**. That is a latent config-loss window, it predates this change, and closing it means not publishing `this` from a base constructor — a larger design change I did not want to bundle with a crash fix. Happy to open a separate issue if you want it tracked.

I also left `Compile`'s `_hasCompiled` check-then-act alone (read at the top, assigned at the bottom, so two threads can both run the body). It is a real latent hazard but not what #4262 was, and it is reachable from `EndpointCollection.EndpointFor` and `ExclusiveListeners`, which call `Compile` with no lock at all. Say the word and I will fold it in or file it.

## Why it is worth fixing rather than shrugging at

The field capture came off the first-publish route-building path, and it **killed the Marten projection shard that was raising the side effect**: the shard was left `Paused` indefinitely, `failure_category=Other`, with a Wolverine-internal stack as its `pause_reason`, while four sibling shards on the same runtime kept running ~250 events ahead. An operator just sees a projection that stopped.

### DIFF
--- a/src/Testing/CoreTests/Configuration/delayed_configuration_is_not_raced_by_compile.cs
+++ b/src/Testing/CoreTests/Configuration/delayed_configuration_is_not_raced_by_compile.cs
@@ -1,3 +1,5 @@
+using System.Collections.Concurrent;
+using System.Runtime.CompilerServices;
 using NSubstitute;
 using Shouldly;
 using Wolverine;
@@ -123,6 +125,74 @@ public class delayed_configuration_is_not_raced_by_compile
                 await Task.WhenAll(registering, compiling);
             });
         }
+    }
+
+    /// <summary>
+    /// The other half of GH-4262: <c>Compile</c>'s own <c>_hasCompiled</c> check-then-act. A plain bool
+    /// read at the top and a plain bool write at the bottom, with nothing excluding the body in between,
+    /// so two threads could both read false and both run it. Concurrent Compile on ONE endpoint is
+    /// reachable -- <c>EndpointCollection.EndpointFor</c>, <c>ExclusiveListeners</c>,
+    /// <c>LeaderPinnedListeners</c> and both <c>StartListenerAsync</c> overloads call it with no lock at
+    /// all, while <c>buildSendingAgent</c> calls it under <c>_channelLock</c>.
+    ///
+    /// <para>Counting endpoint-policy applications is the cheapest way to observe the body running twice.
+    /// The damage it stands in for is quieter: the serializer pre-population at the end of Compile is
+    /// <c>_serializers = _serializers.AddOrUpdate(...)</c>, a read-modify-write on an ImHashMap field, so
+    /// a second thread through the body simply drops the first one's entries.</para>
+    /// </summary>
+    [Fact]
+    public async Task compile_runs_its_body_exactly_once_under_concurrent_callers()
+    {
+        var options = new WolverineOptions();
+        var policy = new CountingEndpointPolicy();
+        options.Transports.AddPolicy(policy);
+
+        var runtime = Substitute.For<IWolverineRuntime>();
+        runtime.Options.Returns(options);
+        runtime.DurabilitySettings.Returns(options.Durability);
+
+        var token = TestContext.Current.CancellationToken;
+        var endpoints = new List<RacedEndpoint>();
+
+        for (var round = 0; round < Rounds; round++)
+        {
+            var endpoint = new RacedEndpoint(new Uri($"stub://gh4262-compile-{round}"));
+            endpoints.Add(endpoint);
+
+            using var gate = new ManualResetEventSlim();
+
+            var callers = Enumerable.Range(0, 8).Select(_ => Task.Run(() =>
+            {
+                gate.Wait(token);
+                endpoint.Compile(runtime);
+            }, token)).ToArray();
+
+            gate.Set();
+            await Task.WhenAll(callers);
+        }
+
+        // Name a few of the endpoints that were compiled more than once rather than just asserting a
+        // count, so a failure says how badly the body re-ran rather than only that it did. Capped,
+        // because without the fix EVERY round fails -- 8 concurrent callers all ran the body on all 750
+        // endpoints when this was written -- and 750 lines of Shouldly output helps nobody.
+        var overCompiled = endpoints
+            .Where(x => policy.CountFor(x.Uri) != 1)
+            .Select(x => $"{x.Uri} applied {policy.CountFor(x.Uri)} times")
+            .ToArray();
+
+        overCompiled.Take(5).ShouldBeEmpty($"{overCompiled.Length} of {endpoints.Count} endpoints ran Compile's body more than once");
+    }
+
+    public class CountingEndpointPolicy : IEndpointPolicy
+    {
+        private readonly ConcurrentDictionary<Uri, StrongBox<int>> _counts = new();
+
+        public void Apply(Endpoint endpoint, IWolverineRuntime runtime)
+        {
+            Interlocked.Increment(ref _counts.GetOrAdd(endpoint.Uri, _ => new StrongBox<int>()).Value);
+        }
+
+        public int CountFor(Uri uri) => _counts.TryGetValue(uri, out var box) ? box.Value : 0;
     }
 
     public class RacedEndpoint : Endpoint
--- a/src/Testing/CoreTests/Configuration/delayed_configuration_is_not_raced_by_compile.cs
+++ b/src/Testing/CoreTests/Configuration/delayed_configuration_is_not_raced_by_compile.cs
@@ -1,0 +1,152 @@
+using NSubstitute;
+using Shouldly;
+using Wolverine;
+using Wolverine.Configuration;
+using Wolverine.Runtime;
+using Wolverine.Transports;
+using Wolverine.Transports.Sending;
+using Xunit;
+
+namespace CoreTests.Configuration;
+
+/// <summary>
+/// GH-4262: <see cref="Endpoint.Compile"/> snapshotted <c>DelayedConfiguration</c> with
+/// <c>ToArray()</c> while that same <see cref="List{T}"/> was mutated from two other places holding
+/// DIFFERENT locks — <c>RegisterDelayedConfiguration</c> (from the
+/// <see cref="DelayedEndpointConfiguration{TEndpoint}"/> constructor, which a routing convention calls
+/// under its own <c>_senderRegistrationLock</c>) and the removal at the end of that class's
+/// <c>Apply()</c>. <c>Compile</c> itself runs under <c>EndpointCollection._channelLock</c>, so neither
+/// mutation was excluded.
+///
+/// <para>A <see cref="List{T}"/> racing its own <c>ToArray()</c> yields an array containing a NULL
+/// element in BOTH directions — <c>RemoveAt</c> nulls the vacated slot after decrementing <c>_size</c>,
+/// and <c>Add</c> publishes the incremented <c>_size</c> before writing the element — and
+/// <c>Compile</c> dereferenced it:</para>
+///
+/// <code>
+/// System.NullReferenceException: Object reference not set to an instance of an object.
+///    at Wolverine.Configuration.Endpoint.Compile(IWolverineRuntime runtime) Endpoint.cs:line 685
+///    at Wolverine.Configuration.EndpointCollection.buildSendingAgent(...)
+///    at Wolverine.Transports.MessageRoutingConvention`4...DiscoverSenders(...)
+///    at Wolverine.Runtime.WolverineRuntime.findRoutes(Type messageType)
+/// </code>
+///
+/// <para>It was captured in the field on the first-publish route-building path, where it killed the
+/// Marten projection shard that was raising the side effect: the shard was left Paused indefinitely
+/// with this Wolverine-internal stack as its pause reason, while four sibling shards on the same
+/// runtime kept running. Reported from CritterWatch.</para>
+///
+/// <para>⚠️ These are stress tests. They cannot prove a race absent, and before the fix they failed
+/// within tens of rounds rather than reliably on round one — so treat a single green run as weak and
+/// the round counts as deliberately generous.</para>
+/// </summary>
+public class delayed_configuration_is_not_raced_by_compile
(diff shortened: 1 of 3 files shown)


jeremydmiller:
Added a second commit (`3e2f6d1c1`) for the half of #4262 this PR hadn't covered: `Compile`'s own `_hasCompiled` check-then-act.

The issue framed that as a tidy-up alongside the `DelayedConfiguration` NRE. The measurement says otherwise. With 8 concurrent callers on one endpoint, **all 750 rounds ran `Compile`'s body between 6 and 8 times** — the window isn't narrow, it's the normal outcome. Concurrent `Compile` on one endpoint is reachable from bare call sites too: `EndpointCollection.EndpointFor`, `ExclusiveListeners`, `LeaderPinnedListeners` and both `StartListenerAsync` overloads all call it with no lock at all; only `buildSendingAgent` holds `_channelLock`.

The cost is quieter than the NRE. Every endpoint policy ran once per caller, and the serializer pre-population at the end of `Compile` is `_serializers = _serializers.AddOrUpdate(...)` — a read-modify-write on an ImHashMap field, not an atomic one — so each thread through the body simply dropped the entries the others had added. A second `GenerationalIdempotencyGuard` could also be built over one a listener had already been handed.

The body moves behind a per-endpoint monitor with a double-check, and `_hasCompiled` becomes `volatile` so a caller taking the fast path and seeing `true` is guaranteed to see everything the body wrote first. Per-*endpoint* deliberately — compiling unrelated endpoints concurrently is normal on every startup and stays parallel.

On lock ordering: the new `_compileLock` is taken *after* `EndpointCollection._channelLock` on the `buildSendingAgent` path, and never while holding `_delayedConfigurationLock` — `snapshotDelayedConfiguration` returns before anything is applied — so it introduces no cycle among Wolverine's own locks. The one theoretical inversion it does open needs a user-supplied `IEndpointPolicy` that calls `runtime.Endpoints` from inside `Compile`; that was a considered trade rather than an oversight.

New test `compile_runs_its_body_exactly_once_under_concurrent_callers` counts endpoint-policy applications per endpoint. Verified both ways: it fails on 750/750 endpoints without the fix, passes with it. Full `CoreTests` is green (2767 passed).

PR #4269, opened 2026-09-04
GH-4267: share and throttle the tenant database list refresh behind FindAllAsync

### LINKED ISSUE #4267: FindAllAsync() re-enumerates the dynamic tenant database list on every call — retry amplification at 512 tenant databases
## Observed

`MessageStoreCollection.FindAllAsync()` refreshes the dynamic tenant database list on **every** call —
no TTL, no single-flight:

```csharp
public async ValueTask<IReadOnlyList<IMessageStore>> FindAllAsync()
{
    if (_onlyOneDatabase) return [Main];

    foreach (var tenantedMessageStore in _multiTenanted)
    {
        if (tenantedMessageStore.Source.Cardinality == DatabaseCardinality.DynamicMultiple)
        {
            await refreshTenantedDatabaseList(tenantedMessageStore);   // unconditional
        }
    }
    ...
}
```

With Marten's sharded tenancy that refresh is not cheap and not local:
`MartenMessageDatabaseSource.RefreshAsync` → `IMartenStorage.AllDatabases()` →
`ShardedTenancy.BuildDatabases()`, which **opens a connection to the master (pool) database** and reads
`mt_database_pool` + `mt_tenant_assignments` every time. `ShardedTenancy.Cardinality` is hardcoded
`DynamicMultiple`, so there is no configuration that opts out.

And `FindAllAsync()` sits on paths that run per listener and are wrapped in retries:

- `ListenerInboxRecovery.RecoverAsync` — per listener, per sweep (the comment there says the refresh is
  deliberate: *"dynamic tenant sources are refreshed on each sweep"*);
- `DelegatingMessageInbox.ReleaseIncomingAsync` ← `DurableReceiver.StopAndDrainAsync`, inside
  `executeWithRetriesAsync` — so a failure re-enters the refresh;
- `MessageDatabaseDiscovery`, `StartScheduledJobProcessing`, `MigrateAsync`, `AssertStorageExistsAsync`,
  dead-letter replay/discard.

That makes it self-amplifying: the refresh is what runs out of connections, the retry calls the refresh
again, and each concurrent caller opens its own.

## Measured

Sharded Marten + Wolverine, **512 tenant databases**, 2 556 tenants, 20 462 async shards over 3 nodes
(~6 800 projection agents and ~170 durability agents per node), Wolverine-managed distribution.
Wolverine 6.33.0 / Marten 9.31.2 / JasperFx 2.61.0. Client `Max Pool Size` is 100 per data source.

One 26-minute container lifetime, from the log the container left behind — 311 error lines:

| count | message |
|---|---|
| 233 | `Database operation at {Uri} failed after {Attempts} attempts; giving up` |
| 32 | `Invocation of {Message} failed!` |
| 29 | `Error trying to run durability agent commands` |
| 9 | `Unexpected failure at {Uri} (attempt {Attempt}/{Max})` |
| 3 | `Error running database operations {Operations} against message database {Database}` |
| 3 | `Error building the orphaned message sweep for database {Database}` |
| 2 | `Error trying to persist the assignment of agent {AgentUri} to Node {NodeId}` |

Underneath them, by exception:

| count | exception |
|---|---|
| 176 | `NpgsqlException: The operation has timed out` (raised inside `NpgsqlConnector.ConnectAsync`) |
| 125 | `NpgsqlException: Failed to connect to <master>:5432` |
| 5 | `NpgsqlException: Exception while reading from stream` |
| 3 | `NpgsqlException: The connection pool has been exhausted, either raise 'Max Pool Size'` |

The stack is the same one every time:

```
Npgsql.NpgsqlException: The operation has timed out
 ---> System.TimeoutException: The operation has timed out.
   at Npgsql.Util.NpgsqlTimeout.CheckAndGetTimeLeft()
   at Npgsql.Internal.NpgsqlConnector.ConnectAsync(NpgsqlTimeout timeout, CancellationToken)
   at Npgsql.PoolingDataSource.OpenNewConnector(...)
   at Npgsql.PoolingDataSource.<Get>g__RentAsync|33_0(...)
   at Npgsql.NpgsqlConnection.<Open>g__OpenAsync|42_0(...)
   at Marten.Storage.ShardedTenancy.BuildDatabases()
   at Marten.DocumentStore.Marten.IMartenStorage.AllDatabases()
   at Wolverine.Marten.MartenMessageDatabaseSource.RefreshAsync(Boolean withMigration)
   at Wolverine.Persistence.MessageStoreCollection.refreshTenantedDatabaseList(MultiTenantedMessageStore)
   at Wolverine.Persistence.MessageStoreCollection.FindAllAsync()
   at Wolverine.Persistence.Durability.DelegatingMessageInbox.ReleaseIncomingAsync(Int32, Uri)
   at Wolverine.Runtime.WorkerQueues.DurableReceiver.executeWithRetriesAsync(Func`1)
```

**The master database is idle while this happens** — 0.5% CPU and ~150 server-side connections against a
3 600 `max_connections` limit, over the whole window. So these are not server-side refusals; the
client-side data source for the master runs out of connections and the connect attempts time out. The
shard server meanwhile climbed from ~540 to ~1 440 average connections (2 510 peak) as the sweeps kept
walking all 512 databases.

## Impact

The process spends its time in connect-retry loops, and then cannot answer a liveness probe — ours is a
plain HTTP endpoint with zero health checks behind it — within 2 seconds, three probes in a row.
Kubernetes kills the pod, its ~6 800 projection agents land on the two survivors, and those are next:
**33 restarts per pod in 11 hours**, staggered, for as long as the load lasts. A projection version bump
that has to rebuild four sources has not been able to finish in that time.

## Ask

Could `refreshTenantedDatabaseList` get a **single-flight guard and a short TTL** per source, so that
concurrent hot-path callers collapse into one refresh instead of one connect each?

Freshness is not what these call sites need: `FindDatabaseAsync` already forces a refresh when a lookup
misses, and that is the path where a newly added tenant database actually has to be discovered
immediately. A bulk enumeration that is a second or two stale costs nothing — the next sweep picks the
database up. Making the window configurable would let a deployment with genuinely volatile tenancy opt
back into per-call refreshes.

Two smaller things I noticed next to it, happy to split them off if you would rather:

- `DelegatingMessageInbox.ReleaseIncomingAsync(ownerId, receivedAt)` walks **every** store serially, so
  one listener drain is 512 round trips even though only the databases that hold rows for that listener
  matter.
- `MartenMessageDatabaseSource.RefreshAsync()` (the parameterless one) passes `withMigration: true`.
  Harmless for known databases, but `RefreshLiteAsync` already exists and is what a hot path wants.

Related: #3376 — same deployment, same "cost scales with databases × callers" shape, different seam.
That one was about which nodes connect at all; this one is about how often a single node re-enumerates.

erdtsieck:
Closes #4267.

## The problem

`MessageStoreCollection.FindAllAsync()` re-enumerated a `DynamicMultiple` tenancy source on every
call. With Marten's sharded tenancy that is a round trip to the tenant registry
(`ShardedTenancy.BuildDatabases()` opens a connection to the master database and reads
`mt_database_pool` + `mt_tenant_assignments`), and `FindAllAsync()` sits on paths that are **retried
on failure**: listener inbox recovery, listener drain via
`DelegatingMessageInbox.ReleaseIncomingAsync`, the durability sweeps, database discovery.

So the retry for a connection failure opened another connection to look the databases up again, and
concurrent callers each opened their own. On our fleet — 512 tenant databases, ~2 550 tenants, 20 462
async shards over 3 nodes — one 26-minute process lifetime logged 233 × *"Database operation at {Uri}
failed after {Attempts} attempts; giving up"* over 176 Npgsql timeouts raised inside
`NpgsqlConnector.ConnectAsync` and 3 × *"the connection pool has been exhausted"*, while the master
database itself sat at 0.5% CPU with ~150 of its 3 600 connections in use. It is the client-side data
source that runs dry, and the retry then asks it for another connection.

## The change

Concurrent callers join one refresh, and a refresh that has just succeeded is not repeated until
`DurabilitySettings.TenantDatabaseListStaleTime` has elapsed — 5 seconds by default, `TimeSpan.Zero`
for the old per-call behaviour.

Three things deliberately left as they were:

- **Only a successful refresh opens the window.** A failure leaves the list unknown, so the next
  caller retries it — one at a time, because the single-flight guard still holds.
- **Categorizing the active stores is not throttled.** It is in-memory, and a store the source built
  for a single-tenant lookup has to reach the collection either way.
- **`FindDatabaseAsync` still forces a refresh on a miss.** That is the path where a newly
  provisioned tenant database has to be found immediately; the window only bounds how often the
  callers that enumerate *every* database go asking.

Single-database and `StaticMultiple` deployments are untouched — `FindAllAsync` returns early for the
first and never entered this branch for the second.

## Tests

`CoreTests.Persistence.throttling_the_tenant_database_list_refresh`, four cases: concurrent callers
share one refresh; a second enumeration inside the window does not ask again; the list is refreshed
once the window has elapsed; a failed refresh is retried by the next caller. First two are red
without the fix, last two guard against over-caching. Committed red first, then the fix.

- `CoreTests` on net10.0: 2 768 passed, 0 failed, 2 skipped.
- `PostgresqlTests.MultiTenancy` per class on net10.0: `exclusive_listener_recovery_across_tenant_databases`,
  `multi_node_tenant_database_connections`, `queue_counts_across_tenant_databases` and
  `clear_all_wolverine_storage_across_tenant_databases` all green.
- `static_multi_tenancy` fails 1 of 6 for me (`have_all_the_correct_durability_agents`, a Shouldly
  compare on the agent list) — it fails identically on unmodified `main` in the same environment, and
  that class uses `StaticMultiple`, which this change does not touch.

## Open questions

- **Is 5 seconds the right default?** It is the same order as the other polling knobs in
  `DurabilitySettings`. Happy to make it shorter, or to leave the default at zero so nobody's
  behaviour changes without opting in.
- **`RefreshAsync()` vs `RefreshLiteAsync()`.** The throttle calls `RefreshAsync()`, so behaviour is
  unchanged, but the parameterless one passes `withMigration: true` and `RefreshLiteAsync` looks like
  what a hot path actually wants. Left alone here because it is a separate decision.
- `DelegatingMessageInbox.ReleaseIncomingAsync(ownerId, receivedAt)` still walks every store
  serially, so one listener drain is 512 round trips. Not touched — say the word and I will look at
  it in a follow-up.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,27 @@
 
 ### WolverineFx (core)
 
+- **Concurrent callers share one tenant database list refresh.** (closes
+  [#4267](https://github.com/JasperFx/wolverine/issues/4267)) `MessageStoreCollection.FindAllAsync()`
+  re-enumerated a `DynamicMultiple` tenancy source on every call, and it sits on paths that are *retried on
+  failure* -- listener inbox recovery, listener drain, the durability sweeps. With Marten's sharded tenancy
+  that enumeration is a round trip to the tenant registry, so the retry for a connection failure opened
+  another connection to look the databases up again, and concurrent callers each opened their own. On a
+  512-database fleet one 26-minute process logged 233 give-ups over 176 Npgsql connect timeouts and three
+  pool exhaustions, while the master database sat at 0.5% CPU with 150 of its 3600 connections in use -- it
+  is the client-side data source that runs dry, and the retry then asks it for another connection.
+
+  Concurrent callers now join a single in-flight refresh. That sharing is not configurable and is the half
+  that closes the storm: the cost was the fan-out, not the frequency.
+
+  `DurabilitySettings.TenantDatabaseListStaleTime` additionally bounds how often that one refresh happens,
+  and **defaults to zero, so no existing behaviour changes**. Raise it on a large fleet. A non-zero value
+  only ever affects *bulk* enumeration -- a lookup that misses (`FindDatabaseAsync`) always forces past the
+  window, because it is refreshing precisely on account of not having found the database, and answering it
+  from a list the window is vouching for would defeat the call. Forcing skips the freshness check, never the
+  single-flight guard.
+
+
 - **A node no longer sweeps up its own in-flight stop as a wedged shard.** (closes
   [#4240](https://github.com/JasperFx/wolverine/issues/4240)) An event-subscription agent could end up
   running on two nodes at once while `wolverine_nodes` credited only one of them, so nothing in the
--- a/src/Testing/CoreTests/Persistence/throttling_the_tenant_database_list_refresh.cs
+++ b/src/Testing/CoreTests/Persistence/throttling_the_tenant_database_list_refresh.cs
@@ -79,6 +79,55 @@ public class throttling_the_tenant_database_list_refresh
     }
 
     [Fact]
+    public async Task a_lookup_that_misses_forces_past_the_stale_window()
+    {
+        // GH-4267 follow-up. FindDatabaseAsync refreshes precisely BECAUSE it could not find the database,
+        // so a list the window is vouching for is the one answer it must not accept: the caller is asking
+        // about a database that, by construction, is not in that list. Left throttled, a tenant database
+        // provisioned moments ago is invisible to a single-database lookup until the window elapses -- and
+        // BuildAgentAsync turns that null into "No database with Uri ... supports a durability agent".
+        theRuntime.Options.Durability.TenantDatabaseListStaleTime = 30.Seconds();
+        theSource.RefreshAsync().Returns(Task.CompletedTask);
+
+        // Opens the window.
+        await theStores.FindAllAsync();
+        await theSource.Received(1).RefreshAsync();
+
+        // Well inside it, and a miss.
+        (await theStores.FindDatabaseAsync(new Uri("wolverinedb://fake/provisioned-a-moment-ago")))
+            .ShouldBeNull();
+
+        await theSource.Received(2).RefreshAsync();
+    }
+
+    [Fact]
+    public async Task a_forced_lookup_still_joins_a_refresh_already_under_way()
+    {
+        // Forcing skips the freshness check, NOT the single-flight guard. That guard is the half that
+        // closes the connection storm -- every concurrent caller and every retry opening its own
+        // connection to the registry -- so a miss must never be a licence to open another one.
+        theRuntime.Options.Durability.TenantDatabaseListStaleTime = 30.Seconds();
+
+        var refreshIsRunning = new TaskCompletionSource(TaskCreationOptions.RunContinuationsAsynchronously);
+        theSource.RefreshAsync().Returns(refreshIsRunning.Task);
+
+        var enumerating = theStores.FindAllAsync();
+        var missing = Enumerable
+            .Range(0, 12)
+            .Select(i => theStores.FindDatabaseAsync(new Uri($"wolverinedb://fake/missing-{i}")).AsTask())
+            .ToArray();
+
+        await theSource.Received(1).RefreshAsync();
+
+        refreshIsRunning.SetResult();
+
+        await enumerating;
+        await Task.WhenAll(missing);
+
+        await theSource.Received(1).RefreshAsync();
+    }
+
+    [Fact]
     public async Task a_failed_refresh_is_retried_by_the_next_caller()
     {
         theRuntime.Options.Durability.TenantDatabaseListStaleTime = 30.Seconds();
--- a/src/Testing/CoreTests/Persistence/throttling_the_tenant_database_list_refresh.cs
+++ b/src/Testing/CoreTests/Persistence/throttling_the_tenant_database_list_refresh.cs
@@ -1,0 +1,96 @@
+using CoreTests.Runtime;
+using JasperFx.Core;
+using JasperFx.Descriptors;
+using NSubstitute;
+using Shouldly;
+using Wolverine.Persistence;
+using Wolverine.Persistence.Durability;
+using Xunit;
+
+namespace CoreTests.Persistence;
+
+// GH-4267. Every FindAllAsync() used to re-enumerate the tenant database list, and the paths that call it
+// are retried on failure, so on a large tenant fleet the retry for a connection failure opened another
+// connection to the tenant registry to look the databases up again.
+public class throttling_the_tenant_database_list_refresh
+{
+    private readonly MockWolverineRuntime theRuntime = new();
+    private readonly ITenantedMessageSource theSource = Substitute.For<ITenantedMessageSource>();
+    private readonly MessageStoreCollection theStores;
+
+    public throttling_the_tenant_database_list_refresh()
+    {
+        theSource.Cardinality.Returns(DatabaseCardinality.DynamicMultiple);
+        theSource.AllActive().Returns(Array.Empty<IMessageStore>());
+
+        var main = Substitute.For<IMessageStore>();
+        main.Uri.Returns(new Uri("wolverinedb://fake/main"));
+        main.Role.Returns(MessageStoreRole.Main);
+
+        var multiTenanted = new MultiTenantedMessageStore(main, theRuntime, theSource);
+
+        theStores = new MessageStoreCollection(theRuntime, [multiTenanted], []);
+    }
+
+    [Fact]
+    public async Task concurrent_callers_share_a_single_refresh()
+    {
+        var refreshIsRunning = new TaskCompletionSource(TaskCreationOptions.RunContinuationsAsynchronously);
+        theSource.RefreshAsync().Returns(refreshIsRunning.Task);
+
+        // The first caller starts the refresh and blocks on it; the rest arrive while it is still running
+        // and have to join it rather than start their own.
+        var first = theStores.FindAllAsync();
+        var joiners = Enumerable.Range(0, 19).Select(_ => theStores.FindAllAsync().AsTask()).ToArray();
+
+        await theSource.Received(1).RefreshAsync();
+
+        refreshIsRunning.SetResult();
+
+        await first;
+        await Task.WhenAll(joiners);
+
+        await theSource.Received(1).RefreshAsync();
+    }
+
+    [Fact]
+    public async Task a_second_enumeration_inside_the_stale_window_does_not_ask_again()
+    {
+        theRuntime.Options.Durability.TenantDatabaseListStaleTime = 30.Seconds();
+        theSource.RefreshAsync().Returns(Task.CompletedTask);
+
+        await theStores.FindAllAsync();
+        await theStores.FindAllAsync();
+        await theStores.FindAllAsync();
+
+        await theSource.Received(1).RefreshAsync();
+    }
+
+    [Fact]
+    public async Task the_list_is_refreshed_again_once_the_stale_time_has_elapsed()
+    {
+        theRuntime.Options.Durability.TenantDatabaseListStaleTime = TimeSpan.Zero;
+        theSource.RefreshAsync().Returns(Task.CompletedTask);
+
+        await theStores.FindAllAsync();
+        await theStores.FindAllAsync();
+
+        await theSource.Received(2).RefreshAsync();
+    }
+
+    [Fact]
+    public async Task a_failed_refresh_is_retried_by_the_next_caller()
+    {
+        theRuntime.Options.Durability.TenantDatabaseListStaleTime = 30.Seconds();
+        theSource.RefreshAsync().Returns(
+            Task.FromException(new TimeoutException("the operation has timed out")),
+            Task.CompletedTask);
+
+        await Should.ThrowAsync<TimeoutException>(async () => await theStores.FindAllAsync());
+
+        // A failure leaves the list unknown, so the window must not have opened.
+        await theStores.FindAllAsync();
+
+        await theSource.Received(2).RefreshAsync();
+    }
+}
--- a/src/Wolverine/DurabilitySettings.cs
+++ b/src/Wolverine/DurabilitySettings.cs
@@ -84,17 +84,32 @@ public class DurabilitySettings : IDescribeMyself
     public bool TenantRegistryRequired { get; set; }
 
     /// <summary>
-    ///     How long a tenant database list discovered from a dynamic tenancy source stays current
-    ///     before the next bulk enumeration re-reads it. Concurrent callers always share one refresh
-    ///     regardless of this value. The default is 5 seconds; zero refreshes on every call.
+    ///     How long a tenant database list discovered from a dynamic tenancy source stays current before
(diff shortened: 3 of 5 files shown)


jeremydmiller:
The `ThrottledTenantRefresh` primitive itself I went through carefully and couldn't fault — the lock ordering is sound, `RunContinuationsAsynchronously` avoids the inline-continuation traps, `_hasRefreshed`/`_refreshedAt` are set under the lock *before* `SetResult` so a caller arriving in that gap correctly takes the fast path, the failure path deliberately doesn't open the window, and `refreshAsync` catches everything so the discarded task can never fault. `_tenantRefreshes` is written only in the constructor (I checked — `_multiTenanted` has no other add path), so the plain `Dictionary` is safe for concurrent reads, and `MultiTenantedMessageStore` is a plain class so keying is reference equality.

One thing does need addressing before this merges.

## `FindDatabaseAsync` is throttled, and the PR says it isn't

The PR body says:

> **`FindDatabaseAsync` still forces a refresh on a miss.** That is the path where a newly provisioned tenant database has to be found immediately; the window only bounds how often the callers that enumerate *every* database go asking.

and the new XML doc on `TenantDatabaseListStaleTime` says the same:

> A lookup that misses still forces an immediate refresh, so a newly provisioned tenant database is never invisible for this long.

Neither holds. `FindDatabaseAsync` (`MessageStoreCollection.cs:254` on `46cbbecb3`) reaches the registry through `refreshTenantedDatabaseList` — which is the exact method the throttle was applied to:

```csharp
// FindDatabaseAsync, on a miss
foreach (var tenantedMessageStore in _multiTenanted)
{
    if (tenantedMessageStore.Source.Cardinality == DatabaseCardinality.DynamicMultiple)
    {
        await refreshTenantedDatabaseList(tenantedMessageStore);   // -> MaybeRefreshAsync()
    }
}
```

So if anything else on that node completed a refresh within `TenantDatabaseListStaleTime`, `MaybeRefreshAsync` returns `Task.CompletedTask` without touching the source, the second `_services.TryFind` misses again, and the method returns null.

### Where that lands hardest

`BuildAgentAsync` (`:365`), since `MessageStoreCollection` is an `IAgentFamily`:

```csharp
var database = await FindDatabaseAsync(uri);
if (database is IMessageStoreWithAgentSupport agentSupport) { return agentSupport.BuildAgent(wolverineRuntime); }
throw new ArgumentOutOfRangeException(nameof(uri), $"No database with Uri {uri} supports a durability agent");
```

In a cluster the leader discovers a newly provisioned tenant database on *its* refresh and assigns the durability agent to another node. That node's `_services` doesn't have the Uri yet — the forced refresh is precisely what used to find it. With the throttle, if anything on that node refreshed in the last 5 seconds the agent start throws instead.

It should recover on a later assignment attempt, so this is a transient failure plus an error log rather than a permanent one. But it is a regression on the one path the PR states it did not change, and it is reachable exactly when a fleet is provisioning tenants — the same fleets this PR is aimed at.

`findStoresAsync` has the same shape for `request.DatabaseUri`: a dead letter query scoped to one database Uri silently returns no results.

### Suggested shape

A `force: true` overload on `MaybeRefreshAsync` that still **joins an in-flight refresh** but never returns early on freshness, used by `FindDatabaseAsync`. That keeps the single-flight guard — which is the half that actually closes the connection storm, and the half your dumps justify — while making the documented behaviour true. Either way the `TenantDatabaseListStaleTime` doc comment needs to change.

## On your open question about the 5s default

I'd keep 5s rather than dropping to zero, **provided the above is fixed**. The single-flight guard is what stops the storm; the window is a bonus. But with `FindDatabaseAsync` currently caught in it, a non-zero default is exactly what makes the regression reachable — so the two decisions are coupled, and zero would be the safer default if the force path doesn't land.

## Two smaller things

- **`InitializeAsync` doesn't open the window.** It calls `tenantedMessageStore.Source.RefreshAsync()` directly (`:130`), so the first `FindAllAsync()` after startup performs a second full enumeration immediately. On the 512-database fleet you describe that is the expensive registry round trip twice back to back, during the window of peak connection pressure this PR exists to relieve. Seeding the throttle there — or a `MarkRefreshed()` the initializer calls after its direct refresh — closes it in a line.

- **`FindForTenantAsync` bypasses the guard.** #4268 merged a new *public* method that calls `Source.RefreshAsync()` directly, so it is an unguarded entry point onto the same registry round trip. Worth routing through the throttle when you rebase. (It has a separate problem too — #4273.)

Reviewed at `46cbbecb3`.

jeremydmiller:
## The CIPersistence failure is real, and it is the staleness window

`PersistenceTests.ModularMonoliths.registration_of_message_stores.using_dynamic_multi_tenancy` fails on this branch, and it is not flaky — it failed in-lane and again on the harness's isolated "retrying alone in a fresh process" pass.

```
services.Select(x => x.Uri).OrderBy(x => x.ToString())
    should be
[.../database1/..., .../database2/..., .../database3/..., .../database4/..., .../postgres/...]
    but was
[.../database1/..., .../database2/..., .../postgres/...]
```

`database3` and `database4` are missing. The test explains itself:

```csharp
await _host.AddTenantDatabaseAsync("t1", connectionString1);
await _host.AddTenantDatabaseAsync("t2", connectionString2);

var services = await collection.FindAllAsync();          // (1) refreshes; opens the window
services.Select(x => x.Uri)...ShouldBe([db1, db2, main]);

await collection.FindDatabaseAsync(db2);                 // hits _services, no refresh
await collection.FindDatabaseAsync(main);

await _host.AddTenantDatabaseAsync("t3", connectionString3);
await _host.AddTenantDatabaseAsync("t4", connectionString4);

services = await collection.FindAllAsync();              // (2) inside the 5s window -> stale
services.Select(x => x.Uri)...ShouldBe([db1, db2, db3, db4, main]);   // fails here
```

Everything between (1) and (2) takes well under five seconds, so the second enumeration is inside `TenantDatabaseListStaleTime` and `MaybeRefreshAsync` returns `Task.CompletedTask` without asking the source. The stale list comes back.

## This is broader than the finding I raised earlier

My earlier comment was about `FindDatabaseAsync`'s forced refresh-on-miss being throttled. That still stands and still needs the `force` path. But this failure is on **`FindAllAsync`**, and it is not a bug in the implementation — it is the feature working exactly as designed. The PR's own unit test `a_second_enumeration_inside_the_stale_window_does_not_ask_again` asserts precisely the behaviour that makes this integration test fail.

So the two tests encode contradictory contracts, and CI has surfaced a design question rather than a defect:

> Is a tenant database provisioned moments ago allowed to be invisible to `FindAllAsync` for up to five seconds?

The PR argues yes — *"A bulk enumeration that is a few seconds stale costs nothing — the next sweep sees the database."* For the durability sweeps that is true. For an application that provisions a tenant and then immediately does work for it, it is not, and this test is the existing expression of that expectation.

## Recommendation

**Default `TenantDatabaseListStaleTime` to `TimeSpan.Zero`, and let operators opt in.**

That keeps every existing behaviour byte for byte — this test passes untouched, provision-then-use keeps working, and `FindDatabaseAsync` keeps forcing on a miss — while still delivering the **single-flight guard**, which is the half that actually fixes what you reported. Re-reading your evidence: 233 give-ups over 176 Npgsql connect timeouts and 3 pool exhaustions, with the master database at 0.5% CPU and 150 of 3600 connections used. That is concurrent callers and retries each opening their own connection. Single-flight collapses that to one regardless of the window; the window only reduces how often the *one* refresh happens. The expensive part is the fan-out, not the frequency.

That also answers your open question about the default from the other direction: with the window at zero by default, its value stops being a correctness decision and becomes a tuning knob, which is a much easier thing to ship.

If you would rather keep a non-zero default, then this test has to be updated deliberately — waiting out the window, or setting the stale time to zero for that fixture — and the behaviour change wants calling out in the docs and CHANGELOG, because "a newly provisioned tenant database may not appear for N seconds" is a real semantic that applications can be relying on today.

Either way, `FindDatabaseAsync` still needs the force path from my earlier comment; that one is a straightforward regression regardless of what the default becomes.

jeremydmiller:
Pushed a follow-on commit (`81ceb278f`) that fixes the CIPersistence failure. Two changes.

## 1. `TenantDatabaseListStaleTime` now defaults to `TimeSpan.Zero`

I reproduced `using_dynamic_multi_tenancy` locally — it fails in about a second, deterministically — and the diagnosis is the one from my previous comment: the second `FindAllAsync()` lands inside the five second window and gets the stale list.

As I said there, that is the feature working as designed, not a defect. Rather than change `ThrottledTenantRefresh`, the default moves to zero. Every existing behaviour is preserved byte for byte, the integration test passes untouched, and the window becomes a tuning knob rather than a semantic change nobody opted into.

**Nothing is lost by this.** Re-reading your evidence — 233 give-ups over 176 Npgsql connect timeouts and three pool exhaustions, with the master database at 0.5% CPU and 150 of its 3600 connections in use — the cost was every concurrent caller and every retry opening its *own* connection to the registry. The single-flight guard collapses that to one regardless of the window, and single-flight is not configurable. The window only bounds how often that one refresh happens. The expensive part was the fan-out, not the frequency.

That also settles your open question about the default from the other direction: at zero, its value stops being a correctness decision.

## 2. `FindDatabaseAsync` now genuinely forces past the window

The behaviour your PR description and the `TenantDatabaseListStaleTime` docs both promised. It reaches the registry through `refreshTenantedDatabaseList`, which is the method the throttle was applied to, so its refresh-on-miss was being answered from the window.

`MaybeRefreshAsync(force: true)` skips the freshness check and **nothing else** — it still joins a refresh already under way, because a miss must never become a licence to open another connection. That is the property that keeps your fix intact, and it has its own test.

Why it matters beyond tidiness: `MessageStoreCollection` is an `IAgentFamily`, and `BuildAgentAsync` turns the null into `ArgumentOutOfRangeException: No database with Uri ... supports a durability agent`. Reachable when a leader assigns a newly provisioned tenant database's durability agent to a node whose cache is inside the window.

## Tests and verification

Two new tests in your existing file, each verified to fail without its change:

* `a_lookup_that_misses_forces_past_the_stale_window`
* `a_forced_lookup_still_joins_a_refresh_already_under_way`

All four of your original throttle tests pass unchanged — including `a_second_enumeration_inside_the_stale_window_does_not_ask_again`, which sets the stale time explicitly, so the design intent is fully preserved for anyone who opts in.

* `CoreTests`: 2770 passed
* `PersistenceTests`: **109/109**, up from 108 passed / 1 failed

Also added the CHANGELOG entry the branch was missing, now that the default is settled.

Please push back if you'd rather keep a non-zero default — it is your PR and that is a legitimate call. It would need `using_dynamic_multi_tenancy` updated deliberately (waiting out the window, or zeroing the stale time for that fixture) and the behaviour called out in the docs, since "a newly provisioned tenant database may not appear in a bulk enumeration for N seconds" is a real semantic applications can rely on today. I went with zero because it needs nobody to opt into a change they didn't ask for, while still delivering the fix you came for.

---

Housekeeping: pushing to your fork, I first mistakenly pushed the branch name to `JasperFx/wolverine` and created a stray `gh-4267/throttle-tenant-database-refresh` branch there. It points at the same commit as this PR's head and is harmless, but a repository ruleset refuses my deletion — a maintainer will need to remove it. Apologies for the litter.

Issue #4261, opened 2026-09-03
AdvisoryLock shares one NpgsqlConnection with no synchronisation; concurrent use desyncs the protocol and ReleaseLockAsync hangs uncancellably in CloseAsync


matee-9:
## Summary

`AdvisoryLock` (in `Wolverine.Postgresql/PostgresqlNodePersistence.cs`) keeps one long-lived `NpgsqlConnection` in `_conn` and **synchronises access to it with nothing** — no `SemaphoreSlim`, no `lock`, no `Interlocked`. `HasLock` performs *synchronous* I/O on that connection while `ReleaseLockAsync` / `DisposeAsync` / `TryAttainLockAsync` perform *async* I/O on the same field. Npgsql explicitly does not support concurrent use of one connection.

When those overlap, the protocol desyncs and shutdown hangs **uncancellably and forever** inside `NpgsqlConnection.CloseAsync()`, which has no `CancellationToken` overload — so `HostOptions.ShutdownTimeout`, correctly threaded down to `WolverineRuntime.StopAsync`, is powerless. Nothing below it can time out.

Unchanged as of `V6.33.0`: `PostgresqlNodePersistence.cs` is byte-identical between `V6.27.1` and `main` today.

## The hang

Two independent full process dumps, captured 2026-09-01 and 2026-09-02 from a CI test host. `dumpasync` returns **one unbroken chain** in each, and the two are **frame-for-frame identical including every `<>1__state` value** (`NextResult` 14, `Consume` 1, `ReleaseLockAsync` 1, `NodeAgentController.StopAsync` 1, `teardownAgentsAsync` 2, `WolverineRuntime.StopAsync` 5). Two runs landing on identical frames *and* identical state numbers is not coincidence.

```
Host.StopAsync -> Host.ForeachService<IHostedService>
  WolverineRuntime.StopAsync
    WolverineRuntime.teardownAgentsAsync
      NodeAgentController.StopAsync
        Wolverine.Postgresql.AdvisoryLock.ReleaseLockAsync
          AdvisoryLock.safeCloseConnectionAsync
            NpgsqlConnection.CloseAsync
              NpgsqlConnector.CloseOngoingOperations
                NpgsqlDataReader.Close -> Consume -> NextResult
                  awaiting ConfiguredValueTaskAwaiter<Npgsql.IBackendMessage>  <-- never completes
```

[...]

PR #4270, opened 2026-09-04
GH-4261 follow-up: a failed attain on the Oracle advisory lock leaked its connection

jeremydmiller:
Stacked on #4266 — it needs that PR's `_gate` and `closeWithinBudgetAsync`, so it cannot sit on `main`. **This means it gets no CI checks until #4266 merges and this is retargeted**; `dotnet.yml` only triggers on PRs to `main`.

## What was wrong

`OracleAdvisoryLock` is the one implementation that opens a connection **per lock** rather than keeping a single shared one, and `TryAttainLockAsync` only gave that connection back on two of its exits:

- **success** — retained in `_heldLocks` to hold the row lock, correct;
- **`ORA-00054` contention** — closed and disposed, correct.

Every other failure fell through to the outer `catch`, logged `"Error trying to attain advisory lock {LockId}"`, and returned `false` with the connection still open and — past `BeginTransactionAsync` — an open transaction on it, owned by nobody.

That is not a once-per-process cost. `TryAttainLeadershipLockAsync` fires on **every health-check tick**, so a failure mode that persists rather than resolving — the lock table's schema missing, credentials expired, a RAC node gone — leaks one connection per tick until the pool is exhausted, at which point the node cannot open a connection for anything else either.

## The fix

The connection and transaction now live in locals so the `catch` can finish what the `try` started. `discardFailedAttainAsync` rolls back within the same one-second bound `releaseLockUnsafeAsync` uses, then closes through `closeWithinBudgetAsync` and disposes. Nothing in it is allowed to throw — it runs from a catch whose whole job is to answer `false`.

It is reached **only when the lock was not attained**: both the success path and the `ORA-00054` path return from inside the `try`, so it can never tear down what `rememberId` has taken ownership of. It is also defensive about being handed something already torn down, which covers the one case that does reach it holding a connection — an `ORA-00054` teardown that itself threw.

## Verification

Red baseline established against this branch's parent with the production change reverted, not assumed:

| | unfixed | fixed |
|---|---|---|
| six attains against a missing schema, pool ceiling 3 | **15.2s** (fourth attain blocks the full 5s connection timeout on an empty pool, three times over) | **621ms** |
| pool can still hand out 3 connections afterwards | no | yes |

- `OracleTests` full suite: **166/166 green**.
- `dotnet build wolverine.slnx -c Release -f net9.0`: 0 warnings, 0 errors.

## Noted, not fixed here

While writing the test: on Oracle the **renewal** attain the heartbeat drives on every tick answers `false` for a lock this node already holds. The row lock the first attain took on its own connection blocks the second attain's `SELECT ... FOR UPDATE NOWAIT` with `ORA-00054` — confirmed by probe (`first=True hasAfterFirst=True second=False hasAfterSecond=True`, with an empty log, which is the signature of the silent ORA-00054 branch).

That path disposes its connection correctly, so it is not this leak. But it is a real difference from the sibling providers, where the renewal is re-entrant on the shared connection and reports `true` — and it is the same `TryAttainLeadershipLockAsync` return value that #4266's Postgres tests assert must stay `true` across renewals. Worth its own issue.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -91,6 +91,22 @@
   with the size stripped off, so a widened `varchar` was invisible to it and an existing table kept
   the narrow column forever.
 
+### WolverineFx.Oracle
+
+- **A failed attempt on the advisory lock gives its connection back.** `OracleAdvisoryLock` is the one
+  implementation that opens a connection per lock rather than keeping a single shared one, and
+  `TryAttainLockAsync` only released that connection on two of its exits: success, where the connection
+  is retained to hold the row lock, and `ORA-00054` contention, which closes and disposes it. Every
+  other failure fell through to the outer `catch`, logged, and returned false with the connection --
+  and, past `BeginTransactionAsync`, an open transaction on it -- owned by nobody.
+
+  That is not a once-per-process cost. `TryAttainLeadershipLockAsync` fires on every health-check tick,
+  so a failure mode that persists rather than resolving -- the lock table's schema missing, credentials
+  expired, a RAC node gone -- leaked one connection per tick until the pool was exhausted, at which
+  point the node could no longer open a connection for anything else either. The regression test drives
+  six attains against a pool of three: unfixed, it spends fifteen seconds queueing on an empty pool;
+  fixed, it finishes in well under one.
+
 ### WolverineFx.Http
 
 - **DataAnnotations validation works with `ServiceLocationPolicy.NotAllowed`.** (closes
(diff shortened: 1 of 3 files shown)


jeremydmiller:
Filed the renewal defect noted in the description as #4271. It turned out to be materially worse than "a difference from the sibling providers": an Oracle leader steps down and re-elects on a repeating cycle, observed as 2 stepdowns in 15s on a healthy single-node host (`the leadership advisory lock could not be renewed`). Three of the four siblings carry an explicit `if (hasLockUnsafe(lockId)) return true;` short-circuit that Oracle never got. Still not this PR's leak — that path disposes correctly.

PR #4274, opened 2026-09-04
Re-land #4270: a failed attain on the Oracle advisory lock leaked its connection

jeremydmiller:
Re-lands #4270, which merged but **did not reach `main`**.

#4270 was stacked on `fix/gh-4261-advisory-lock-connection-races` (the #4266 branch). #4266 was **squash**-merged into `main` as `7b882b768`, which left that feature branch alive but orphaned — its history is no longer an ancestor of `main`. #4270 was then merged into *that* branch, so its commit landed somewhere nothing pulls from. Neither `discardFailedAttainAsync` nor `Bug_4261_failed_attain_returns_its_connection` is on `main` today; that is straightforward to confirm:

```bash
git show origin/main:src/Persistence/Oracle/Wolverine.Oracle/OracleAdvisoryLock.cs | grep -c discardFailedAttainAsync   # 0
```

It also never ran CI, for the same reason — `dotnet.yml` only triggers on PRs to `main`, so #4270 showed *no checks reported* for its whole life.

## This PR

`6830cba18` rebased with `git rebase --onto origin/main 455ea3592`, dropping the unsquashed copy of #4266 that the branch still carried. The content is unchanged from what was reviewed on #4270 — one commit, the same three files, and #4266's `_gate` / `closeWithinBudgetAsync` now come from `main` rather than from the branch.

Verified after the rebase, not assumed:

- `OracleTests` full suite: **166/166 green** (`docker compose up -d oracle`).
- The `_gate` and `closeWithinBudgetAsync` the change depends on are present from `main`.

## What it fixes (unchanged from #4270)

`OracleAdvisoryLock` is the one implementation that opens a connection **per lock**, and `TryAttainLockAsync` only gave that connection back on two of its exits: success (retained in `_heldLocks` to hold the row lock) and `ORA-00054` contention (closed and disposed). Every other failure fell through to the outer `catch`, logged, and returned `false` with the connection — and past `BeginTransactionAsync`, an open transaction on it — owned by nobody.

`TryAttainLeadershipLockAsync` fires on every health-check tick, so a failure mode that persists rather than resolving (missing schema, expired credentials, a RAC node gone) leaked one connection per tick until the pool was exhausted, at which point the node could not open a connection for anything else either.

`discardFailedAttainAsync` now rolls back within the same one-second bound `releaseLockUnsafeAsync` uses, then closes and disposes. The regression test drives six attains against a pool of three: unfixed it spends ~15s queueing on an empty pool and cannot reclaim its connections afterwards; fixed it finishes in well under one.

## Review findings from #4270 are NOT addressed here

This is a faithful re-land so the fix stops being stranded. Three findings from the review of #4270 still stand and are worth a follow-up commit here before merge — see the review comment below.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -91,6 +91,22 @@
   with the size stripped off, so a widened `varchar` was invisible to it and an existing table kept
   the narrow column forever.
 
+### WolverineFx.Oracle
+
+- **A failed attempt on the advisory lock gives its connection back.** `OracleAdvisoryLock` is the one
+  implementation that opens a connection per lock rather than keeping a single shared one, and
+  `TryAttainLockAsync` only released that connection on two of its exits: success, where the connection
+  is retained to hold the row lock, and `ORA-00054` contention, which closes and disposes it. Every
+  other failure fell through to the outer `catch`, logged, and returned false with the connection --
+  and, past `BeginTransactionAsync`, an open transaction on it -- owned by nobody.
+
+  That is not a once-per-process cost. `TryAttainLeadershipLockAsync` fires on every health-check tick,
+  so a failure mode that persists rather than resolving -- the lock table's schema missing, credentials
+  expired, a RAC node gone -- leaked one connection per tick until the pool was exhausted, at which
+  point the node could no longer open a connection for anything else either. The regression test drives
+  six attains against a pool of three: unfixed, it spends fifteen seconds queueing on an empty pool;
+  fixed, it finishes in well under one.
+
 ### WolverineFx.Http
 
 - **A Static-mode endpoint type mismatch is now reported as itself.** (closes
--- a/src/Persistence/Oracle/OracleTests/Agents/Bug_4261_failed_attain_returns_its_connection.cs
+++ b/src/Persistence/Oracle/OracleTests/Agents/Bug_4261_failed_attain_returns_its_connection.cs
@@ -1,0 +1,134 @@
+using System.Diagnostics;
+using IntegrationTests;
+using Microsoft.Extensions.Logging.Abstractions;
+using Oracle.ManagedDataAccess.Client;
+using Shouldly;
+using Wolverine;
+using Wolverine.Oracle;
+using Wolverine.Persistence.Durability;
+using Wolverine.RDBMS;
+using Wolverine.RDBMS.Sagas;
+
+namespace OracleTests.Agents;
+
+/// <summary>
+/// GH-4261 follow-up. <c>OracleAdvisoryLock.TryAttainLockAsync</c> opens a connection and only gave it
+/// back on two of its exits: the success path, where the connection is retained in <c>_heldLocks</c>,
+/// and the ORA-00054 contention path, which closes and disposes it. Every other failure fell through to
+/// the outer <c>catch</c>, which logged and returned false with the connection -- and, once
+/// <c>BeginTransactionAsync</c> had run, an open transaction on it -- still held by nobody.
+///
+/// <para>That is not a once-per-process cost. <c>TryAttainLeadershipLockAsync</c> fires on every
+/// health-check tick, so any failure mode that persists rather than resolving -- the lock table's schema
+/// missing, credentials expired, a RAC node gone -- leaks one connection per tick until the pool is
+/// exhausted, at which point the node cannot open a connection for anything else either.</para>
+///
+/// <para>The pool ceiling here is deliberately tiny so the leak has to show. Against the unfixed code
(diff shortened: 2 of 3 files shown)


jeremydmiller:
Review of the change carried over from #4270 (which never ran CI, so this is the first time any of it is being checked).

The diagnosis is right, and I verified the mechanism rather than taking it on trust: the MERGE's `catch (OracleException) { }` swallows ORA-00942 from the missing schema, `BeginTransactionAsync` then succeeds, and the `SELECT ... FOR UPDATE NOWAIT` raises ORA-00942 — not 54 — so it falls to the outer catch. That is genuinely the leak path the test arranges. The red baseline was established against the parent with the production change reverted rather than assumed. `OracleContext` is `[Collection("oracle")]`, so the two tests cannot race each other on the lock row.

Three findings.

## 1. Dispose after an abandoned close can block (`OracleAdvisoryLock.cs`, `discardFailedAttainAsync`)

`discardFailedAttainAsync` ignores `closeWithinBudgetAsync`'s return value and calls `conn.DisposeAsync()` unconditionally. The other four providers deliberately do not — `PostgresqlNodePersistence.cs` on `main`:

```csharp
if (conn.State == ConnectionState.Open && !await closeWithinBudgetAsync(conn).ConfigureAwait(false))
{
    // Deliberately no DisposeAsync: that would race the CloseAsync still parked on this connection.
    return;
}
await conn.DisposeAsync();
```

SqlServer, MySQL and Sqlite match that. Oracle does not, here or on the ORA-00054 path. Oracle's own doc on `closeWithinBudgetAsync` gives the rationale: *"the caller's DisposeAsync then runs against a connection whose close never completed, which the surrounding catch absorbs."*

That only covers a **throw**. The failure the budget exists for is a close that never *returns*, and ODP.NET's `DisposeAsync` closes the connection — so on precisely the wedged connection the budget just abandoned, dispose can block on the same thing, indefinitely, with `_gate` still held. That is the GH-4261 hang reintroduced, on the one path that by definition runs against a connection that has just misbehaved.

To be clear about ownership: **I introduced this pattern into Oracle in #4266**, and the divergence from the other four is mine, not this PR's. It is worth closing here because `discardFailedAttainAsync` is the method being added — check the return, skip the dispose when the close was abandoned, and bring Oracle's doc comment in line with the other four.

## 2. The discard can still run after `rememberId` took ownership

The comment states the outer catch is *"reached only when the lock was NOT attained: the success and ORA-00054 paths both return from inside the try."* That invariant does not quite hold.

`ensureCmd` and `lockCmd` are `await using var` in the **outer** try block. `return true` does not leave the method before their disposals run — the runtime disposes them as it unwinds that scope. If either `DisposeAsync` throws, the exception is raised *within* the outer try, the outer catch handles it, and `conn`/`tx` are still non-null locals — so `discardFailedAttainAsync` rolls back the transaction holding the row lock and disposes the connection `rememberId` just registered in `_heldLocks`.

Result: a lock id in `_locks`/`_heldLocks` pointing at a disposed connection, and `false` returned for a lock that was actually attained. It self-heals (the next `HasLock` pings the disposed connection, throws, and `forgetId` prunes it), and `OracleCommand.DisposeAsync` throwing is uncommon — so this is narrow. But it is cheap to close and it makes the comment true: set an `attained` flag alongside `rememberId` (or null the locals there, since ownership has moved) and have the catch skip the discard when it is set.

## 3. The wall-clock assertion is flaky and redundant

```csharp
stopwatch.Elapsed.ShouldBeLessThan(PoolWait, ...);
```

Six attains against a real Oracle, each opening a pooled connection, running a MERGE that raises and swallows ORA-00942, beginning a transaction, running a `SELECT ... FOR UPDATE NOWAIT` that raises ORA-00942, and unwinding through two catch blocks plus the new discard. Measured at 621ms on a warm local container; CIOracle runs `gvenzl/oracle-free` in an Actions container alongside the rest of the lane, where a cold buffer cache or a noisy runner can cost several times that. The 8x margin is a measured number, not an argument.

A red build here would read *"six attains took longer than five seconds"*, which looks like a leak and is not one — the worst kind of flake, because it points at the wrong thing.

The pool-exhaustion assertion immediately below it is exact: against the leak, three connections are stranded, `MaxPoolSize` is 3, and the reclaim loop blocks for `ConnectionTimeout` and throws. It proves the same property, fails for the right reason, and cannot be timing-dependent. Suggest dropping the stopwatch assertion, or keeping the stopwatch only to enrich the pool assertion's failure message.

## Separately

The renewal behaviour #4270 flagged in "Noted, not fixed here" is filed with its own analysis — it is more serious than that note suggested, and it is not caused by either PR.

Issue #3987, opened 2026-08-20
Agent assignment churns on every pod during a Kubernetes rolling deploy, leaving agents assigned but not running


erdtsieck:
## What I see

On a Kubernetes rolling deploy, agent assignment reacts to every individual pod coming and going. With a large agent count that turns a routine 3-pod rollout into a reassignment storm, and some agents end up not running at all afterwards — assigned in `wolverine_node_assignments`, but never actually started.

Our fleet: 512 shard databases, ~1650 active tenants, six async projections plus one subscription, so roughly 13.5k distributed agents. `Daemon:Mode` is `Distributed`, `UseWolverineManagedEventSubscriptionDistribution` is on, database-affine assignment is always-on since jasperfx/jasperfx#487. Wolverine 6.29.1, Marten 9.28.0, JasperFx 2.52.1.

## Measurements from today

Two rolling restarts of a 3-replica deployment, roughly 90 minutes apart. From `wolverine_node_records`:

| event | records |
|---|---|
| `AssignmentChanged` | 24 257 |
| `AgentStarted` | 22 389 |
| `AgentStopped` | 2 581 |
| `AgentPaused` | 769 |
| `NodeStarted` | 2 |
| `DormantNodeEjected` | 1 |

Six pod replacements produced 24k assignment changes and 22k agent starts. One node was also ejected as dormant during the churn.

We had just introduced one new Marten subscription, which adds one agent per tenant (~2400). For that subscription alone:

| event | records | distinct agents |
|---|---|---|
| `AssignmentChanged` | 3 345 | 3 208 |
| `AgentStarted` | 3 115 | 2 366 |
| `AgentPaused` | 746 | 653 |
| `AgentStopped` | 355 | 349 |

The outcome that made us look: **798 of 1655 tenants had a checkpoint row in `mt_event_progression` and that number did not move for over half an hour.** The remaining tenants were not idle — two I sampled had 2158 and 2417 matching events waiting. `AgentPaused` log volume had already collapsed by then, so these were not paused agents; they simply were not running. A second `kubectl rollout restart` got them going.

`wolverine_node_assignments` looked healthy throughout, which is the trap: it records who *should* own an agent, not who is running one. That is also what made #3698 hard to see.

## What I would like

Something that recognises "the cluster is mid-rollout" and does not rebalance per pod event. Concretely, any of these would help us:

[...]

Issue #3959, opened 2026-08-16
Agent distribution has no per-node capacity ceiling: one node dying pushes its whole share onto the survivors


erdtsieck:
## Summary

`AssignmentGrid` distributes agents across every *live* node without any notion of how much a node can actually host. When a node dies, its share is redistributed over the survivors — which can push them past their own limits, killing them in turn. There is no floor to that process, so a cluster under memory pressure cannot converge, and adding replicas makes it worse rather than better.

We hit this on production last weekend: 34 hours of intermittent unavailability, restart counters up to 334, and no self-recovery.

## What already works

To be clear about what I'm *not* asking for: `EventSubscriptionAgentFamily.EvaluateAssignmentsAsync` already does the right thing for sharded stores. Multi-database stores get `DistributeByGroupAffinity(SchemeName, DatabaseKeyOf, …)`, keeping all of a shard database's per-tenant agents together on one node so pools scale with databases rather than nodes × databases (jasperfx#486 / marten#4806). That behaviour is exactly right for our topology and it worked.

## The gap

Neither distributor bounds what a node takes on.

`AssignmentGrid.Distribution.cs`:

```csharp
var spread = (double)agents.Count / _nodes.Count;
var minimum = (int)Math.Floor(spread);
var maximum = (int)Math.Ceiling(spread);
```

and group affinity places groups "largest-first onto the least-loaded node". Both balance, neither caps. The implicit assumption is that any live node can host whatever share it is handed.

That assumption breaks whenever an agent's cost is not uniform over time.

## What happened

Topology: Marten with a sharded store, 512 shard databases, per-tenant event partitioning. Six projections × 1.463 tenants = **8.792 agents** across **6 nodes** (one pod per node, enforced by pod anti-affinity).

A release bumped two projection versions, so those read models rebuilt from event 0 — about 381M events. We measured that an agent *working off a backlog* holds roughly **4.4 MB**, against roughly **1 MB** for an agent that is caught up. Same agents, same code, four times the memory, purely as a function of having work to do.

At ~1,465 agents per node that is ~6.4 GB during catch-up, against a .NET GC hard limit of 9,216 MiB. Tight but survivable — until one node went over and died. Its ~1,465 agents were then spread over five survivors, taking each of them to ~1,760 agents, over the limit. Then four. Then three.

Two properties made it unrecoverable:

[...]

PR #4297, opened 2026-09-04
GH-3959: capacity-aware agent assignment

mlh758:
Rebased onto current `main` and narrowed to the capacity work. The node-side reconciliation
sweep this PR originally carried shipped separately in #4404 (hardened in #4408), so nothing
of GH-3987 is left here.

## GH-3959: capacity-aware agent assignment

Opt-in via `CapacityAwareAssignment`, off by default. Each node samples its own load through
`INodeLoadMonitor` — the default reads memory pressure from `Environment.WorkingSet` against
the GC budget — and advertises it on every heartbeat. The leader then:

- prefers the least-loaded nodes;
- never places onto a node at or above `NodeOverloadThreshold` (default 90);
- sheds at most `OverloadShedBatchSize` agents per scheme per evaluation off overloaded nodes,
  never a pinned one;
- leaves agents unassigned when no node has headroom, rather than piling them onto a survivor
  that cannot start them.

Placement stops a 10-point hysteresis band below the shed line, so the two passes cannot
oscillate around one threshold. A node advertising no load always counts as having headroom,
keeping stores without load persistence on today's behavior.

The advertisement persists in a new `wolverine_nodes` column provisioned only when the feature
is on, leaving existing schemas and `AutoCreate` settings untouched. PostgreSQL today; other
stores simply never advertise.

## Why

In the GH-3959 incident shape — three nodes collapsing to one — the survivor accepted every
assignment, ran out of memory starting them, and the leader re-decided forever: ~125
`AssignmentChanged` rows/minute with sustained `OutOfMemoryException`. Now the survivor holds
at its ceiling, the deficit surfaces as waiting agents, and recovery on scale-out is immediate.

All defaults preserve existing behavior.

### DIFF
--- a/src/Persistence/PostgresqlTests/Agents/capacity_aware_column_is_opt_in.cs
+++ b/src/Persistence/PostgresqlTests/Agents/capacity_aware_column_is_opt_in.cs
@@ -1,0 +1,157 @@
+using IntegrationTests;
+using JasperFx;
+using JasperFx.Resources;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Hosting;
+using Npgsql;
+using Shouldly;
+using Weasel.Postgresql;
+using Wolverine;
+using Wolverine.Postgresql;
+using Wolverine.Runtime;
+using Wolverine.Runtime.Agents;
+
+namespace PostgresqlTests.Agents;
+
+/// <summary>
+/// GH-3959. The load_factor column and every statement naming it are gated on
+/// CapacityAwareAssignment, so upgrading without opting in migrates nothing and reads nothing.
+/// </summary>
+public class capacity_aware_column_is_opt_in : PostgresqlContext
+{
+    private static CancellationToken Ct => TestContext.Current.CancellationToken;
+
+    private static Task<IHost> startAsync(string schema, bool capacityAware, AutoCreate autoCreate)
+    {
+        return Host.CreateDefaultBuilder()
+            .UseWolverine(opts =>
+            {
+                opts.PersistMessagesWithPostgresql(Servers.PostgresConnectionString, schema);
+                opts.Durability.CapacityAwareAssignment = capacityAware;
+                opts.AutoBuildMessageStorageOnStartup = autoCreate;
+
+                if (autoCreate != AutoCreate.None)
+                {
+                    opts.Services.AddResourceSetupOnStartup();
+                }
+            }).StartAsync(Ct);
+    }
+
+    private static async Task<long> loadFactorColumnCountAsync(string schema)
+    {
+        await using var conn = new NpgsqlConnection(Servers.PostgresConnectionString);
+        await conn.OpenAsync(Ct);
+        await using var cmd = conn.CreateCommand();
+        cmd.CommandText =
+            $"select count(*) from information_schema.columns where table_schema = '{schema}' and table_name = 'wolverine_nodes' and column_name = 'load_factor'";
+        return Convert.ToInt64(await cmd.ExecuteScalarAsync(Ct));
+    }
+
+    private static async Task dropSchemaAsync(string schema)
+    {
+        await using var conn = new NpgsqlConnection(Servers.PostgresConnectionString);
+        await conn.OpenAsync(Ct);
+        await conn.DropSchemaAsync(schema, ct: Ct);
+    }
+
+    [Fact]
+    public async Task no_column_and_a_working_agent_plane_when_the_flag_is_off()
+    {
+        const string schema = "capacity_off";
+        await dropSchemaAsync(schema);
+
+        var host = await startAsync(schema, capacityAware: false, AutoCreate.CreateOrUpdate);
+
+        try
+        {
+            var runtime = host.Services.GetRequiredService<IWolverineRuntime>();
+
+            (await runtime.Storage.Nodes.MarkHealthCheckAsync(WolverineNode.For(runtime.Options), Ct))
+                .ShouldBeTrue();
+
+            var nodes = await runtime.Storage.Nodes.LoadAllNodesAsync(Ct);
+            nodes.ShouldNotBeEmpty();
+            nodes.ShouldAllBe(x => x.LoadFactor == null);
+
+            await runtime.Storage.Nodes.LoadNodeAgentStateAsync(Ct);
+        }
+        finally
+        {
+            await host.StopAsync(Ct);
+            host.Dispose();
+        }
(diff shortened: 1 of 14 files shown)


mlh758:
I'll address those review comments and rebase so that if you want only the code changes it can still be easy to pull out.

Edit: Fixed

jeremydmiller:
@mlh758 Hey man, can you please back off a little bit? And your characterization of the leadership election in the docs was off anyway. It's a semi-bully algorithm that always looks for the known node with the lowest node number and uses the advisory lock *only* as a way of taking the leadership. Really a back up. 

One more time, I will not take the "P" thing. I cannot afford to have some kind of unusual alternative testing framework nobody knows that will instantly get out of sync with the code.

I'm not enthusiastic about any of this. What are you trying to do here? Is this for your own work, or @erdtsieck's issues? Anne's usage probably needs to be dealt with in the context of the Marten async daemon rather than through the IAgent mechanics IMO.

mlh758:
I was just trying to show value for it, I didn't realize I was annoying you that hard. I kept the commit with the actual code change separate figuring you might say no. I just thought that a more interesting issue would show some value for modeling like that. I've rebased and now this PR only has the changes. I won't send anything about P or TLA+ at you again.

I picked those issues because they looked interesting and I knew there was similar work I could reference in Orleans. I'm using these projects so I wanted to contribute back. I also knew a simulator/harness for the issue should be possible to create and verify against.

If you don't want this either that's alright and I can back off. I'm just trying to not be someone who makes requests and never sends anything back.

Edit: tldr, sorry, I got too enthusiastic and I see how that was a pain.

jeremydmiller:
I'm not 100% opposed to the actual approach, but I think there are some other, much simpler ameliorations for the Kubernetes thing I'd rather try first. Just simply saying "don't bother assigning anything until at least this much time has passed or this many nodes exist" would help a ton. 

The async daemon thing for Anne that leads to the 10k agents is unique, and I think I'd rather pay more attention to finding ways of balancing work inside of the daemon layer than fighting that at the Wolverine agent distribution level beyond what we've already done.

The "P" thing? I had asked you not to do that before. And the first thing I'm going to do is have Claude translate that into a large battery of tests that mimic that, but in a way that's idiomatic to the project. 

By and large, don't try to introduce all new testing or build automation or GitHub workflows to an OSS project as a first time contributor w/o discussing that first. I've been burned by that before, and it's a personal pet peeve of mine because of past experiences.

And sorry for being grumpy

mlh758:
> The "P" thing? I had asked you not to do that before. And the first thing I'm going to do is have Claude translate that into a large battery of tests that mimic that, but in a way that's idiomatic to the project.

I thought I had picked a bad example with leader election and tried to go for something more meaningful that might actually help out.

> By and large, don't try to introduce all new testing or build automation or GitHub workflows to an OSS project as a first time contributor w/o discussing that first. 

I've been a programmer for a while, I should have known better. I've been rewatching some old strange loop talks where formal methods were brought up and I though, "well, AI might actually make this feasible to introduce and help out!" and since I was using Wolverine and it wouldn't touch existing code I thought I could try it out.

> And sorry for being grumpy

I'm the one who threw a giant PR at you with unknown technology, don't worry about it. I've been on the other end of that and it wasn't fun. I saw you communicating with AI a lot and wrongly assumed it wouldn't cause you much extra burden.

> The async daemon thing for Anne that leads to the 10k agents is unique, and I think I'd rather pay more attention to finding ways of balancing work inside of the daemon layer than fighting that at the Wolverine agent distribution level beyond what we've already done.

I'd be interested in talking more about that if you want.

mlh758:
Rebased this and dropped the changes for `AssignmentStabilityWindow` since #4367 introduced something similar. Re-ran against the simulator repo to make sure it didn't regress.

jeremydmiller:
Hey, I think the most likely thing is that we take the reproduction harness but not the fix. There'd have to be more than this anyway because we can't ship breaking database changes w/o doing a major version bump. 

And I think the priority needs to be in other areas as I was trying to say earlier.

mlh758:
I could guard the use of the new column on using the option and document that it requires running a migration to use the feature. I should have done that in the first place to make the PR acceptable. (Edit: there is a guard now so that the column is only used if the feature is enabled).

If it's in the wrong place though then that's not worth doing. Where would you rather see a mitigation added?

Either way, you're welcome to the reproduction code.

mlh758:
The other thing this PR has is:

> Compares this node's durable assignment rows against its actually-registered agents and heals the divergences the leader cannot see

This was another thing mentioned in the source issues where it seemed like pods would come up and not all the agents would start. I've also had runs in the simulator repo where after some rollouts there are duplicate agents. There's probably a deeper fix there but the reconciliation would at least heal it. I could pull that out of this PR and open one separately.

mlh758:
@jeremydmiller okay I've rebased this PR and cleaned up the title/description. I'll make one more pitch for it and then close it out if you're not satisfied. Here's the highlight from my testing in that sim repo:

image

The new settle period helps with churn during rollout but doesn't change the behavior of losing nodes. If a cluster is under pressure, losing a node can still cause a cascade since a departed node will immediately trigger re-balancing. A wait period might not help even if it didn't though since it could take a while for a new pod to come up if there are other issues with the cluster like nodes being rotated out for maintenance.

Since Wolverine can host custom agents it seems like the spot that would be best aware of how to allocate them, and there's now a guard to keep the new column from being used unless you opt into the feature. I could also be missing a better spot though.

jeremydmiller:
Thanks @mlh758 — re-verified today: still `MERGEABLE`/`CLEAN` against current `main` with all checks green.

We're not landing this in the current pass, but it's queued for the next one. Plan when we pick it up is a review of the placement/shed hysteresis logic in `AssignmentGrid.Distribution` and of the opt-in `wolverine_nodes` column provisioning, then merge.

Also noting for the record that the GH-3987 half shipped separately in #4404 (hardened in #4408), so despite the branch name this PR is purely GH-3959 — that matches your description, just confirming we read it the same way.

mlh758:
Sounds good, let me know if there are any changes you want against the branch in the mean time. I tried to make it passive per your previous comment but that's the part I'm most worried about having gotten right.

mlh758:
@jeremydmiller that simulator project I keep linking is no longer a hellscape of random bash scripts and adhoc python parsing if you want to run any of those changes against it for those new tickets. Or tell me to on a PR. Happy to check changes against it for you.

I've also added RavenDB and MySQL backends to it for most of the tests.

PR #4361, opened 2026-09-06
/claim #4360: Use ephemeral ports in HTTP tests to avoid macOS AirPlay port 5000 collision

### LINKED ISSUE #4360: Bug_using_host_stop hard-binds port 5000 and cannot pass on a Mac with AirPlay Receiver enabled
Found while running the full supervised CI suite locally (GH-4316 perf wave verification).

`CIHttp` reports 1016 passed, 2 failed. Both failures are the same cause:

```
Wolverine.Http.Tests.Bugs.Bug_using_host_stop.wolverine_runtime_stops_when_host_is_disposed(type: WebApplicationBuilder)
Wolverine.Http.Tests.Bugs.Bug_using_host_stop.wolverine_runtime_stops_when_host_is_stopped(type: WebApplicationBuilder)

System.IO.IOException : Failed to bind to address http://127.0.0.1:5000: address already in use.
---- Microsoft.AspNetCore.Connections.AddressInUseException
-------- System.Net.Sockets.SocketException : Address already in use
```

Port 5000 on this machine is held by **macOS ControlCenter** — that is the AirPlay Receiver, which listens on 5000 (and 7000) and is **on by default** on current macOS:

```
COMMAND    PID  USER          TYPE  NAME
ControlCe  723  jeremymiller  IPv4  TCP *:5000 (LISTEN)
```

So these two tests cannot pass on any Mac with the default AirPlay setting. The workaround is System Settings → General → AirDrop & Handoff → turn off AirPlay Receiver, but that is asking every macOS contributor to change a system setting to run the suite.

**Suggested fix**: bind port 0 and read back the assigned port, rather than hard-coding 5000. The tests are about runtime shutdown behaviour (`wolverine_runtime_stops_when_host_is_disposed` / `..._is_stopped`) — the specific port is incidental to what they assert.

Not urgent, and not related to any current change: it reproduces on clean `main`. Filing so the next person who sees a red `CIHttp` locally does not spend time hunting a phantom regression.

jialfaro:
Summary:
- Force test hosts to bind to an ephemeral port (port 0) instead of the platform default (5000).
- Applied for WebApplicationBuilder-created hosts and AlbaHost factory hosts; environment variable changes are temporary and restored.

Why:
- On macOS the AirPlay Receiver (ControlCenter) commonly listens on port 5000, causing test failures with "address already in use". These tests only need a runnable host and do not depend on a specific port.

Files changed:
- src/Http/Wolverine.Http.Tests/Bugs/Bug_using_host_stop.cs

How to verify:
- On a macOS machine with AirPlay Receiver enabled (or any machine where port 5000 is in use), run the affected tests:
  - dotnet test src/Http/Wolverine.Http.Tests -f net7.0 --filter "Bug_using_host_stop"
- Expect the two tests in Bug_using_host_stop to pass instead of failing with an "address already in use" IOException.

Closes #4360

Closes #4360

### DIFF
--- a/src/Http/Wolverine.Http.Tests/Bugs/Bug_using_host_stop.cs
+++ b/src/Http/Wolverine.Http.Tests/Bugs/Bug_using_host_stop.cs
@@ -1,12 +1,14 @@
-﻿using Alba;
+using Alba;
 using JasperFx.Core.Reflection;
 using Microsoft.AspNetCore.Builder;
+using Microsoft.AspNetCore.Hosting;
 using Microsoft.Extensions.Hosting;
 using Shouldly;
 using System.Reflection;
 using Microsoft.Extensions.DependencyInjection;
 using Wolverine.Runtime;
 using Wolverine.Tracking;
+using System;
 
 namespace Wolverine.Http.Tests.Bugs;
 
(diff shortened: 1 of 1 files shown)


jeremydmiller:
@jialfaro Nice, thank you!

PR #4367, opened 2026-09-06
GH-3987: let the leader wait for the node set to settle before assigning agents

### LINKED ISSUE #3987: Agent assignment churns on every pod during a Kubernetes rolling deploy, leaving agents assigned but not running

erdtsieck:
Toward #3987. This is the "much simpler amelioration" you described on #4297:

> Just simply saying "don't bother assigning anything until at least this much time has passed or this many nodes exist" would help a ton.

It is deliberately just that, and nothing else — no capacity model, no reconciliation, no new test tooling. It does not close #3987 on its own and it does not overlap with #4297's capacity work.

## The problem

A rolling deploy replaces pods one at a time. On our last rollout the three new pods came up at 19:02:31, 19:03:12 and 19:03:52 while the two old ones were still serving. The leader assigns agents against whatever nodes exist at that instant, so every pod that joins afterwards triggers another rebalance that moves agents again.

At 512 shards, ~2 500 tenants and one agent per tenant per projection, each of those waves is thousands of `AssignmentChanged` rows and thousands of agent starts and stops. After that rollout the new projection's agents arrived gradually — 735 tenants had one at 19:12, 2 310 at 19:49, and the full 2 556 only at 19:56. Roughly fifty minutes from rollout to complete coverage, with `AgentStartBatchSize` and `MaxAgentStartParallelism` already at 400 and `CheckAssignmentPeriodSeconds` at 15.

Nothing today waits: `FirstNodeReassignmentExecution` governs reassigning persisted *messages* from offline nodes, and `AgentStartBatchSize` / `MaxAgentStartParallelism` / `CheckAssignmentPeriod` / `FirstHealthCheckExecution` change how fast and how often assignment runs, never how long the leader waits before the first pass.

## The change

Three settings on `DurabilitySettings`, all off or inert by default:

```csharp
opts.Durability.AssignmentSettlePeriod = TimeSpan.FromSeconds(30);  // default TimeSpan.Zero — off
opts.Durability.AssignmentSettleNodeCount = 0;                      // default — off
opts.Durability.MaxAssignmentSettleTime = TimeSpan.FromMinutes(1);  // default
```

With `AssignmentSettlePeriod` at zero — the default — `trackMembershipAndDecideWait` returns `false` on every tick and the behaviour is bit-for-bit what it is today.

Set it, and the leader holds its assignment pass back until the node set has been unchanged for that long. The wait ends early once `AssignmentSettleNodeCount` nodes are live (the "or this many nodes exist" half — the fast path for a cold start of a deployment whose replica count you know), and `MaxAssignmentSettleTime` caps it so a cluster that keeps gaining nodes is still served.

Two properties that seemed worth getting right:

**A node leaving is never waited out.** Its agents are running nowhere, so a departure assigns immediately and cancels any wait in progress. That is what keeps a real node loss — and a leader takeover — exactly as responsive as it is today; only joins are debounced.

**Membership is tracked on every node, whatever its role.** A follower that keeps this warm already knows the cluster it inherits the moment it wins an election, so a new leader can tell "the old leader died" (a departure — act now) from "the cluster is still coming up" (a join — wait). Without that, every new leader would see a brand-new roster and wait out the full period at the worst possible moment.

The gate sits on both leader-side calls into `EvaluateAssignmentsAsync` — the `IsLeader` branch and the first pass inside `tryStartLeadershipAsync`. It never touches leadership attainment or lease renewal, so a waiting leader still holds its lock normally.

## Trade-off

During the wait no agents are assigned at all, so a cold start runs nothing until it elapses. That is the honest cost and it is why the default is off, why the departure case is exempt, and why both the node-count and the hard cap are there. Sizing guidance is in the docs page.

## Testing

Six tests in `CoreTests.Runtime.Agents.waiting_for_the_node_set_to_settle`, on the existing `NodeAgentController` unit harness with a frozen `TimeProvider`: default-off assigns on the first tick, the first pass is held back, a join restarts the wait, quiet ends it, the expected node count ends it early, a departure assigns immediately, and the cap serves a cluster that never settles. Full `CoreTests.Runtime.Agents` namespace is green (375 tests).

Happy to change any of the three names, or to drop `AssignmentSettleNodeCount` or the cap if you would rather ship the timer alone.

### DIFF
--- a/docs/guide/durability/marten/distribution.md
+++ b/docs/guide/durability/marten/distribution.md
@@ -221,6 +221,37 @@ Only `Other` is treated as potentially self-healing, so it is the only category 
 will auto-restart. The rest are left alone until you resolve the underlying problem, at which point
 restarting or rewinding the agent picks it back up.
 
+## Waiting for the Node Set to Settle <Badge type="tip" text="6.x" />
+
+A rolling deploy brings pods up one at a time. The leader assigns agents against whatever nodes exist at
+that instant, so every pod that joins afterwards triggers another rebalance that moves agents again. On a
+large agent universe — database-per-tenant Marten with one subscription agent per tenant per projection —
+each of those waves is thousands of agent starts and stops, and the fleet is below full throughput until
+the last one lands.
+
+`AssignmentSettlePeriod` makes the leader wait for the node set to go quiet before it assigns:
+
+```csharp
+opts.Durability.AssignmentSettlePeriod = TimeSpan.FromSeconds(30);  // default: TimeSpan.Zero, off
+opts.Durability.MaxAssignmentSettleTime = TimeSpan.FromMinutes(1);  // default
+opts.Durability.AssignmentSettleNodeCount = 0;                      // default, off
+```
+
+Only nodes *joining* are waited out. A node that leaves has taken its agents with it, so that work is
+running nowhere — a departure is assigned immediately and cancels any wait in progress, which is what keeps
+a real node loss (or a leader takeover) as responsive as it was before. `MaxAssignmentSettleTime` bounds the
+wait from the first join, so a cluster that keeps gaining nodes — an autoscaler ramping up, or a node
+flapping — is still served.
+
+`AssignmentSettleNodeCount` ends the wait early: once that many nodes are live the leader assigns without
+waiting the period out. Use it on a cold start of a deployment whose replica count you know, where the timer
+is then only the fallback for a pod that never arrives. Leave it off for a rolling deploy — the outgoing
+pods are still live and counted, so the target is met on the first tick and nothing is ever waited out.
+
+The cost of a non-zero settle period is that a cold start runs no agents at all until it elapses, which is
+why the default is off. Size it against the rollout it is meant to absorb: tens of seconds for a deploy whose pods
+come up within a minute of each other.
+
 ## Agent Start Retries <Badge type="tip" text="6.x" />
 
 An agent's very first assignment can race the subsystems it depends on coming up — an event-subscription
--- a/src/Testing/CoreTests/Runtime/Agents/waiting_for_the_node_set_to_settle.cs
+++ b/src/Testing/CoreTests/Runtime/Agents/waiting_for_the_node_set_to_settle.cs
@@ -1,0 +1,228 @@
+using CoreTests.Transports;
+using JasperFx.Core;
+using Microsoft.Extensions.Logging.Abstractions;
+using NSubstitute;
+using Wolverine.Configuration;
+using Wolverine.Runtime;
+using Wolverine.Runtime.Agents;
+using Xunit;
+
+namespace CoreTests.Runtime.Agents;
+
+/// <summary>
+/// A rolling deploy brings pods up one at a time, and the leader assigns agents against whatever nodes
+/// exist at that instant, so every pod that joins afterwards triggers another rebalance. On a large agent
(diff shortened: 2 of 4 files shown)


mlh758:
#4297 has a little of both. `AssignmentStabilityWindow` is the stability window to prevent churn during a rolling deploy. `CapacityAwareAssignment` is for #3959 to keep a failure from spreading and knocking over other nodes. If this PR merges I'll take out `AssignmentStabilityWindow` since they're targeting the same thing.

[The churn simulator](https://github.com/mlh758/wolverine-churn-simulator) wasn't tied to the modeling work, I wanted something to test changes against locally. It's just bash scripts to bring up some pods in minikube, poke them and measure what happens without having to deploy anything remotely.

PR #4402, opened 2026-09-09
Link retry attempts to the prior failed attempt's Activity via ActivityLink

### LINKED ISSUE #4398: [OTEL] Retries don't correlate to the failed attempt they follow - sibling spans, not linked spans
## Summary

When a message fails and is retried (inline, scheduled, or requeued), Wolverine starts a brand
new `Activity` for the next attempt that is only ever a sibling of the failed attempt's
`Activity` — both hang off whatever `Envelope.ParentId` the envelope originally carried on
receipt. There's no trace-level reference from "attempt 2" back to "attempt 1", beyond an
`ActivityEvent` (`wolverine.envelope.retried` / `wolverine.envelope.rescheduled` /
`wolverine.envelope.requeued`) added to the failed attempt's own span, which does nothing to
help you navigate forward from that failure to what happened next.

In a trace backend (Jaeger, Tempo, Application Insights, etc.) this means retries of the same
logical message show up as unrelated spans scattered across the trace of whatever originally
published the message — you can't click from "attempt 1 failed with `TimeoutException`" to
"here's attempt 2, which succeeded."

## Where this lives in the code

[`WolverineTracing.StartEnvelopeActivity`](https://github.com/JasperFx/wolverine/blob/main/src/Wolverine/Runtime/WolverineTracing.cs)
parents every new envelope `Activity` off `Envelope.ParentId` only.

`RetryInlineContinuation.ExecuteAsync`, `ScheduledRetryContinuation.ExecuteAsync`, and
`RequeueContinuation.ExecuteAsync` all call `activity?.AddEvent(...)` on the failed attempt and
then move on — none of them touch `Envelope.ParentId` or anything else that would let the next
attempt's `Activity` reference this one.

## Why this isn't just "add ParentId reparenting"

It's tempting to fix this by setting `envelope.ParentId = activity.Id` before retrying, and
that's exactly what we did as a local workaround in our own codebase via a custom
`IChain`/receive middleware. But per the [OpenTelemetry messaging semantic
conventions](https://opentelemetry.io/docs/specs/semconv/messaging/messaging-spans/), a
parent/child relationship is supposed to represent temporal containment — the child's duration
should live inside the parent's. A retry doesn't satisfy that: attempt 1 has already finished
(with an exception) by the time attempt 2 starts, and for a `ScheduledRetryContinuation` there
can be seconds or minutes between them. The documented mechanism for "these two spans are
related, but neither contains the other" is `ActivityLink`, not reparenting.

## Prior art in other .NET messaging libraries

Two other .NET messaging libraries already made this exact call for the same scenario:

- **MassTransit** propagates an `MT-Activity-Propagation: "Link"` header on redelivered/retried
  jobs specifically so the consumer links back to the original activity instead of reparenting
  under it (`JobStateMachine.cs`, `LogContextActivityExtensions.cs`).
- **Brighter**'s `BrighterTracer.GetMessagePumpLinks` explicitly builds `ActivityLink`s for its
  message-pump-to-consume-span relationship, with an in-code comment citing the same OTel
  semantic convention as the reason parent/child isn't used there.

## Proposed fix

Add `Envelope.PreviousAttemptActivityId`, wired through the same serialization and
`EnvelopeMapper` header-propagation path that `ParentId` already uses (so it survives a durable
reschedule via the inbox, and round-trips over external transports the same way `ParentId`
does). The three retry continuations stamp it from the failed attempt's `Activity.Id`
immediately before retrying/rescheduling/requeuing. `WolverineTracing.StartEnvelopeActivity`
consumes it as a single-use `ActivityLink` when starting the next attempt's `Activity`, in
addition to (not instead of) the existing `ParentId`-based parent, which still correctly
reflects the original inbound context.

Happy to open this as a PR if the general direction (new envelope field + `ActivityLink` on
retry, rather than reparenting) looks right. Wanted to raise it as an issue first since it
touches the wire format (`EnvelopeConstants`, `EnvelopeSerializer`, `EnvelopeMapper`) and I'd
rather confirm the approach before asking for review on that.

## Environment

- Wolverine: main branch, current HEAD as of this report.
- Discovered while auditing a custom middleware in our own codebase that manually did
  `envelope.ParentId = Activity.Current?.Id ?? envelope.ParentId` on receive — which "worked"
  for visually connecting attempts in our trace backend, but reparents rather than links, which
  is the wrong shape per the OTel convention above.

outofrange-consulting:
## Problem

Closes #4398.

When a message fails and is retried — inline, on a delay, or requeued — the next attempt's
`Activity` is only ever a sibling of the failed attempt's `Activity`; both are parented off the
same `Envelope.ParentId` from the original inbound context. There's nothing that lets you jump
from "attempt 1 failed" to "attempt 2, which is what happened next" in a trace backend. The only
trace of the retry today is an `ActivityEvent` added to the *failed* attempt's own span
(`wolverine.envelope.retried` / `.rescheduled` / `.requeued`), which is a dead end rather than a
forward pointer.

## Why not just reparent under the failed attempt?

That's the obvious first fix, and it's what we shipped as a local workaround before writing this
PR: set `envelope.ParentId = activity.Id` before retrying. It looked right in the trace UI, but
it's the wrong relationship per the [OpenTelemetry messaging semantic
conventions](https://opentelemetry.io/docs/specs/semconv/messaging/messaging-spans/) —
parent/child is supposed to mean temporal containment, and a retry's span does not live inside
the span of the attempt that failed before it (for `ScheduledRetryContinuation` there can be
minutes between them). The convention's answer for "related but not containing" is
`ActivityLink`. MassTransit (`MT-Activity-Propagation: "Link"` header on redelivery) and Brighter
(`BrighterTracer.GetMessagePumpLinks`) both already do this for the same scenario.

## What this changes

Adds `Envelope.PreviousAttemptActivityId`, wired through the exact same plumbing `ParentId`
already uses:

- `EnvelopeConstants.PreviousAttemptActivityIdKey` — new wire header key, alongside
  `ParentIdKey`.
- `EnvelopeSerializer` — read/write in the binary envelope format (durable storage +
  native-transport round-trip), and added to `ReservedHeaderKeys`.
- `EnvelopeMapper<TIncoming,TOutgoing>` — one `MapPropertyToHeader` line, so every transport
  mapper (Rabbit, Kafka, Azure Service Bus, SQS, ...) picks it up automatically with no
  per-transport changes.
- `WolverineTracing.StartEnvelopeActivity` — if the envelope carries a
  `PreviousAttemptActivityId`, parses it with `ActivityContext.TryParse` and passes it as an
  `ActivityLink` to `ActivitySource.StartActivity`, *in addition to* the existing
  `ParentId`-based parent (which is left untouched — it still correctly represents the original
  inbound context, not the retry chain). The field is cleared immediately after being consumed,
  so it's single-use: a link should point at the attempt immediately before it, not an
  arbitrarily stale one from several retries back.
- `RetryInlineContinuation`, `ScheduledRetryContinuation`, `RequeueContinuation` — each stamps
  `lifecycle.Envelope.PreviousAttemptActivityId = activity.Id` from the failed attempt's
  `Activity`, right next to the existing `activity?.AddEvent(...)` call, before retrying /
  rescheduling / requeuing.

Piggybacking on `ParentId`'s existing infrastructure rather than inventing a new mechanism means
this survives everything `ParentId` already survives: a durable reschedule picked up by a
different process/node via the inbox, and round-tripping over every external transport, with a
one-line change per concern rather than one per transport.

## Open question for maintainers

Right now this is single-use — only the immediately preceding attempt gets linked, not the whole
retry chain. That matches what MassTransit/Brighter do and keeps the wire payload bounded, but if
you'd rather see every prior attempt linked (`List<ActivityLink>` accumulated across retries), I
can rework it that way instead — wanted to raise the tradeoff rather than assume.

## Testing

- New unit tests: `WolverineTracingTests` (link is added when `PreviousAttemptActivityId` is set,
  cleared after being consumed, absent when not set), plus one test per continuation
  (`RetryNowContinuationTester`, `ScheduledRetryContinuationTester`, `RequeueContinuationTester`)
  confirming each stamps `PreviousAttemptActivityId` from the passed-in `Activity`, and that
  `ScheduledRetryContinuation` leaves it null when no `Activity` is available.
- Full `CoreTests` suite: 2909 passed, 2 pre-existing/unrelated skips, 0 failed.

### DIFF
--- a/src/Testing/CoreTests/ErrorHandling/RequeueContinuationTester.cs
+++ b/src/Testing/CoreTests/ErrorHandling/RequeueContinuationTester.cs
@@ -1,3 +1,4 @@
+using System.Diagnostics;
 using CoreTests.Runtime;
 using NSubstitute;
 using Shouldly;
@@ -38,5 +39,21 @@ public class RequeueContinuationTester
         var strategy = new FixedMultiplierJitter(2.0);
 
         ((IJitterable)RequeueContinuation.Instance).TrySetJitter(strategy).ShouldBeFalse();
+    }
+
+    [Fact]
+    public async Task links_the_requeued_envelope_to_the_failed_attempt_activity()
+    {
+        var envelope = ObjectMother.Envelope();
+
+        var context = Substitute.For<IEnvelopeLifecycle>();
+        context.Envelope.Returns(envelope);
+
+        var activity = new Activity("process");
+        activity.Start();
+
+        await RequeueContinuation.Instance.ExecuteAsync(context, new MockWolverineRuntime(), DateTime.Now, activity);
+
+        envelope.PreviousAttemptActivityId.ShouldBe(activity.Id);
     }
 }
--- a/src/Testing/CoreTests/ErrorHandling/RetryNowContinuationTester.cs
+++ b/src/Testing/CoreTests/ErrorHandling/RetryNowContinuationTester.cs
@@ -58,4 +58,23 @@ public class RetryNowContinuationTester
 
         ((IJitterable)RetryInlineContinuation.Instance).TrySetJitter(strategy).ShouldBeFalse();
     }
+
+    [Fact]
+    public async Task links_the_retried_envelope_to_the_failed_attempt_activity()
+    {
+        var continuation = RetryInlineContinuation.Instance;
+
+        var envelope = ObjectMother.Envelope();
+        envelope.Attempts = 1;
+
+        var context = Substitute.For<IEnvelopeLifecycle>();
+        context.Envelope.Returns(envelope);
+
+        var activity = new Activity("process");
+        activity.Start();
+
+        await continuation.ExecuteAsync(context, new MockWolverineRuntime(), DateTimeOffset.Now, activity);
+
+        envelope.PreviousAttemptActivityId.ShouldBe(activity.Id);
+    }
 }
--- a/src/Testing/CoreTests/ErrorHandling/ScheduledRetryContinuationTester.cs
+++ b/src/Testing/CoreTests/ErrorHandling/ScheduledRetryContinuationTester.cs
@@ -51,4 +51,42 @@ public class ScheduledRetryContinuationTester
 
         await lifecycle.Received(1).ReScheduleAsync(now.AddSeconds(10));
     }
+
+    [Fact]
+    public async Task links_the_rescheduled_envelope_to_the_failed_attempt_activity()
+    {
+        var continuation = new ScheduledRetryContinuation(TimeSpan.FromSeconds(10));
+
+        var envelope = ObjectMother.Envelope();
+        envelope.Attempts = 1;
+
+        var lifecycle = Substitute.For<IEnvelopeLifecycle>();
+        lifecycle.Envelope.Returns(envelope);
+
+        var now = new DateTimeOffset(2026, 4, 13, 12, 0, 0, TimeSpan.Zero);
+        var activity = new Activity("process");
+        activity.Start();
+
+        await continuation.ExecuteAsync(lifecycle, new MockWolverineRuntime(), now, activity);
+
+        envelope.PreviousAttemptActivityId.ShouldBe(activity.Id);
+    }
+
+    [Fact]
+    public async Task does_not_set_previous_attempt_activity_id_when_activity_is_null()
+    {
+        var continuation = new ScheduledRetryContinuation(TimeSpan.FromSeconds(10));
+
+        var envelope = ObjectMother.Envelope();
+        envelope.Attempts = 1;
+
+        var lifecycle = Substitute.For<IEnvelopeLifecycle>();
+        lifecycle.Envelope.Returns(envelope);
+
+        var now = new DateTimeOffset(2026, 4, 13, 12, 0, 0, TimeSpan.Zero);
+
+        await continuation.ExecuteAsync(lifecycle, new MockWolverineRuntime(), now, null);
+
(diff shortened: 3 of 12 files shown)


jeremydmiller:
Thank you @outofrange-consulting, this was a well-argued PR. Using an \`ActivityLink\` instead of reparenting is the right call, and the design carries straight into the replacement. I've superseded it with #4405, and you're credited as co-author on the commit.

Two things changed in review:

1. **Where the link lands.** \`StartEnvelopeActivity\` is shared by the send, receive and execute spans. Every receiver (Buffered, Durable, NativeAck, Inline) starts its \`receive\` span before the pipeline starts the execute span, so the receive span took the link and cleared it, and the process span got nothing. #4405 attaches the link in \`HandlerPipeline.InvokeAsync\`, the one place every attempt goes through.
2. **How it's carried.** A mapped \`Envelope\` property adds one more reserved-header read to every incoming message in every transport mapper. A loose \`Envelope.Headers\` entry already round-trips through every mapper and the serializer, and a message that is never retried pays nothing for it.

On your open question, #4405 keeps your single-use, previous-attempt-only choice. One more thing turned up: a durable-inbox reschedule doesn't rewrite the stored body, so the link can't survive that path. That limitation is written up in #4405.

#4405 also adds end-to-end coverage through the OpenTelemetry SDK's in-memory exporter, including a RabbitMQ round trip on both Inline and BufferedInMemory listeners. Thanks again for raising this and doing the research.

PR #4408, opened 2026-09-10
GH-4407: owner-scoped assignment deletes, claim-if-absent, and reconciliation-sweep hardening

### LINKED ISSUE #4407: Follow-ups from #4404: owner-scoped assignment deletes on RavenDB/Cosmos, reconciliation-sweep hardening, compliance coverage for assignment divergence
Follow-up to #4404 (merged as d37bbf3), which added the node-side reconciliation sweep for #3987. The sweep checks, on each node, whether the agents the assignment table says it owns match the agents actually running there. Review turned up one real defect on two storage providers and a few hardening items. Also related: #4297.

## 1. RavenDB and Cosmos DB delete other nodes' assignment rows

The SQL providers remove an assignment only if it belongs to the calling node (`where id = @id and node_id = @node`): Postgres, SQL Server, MySQL, Oracle and SQLite. RavenDB and Cosmos DB delete by agent id alone:

- `RavenDbMessageStore.NodeAgents.cs`: `session.Delete(AgentAssignment.ToId(agentUri))`
- `CosmosDbMessageStore.NodeAgents.cs`: an unconditional `DeleteItemAsync(CosmosAgentAssignment.ToId(agentUri), ...)`

The bug predates #4404, but #4404 makes it trigger every time. The sweep's "stop the orphan" branch calls `StopAgentAsync` exactly when **another live node owns the row**, and `StopAgentAsync` ends in `RemoveAssignmentAsync`. On these two stores:

1. Node A stops its orphan copy and deletes node B's row, while B is still running the agent.
2. The leader builds its view from persisted rows, so it now sees the agent as unassigned and places it on node C. The agent now runs on both B and C.
3. B's sweep sees C owns the row, stops its own copy, and deletes C's row. The next round begins.

This repeats until the leader happens to pick a node that's already running the agent. Each round creates a new duplicate.

- [ ] **The sweep's orphan stop shouldn't touch assignment rows at all**, since the row belongs to someone else. For example, a `StopAgentAsync` overload or a private stop that skips `RemoveAssignmentAsync`. This fixes it for every provider. Add a unit test asserting the orphan stop never calls `RemoveAssignmentAsync`.
- [ ] **Make `RemoveAssignmentAsync` owner-scoped on RavenDB and Cosmos DB**, matching the SQL providers. RavenDB: load the document and delete only if `NodeId` matches. Cosmos: read, then do a conditional delete on the document's ETag.

## 2. Restoring a claim can overwrite a newer one

When a running copy has no row anywhere, the sweep restores this node's claim through `AddAssignmentAsync`. That's an upsert, so the last writer wins. The sweep acts on a snapshot read at the start of the health check, so a claim a peer wrote in the meantime gets overwritten. On the SQL providers this converges through the peer's own sweep, but it's avoidable churn, and on RavenDB/Cosmos it feeds item 1.

- [ ] Add an "insert only if absent" assignment operation to `INodeAgentPersistence`, and use it for restoring a claim.

## 3. No limit on sweep actions per health check

The sweep runs its starts and stops one after another inside the health check. On the leader node that happens before the leader's assignment pass. On a fleet like the one in #3987 (about 13,500 agents), a leader crash can leave hundreds of divergences on one node, and each start can take seconds with retries. That would delay the leader's assignment pass by that much. Heartbeats run on their own timer, so nothing gets ejected, but assignment stalls.

- [ ] Cap the actions per health check, leaving the rest for the next tick, or run the sweep after the leader's assignment pass.

## 4. Sweep starts skip the revocation guard

The sweep calls `StartAgentAsync` directly rather than the revocation-aware `StartAgentGuardedAsync`, and nothing tracks starts that are still in progress. The wedged-agent restart path in `StartAgentAsync` deregisters the agent before a potentially slow retrying start, while the row still names this node. The narrow risks are a second concurrent start, or restarting an agent that a stop has revoked since the snapshot.

- [ ] Capture the command sequence when the snapshot is taken and start through `StartAgentGuardedAsync`. Also track starts in progress, the way `_stoppingAgents` tracks stops, and skip them.

## 5. Compliance coverage for assignment divergence

**`NodePersistenceCompliance`**, the storage contract run for every provider:
- [ ] `removing_an_assignment_owned_by_another_node_is_a_no_op`. This fails today on RavenDB and Cosmos DB, so it's the regression test for item 1.
- [ ] `add_assignment_moves_the_row_to_the_new_node`, documenting today's last-writer-wins behaviour.
- [ ] If item 2 lands, a test that the new "only if absent" operation never takes over a peer's row.

**`LeadershipElectionCompliance`**, which runs real hosts against real storage for every provider in `src/Persistence/LeaderElection/*` plus the RabbitMQ, Azure Service Bus and SharedMemory control-queue variants. Each scenario fakes one divergence the sweep is meant to heal, then asserts with the existing `expectExactlyOneCopyOfEachAsync`, which already catches missing agents and duplicates:
- [ ] **Orphan duplicate after a leadership handover.** Let assignments settle, then start a second copy of an agent on a node that doesn't own it, while the owner's row stays put. Expect exactly one copy and the owner's row still there. This is the RavenDB/Cosmos loop from item 1, exercised end to end.
- [ ] **Assigned but not running.** Stop an agent locally on its owner without removing its row. Expect it running again within about the threshold × `HealthCheckPollingTime`, which is roughly 3 seconds in this suite.
- [ ] **Row lost for a running agent** (the peer-ejection case). Delete a running agent's assignment row directly through storage. Expect the claim restored and exactly one copy.
- [ ] **Leader lost with starts in flight.** Add a node, then shut down the leader before its start commands land. Expect exactly one copy of every agent once things settle.

Faking these may need a small test hook, for example a local stop that leaves rows alone, or direct access to the node store from the compliance assembly.

jeremydmiller:
Closes #4407. These are the follow-ups from reviewing #4404, the node-side reconciliation sweep for #3987.

## 1. RavenDB and Cosmos DB no longer delete other nodes' assignment rows

On the SQL providers, `RemoveAssignmentAsync` only deletes a row the calling node owns (`where id = @id and node_id = @node`). **RavenDB and Cosmos DB deleted by agent id alone.** #4404's orphan-stop branch calls `StopAgentAsync` exactly when another node owns the row, so on those stores it deleted the owner's row. The leader then saw the agent as unassigned and placed it again, creating a new duplicate that the next sweep would stop, deleting that row too.

- **The orphan stop leaves assignment rows alone.** `StopAgentAsync(uri)` keeps its behaviour; the sweep calls a private `stopAgentAsync(uri, removeAssignment: false)`. This fixes the problem for every provider.
- **RavenDB and Cosmos DB deletes are now owner-scoped.** RavenDB loads the row and deletes it only if the `NodeId` matches, with optimistic concurrency. Cosmos reads the row and deletes it only if the node matches, conditional on the ETag. A row that changes hands in between survives.

## 2. Restoring a claim no longer overwrites a newer one

Adds `INodeAgentPersistence.TryClaimAssignmentAsync(nodeId, agentUri)`. It inserts a row only if no node owns the agent, and returns whether the row belongs to the caller afterwards. The sweep uses it instead of the last-writer-wins `AddAssignmentAsync` when it restores a claim for a running copy, so a peer that claimed the agent after the snapshot keeps it.

- **SQL providers:** a race-safe insert-if-absent for each dialect: Postgres `on conflict do nothing`, SQL Server `not exists` under `updlock, holdlock`, MySQL `on duplicate key` as a no-op, Oracle `merge ... when not matched` (tolerating ORA-00001), SQLite `insert or ignore`. Each is followed by a `count(*)` on `id` and `node_id` to check ownership.
- **RavenDB:** a create-only store (empty change vector), re-reading to see who won if it loses the race.
- **Cosmos DB:** `CreateItemAsync`, reading on conflict.
- **Wrappers:** `MultiTenantedMessageStore` forwards the new member explicitly, and `NullMessageStore` returns true.
- **Deliberately no default implementation.** The multi-tenant store implements the interface explicitly and forwards every member. With a default, a wrapper that forwarded everything but this one would silently fall back to an upsert. Anyone implementing `INodeAgentPersistence` outside this repo will need to add the method.

## 3. A cap on reconciliation actions per health check

New `DurabilitySettings.MaxLocalAgentReconciliationsPerTick`, default 50; 0 or less means no cap. The sweep runs inside the serialized health check, before the leader's assignment pass on the leader node, so after something like a leader crash on a large fleet an unbounded sweep could hold that pass back. Deferred divergences keep their streak count and are handled on later ticks, oldest first.

## 4. Sweep starts respect stop revocations and starts still in progress

- **Revocation:** the command sequence is captured *before* the health check reads its snapshot. The sweep starts agents through `StartAgentGuardedAsync(uri, snapshotSequence)`, and skips restoring a claim if a stop has arrived since. A stop the snapshot couldn't have known about always wins.
- **Starts in progress:** a `_startingAgents` set tracks them, the way `_stoppingAgents` tracks stops. The marker is set before the wedged-restart path deregisters the agent and released only after registration, so the sweep never sees a slow start as "assigned here but not running" and starts it again.

## 5. Compliance coverage for assignment ownership and divergence

**`NodePersistenceCompliance`**, run for every provider:
- `removing_an_assignment_owned_by_another_node_is_a_no_op`, the regression test for item 1
- `add_assignment_moves_the_row_to_the_new_node`, documenting last-writer-wins
- `claiming_an_unowned_assignment_takes_it`
- `claiming_an_assignment_a_peer_owns_leaves_it_alone`
- `claiming_an_assignment_this_node_already_owns_succeeds`

**`LeadershipElectionCompliance`**, run against real hosts and real storage on every leadership suite. Each scenario fakes a divergence, then requires exactly one copy of every agent, and for the agent under test a row naming the node that runs it:
- a duplicate copy on a node the table doesn't name
- an agent assigned to a node but not running there
- a running agent whose row was deleted
- the leader dying while its start commands are still in flight

The scenarios deliberately don't assert *which* node ends up with the agent. The leader may legitimately re-place or rebalance it while the divergence heals, and in practice it does.

Also fixed: `persist_and_load_node_records` never advanced its retry counter, so on the SharedMemory variant it spun forever. That was a 40-minute hang locally; SlowTests only runs on manual dispatch, so CI never saw it. It now makes 10 attempts with a delay between them, and on SharedMemory it now fails fast rather than hanging (details under Verification).

## Verification

- `dotnet build wolverine.slnx -c Release -f net9.0`: 0 errors, 0 warnings
- CoreTests `Runtime.Agents`: 386 of 386. New unit tests cover: the orphan stop never calls `RemoveAssignmentAsync`, restoring goes through the claim and never the upsert, a peer winning the claim race keeps the row, a start in progress is never started again, and the per-tick cap works through a backlog over later ticks.
- `NodePersistenceCompliance`, all passing:
  - Postgres 18/18
  - SQL Server 18/18
  - MySQL 17/17
  - Oracle 17/17
  - SQLite 17/17
  - RavenDB 19/19
- **RavenDB negative control:** with only the old delete-by-id `RemoveAssignmentAsync` swapped back in, `removing_an_assignment_owned_by_another_node_is_a_no_op` fails (1 of 1), and passes again once the fix is restored.
- **Leadership suites:** all four new scenarios pass on MySQL, Oracle, Postgres, RavenDB (both variants), SQL Server, RabbitMQ and SharedMemory. The only failures were `singular_agent_is_only_running_on_one` (below) and SharedMemory's `persist_and_load_node_records`, which predates this PR and now fails fast instead of hanging.
- **`singular_agent_is_only_running_on_one`:** a known flake in the leadership suites (see #3451 for the CosmosDB occurrence). It failed with two copies in three suite runs on this branch, and in 1 of 5 isolated Oracle runs against 0 of 5 on the #4404 baseline. **In every failing run the reconciliation sweep took no action at all**, and every behaviour this PR changes on the SQL providers goes through a sweep action, so it isn't involved. Each failing log instead shows the leader re-placing a start that hadn't been confirmed yet (`Node 1 confirmed 0 of 5 requested agents; ... remain unconfirmed`), then `simple:///` starting on two nodes. A larger interleaved A/B is running; its numbers will follow as a comment.
- **Not run locally:** Cosmos DB (needs the emulator) and Azure Service Bus. CI covers both.

https://claude.ai/code/session_01VDUrBeB4tTnKj4AExCS1nj

### DIFF
--- a/src/Persistence/MySql/Wolverine.MySql/MySqlNodePersistence.cs
+++ b/src/Persistence/MySql/Wolverine.MySql/MySqlNodePersistence.cs
@@ -300,6 +300,25 @@ internal class MySqlNodePersistence : DatabaseConstants, INodeAgentPersistence
             .ExecuteNonQueryAsync(cancellationToken);
     }
 
+    public async Task<bool> TryClaimAssignmentAsync(Guid nodeId, Uri agentUri, CancellationToken cancellationToken)
+    {
+        // GH-4407: insert only if nobody owns the agent yet (the duplicate-key branch is a no-op), then report
+        // whether the row is ours
+        await _dataSource.CreateCommand(
+                $"INSERT INTO {_assignmentTable} (id, node_id) VALUES (@id, @node) ON DUPLICATE KEY UPDATE id = id")
+            .With("id", agentUri.ToString())
+            .With("node", nodeId)
+            .ExecuteNonQueryAsync(cancellationToken);
+
+        var owned = await _dataSource.CreateCommand(
+                $"SELECT COUNT(*) FROM {_assignmentTable} WHERE id = @id AND node_id = @node")
+            .With("id", agentUri.ToString())
+            .With("node", nodeId)
+            .ExecuteScalarAsync(cancellationToken);
+
+        return Convert.ToInt64(owned) > 0;
+    }
+
     public async Task OverwriteHealthCheckTimeAsync(Guid nodeId, DateTimeOffset lastHeartbeatTime)
     {
         await _dataSource.CreateCommand($"UPDATE {_nodeTable} SET health_check = @now WHERE id = @id")
--- a/src/Persistence/Oracle/Wolverine.Oracle/OracleNodePersistence.cs
+++ b/src/Persistence/Oracle/Wolverine.Oracle/OracleNodePersistence.cs
@@ -285,6 +285,39 @@ internal class OracleNodePersistence : DatabaseConstants, INodeAgentPersistence
         await conn.CloseAsync();
     }
 
+    public async Task<bool> TryClaimAssignmentAsync(Guid nodeId, Uri agentUri, CancellationToken cancellationToken)
+    {
+        await using var conn = await _dataSource.OpenConnectionAsync(cancellationToken);
+
+        // GH-4407: insert only if nobody owns the agent yet, then report whether the row is ours
+        await using (var insert = conn.CreateCommand(
+                         $"MERGE INTO {_assignmentTable} t USING DUAL ON (t.id = :id) " +
+                         "WHEN NOT MATCHED THEN INSERT (id, node_id) VALUES (:id, :node)"))
+        {
+            insert.With("id", agentUri.ToString());
+            insert.With("node", nodeId);
+
+            try
+            {
+                await insert.ExecuteNonQueryAsync(cancellationToken);
+            }
+            catch (global::Oracle.ManagedDataAccess.Client.OracleException e) when (e.Number == 1)
+            {
+                // ORA-00001: a peer inserted between the MERGE's match and its insert, so it owns the row
+            }
+        }
+
+        await using var select = conn.CreateCommand(
+            $"SELECT COUNT(*) FROM {_assignmentTable} WHERE id = :id AND node_id = :node");
+        select.With("id", agentUri.ToString());
+        select.With("node", nodeId);
+        var owned = await select.ExecuteScalarAsync(cancellationToken);
+
+        await conn.CloseAsync();
+
+        return Convert.ToInt64(owned) > 0;
+    }
+
     public async Task OverwriteHealthCheckTimeAsync(Guid nodeId, DateTimeOffset lastHeartbeatTime)
     {
         await using var conn = await _dataSource.OpenConnectionAsync();
--- a/src/Persistence/Wolverine.CosmosDb/Internals/CosmosDbMessageStore.NodeAgents.cs
+++ b/src/Persistence/Wolverine.CosmosDb/Internals/CosmosDbMessageStore.NodeAgents.cs
@@ -260,14 +260,54 @@ public partial class CosmosDbMessageStore : INodeAgentPersistence
 
     public async Task RemoveAssignmentAsync(Guid nodeId, Uri agentUri, CancellationToken cancellationToken)
     {
+        // GH-4407: only this node's own claim. Deleting by id alone let a node stopping a copy it did not own
+        // delete the owner's row, which the leader then read as an unassigned agent and placed again. The
+        // delete is conditioned on the ETag of the read, so a row that changes hands in between survives.
+        var id = CosmosAgentAssignment.ToId(agentUri);
+        var partition = new PartitionKey(DocumentTypes.SystemPartition);
+
         try
         {
-            await _container.DeleteItemAsync<dynamic>(
-                CosmosAgentAssignment.ToId(agentUri), new PartitionKey(DocumentTypes.SystemPartition),
+            var existing = await _container.ReadItemAsync<CosmosAgentAssignment>(id, partition,
                 cancellationToken: cancellationToken);
+            if (existing.Resource.NodeId != nodeId.ToString())
+            {
+                return;
+            }
+
+            await _container.DeleteItemAsync<CosmosAgentAssignment>(id, partition,
+                new ItemRequestOptions { IfMatchEtag = existing.ETag }, cancellationToken);
         }
-        catch (CosmosException e) when (e.StatusCode == HttpStatusCode.NotFound)
+        catch (CosmosException e) when (e.StatusCode is HttpStatusCode.NotFound or HttpStatusCode.PreconditionFailed)
         {
+            // Already gone, or it changed hands between the read and the delete -- not ours to remove
+        }
+    }
+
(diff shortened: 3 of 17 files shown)


jeremydmiller:
**Singular-agent A/B result:** in 15 interleaved rounds on Oracle, `singular_agent_is_only_running_on_one` passed **15 of 15 on the #4404 baseline and 15 of 15 on this branch**. That means no measurable difference. Together with the logs, where the reconciliation sweep took no action in any failing run and the duplicate comes from the leader re-placing a start that hadn't been confirmed yet, it confirms that flake predates this PR and isn't affected by it.

Issue #4421, opened 2026-09-13
Codegen output is deterministic as of JasperFx 2.69.2 — regenerate the committed output, and a drift check is now worth having


jeremydmiller:
Downstream of **JasperFx/jasperfx#832**, fixed in **JasperFx 2.69.2**. This is the root cause behind wolverine#3059, which was closed without one.

## What changed upstream

`MethodFrameArranger.compileFrames` step 3 enumerated `DependencyGatherer.Dependencies`, an `ImHashMap` keyed by `Frame`. `Frame` does not override `GetHashCode`, so that walk followed **identity hash codes** — arbitrary per object, per process — and two arrangements of the same logical method emitted the same statements in different orders. jasperfx#196 had fixed this for `DependencyGatherer.Variables`, which is why constructor parameters and fields have been stable and the method bodies were not.

It was never really a macOS-vs-Linux thing (wolverine#3059's framing). Identity hash codes differ between any two processes, so one machine reproduces it fine.

Measured by running `codegen write` twice in the same environment on an app with committed generated code: **15 of 60 files differed at JasperFx 1.24.1, 6 of 60 still differed at 1.31.0**. What was left moving is statement order. That is now fixed.

## What Wolverine should do

**1. Bump to JasperFx 2.69.2.**

**2. Regenerate the committed generated code — expect a one-time diff.** Wolverine tracks **61 generated `.cs` files** under `src/**/Internal/Generated` (`CodeGenTarget`, `DeepMiddlewareUsage`, `StartupStyleTarget`, and others). Statements inside generated methods may reorder once as they settle onto the deterministic order. That is this fix landing, not a semantic change — and it is the last such diff, which is the point.

Worth doing as its own commit with nothing else in it, so the reordering is legible in review rather than buried.

**3. A codegen drift check is now meaningful.** This is the part worth more than the tidy diff. Until now, anything that byte-compared generated output — a CI drift check, a reproducible build, a reviewer reading a regenerated file — could not distinguish hash-order noise from a real change. Wolverine has ~42 files referencing `codegen write` / `AssertConfigurationIsValid`; if any drift check was weakened, made advisory, or never added because it was too noisy to trust, it can now be a real gate.

If one already exists and has been quietly tolerating diffs, it should start being believed.

## Acceptance

- [ ] `JasperFx` pinned to 2.69.2.
- [ ] Committed generated code regenerated, in a commit of its own.
- [ ] Regenerating twice in a row produces **byte-identical** output.
[...]

Issue #4420, opened 2026-09-13
TypeLoadMode.Static: generated HandlerRegistry.MessageTypes() is empty when message type lives in a different assembly than the handler


derekchoate:
## Summary

When the handler class and the message class it handles live in **different assemblies**, `codegen write` produces a `GeneratedHandlerRegistry` whose `HandlerTypes()` correctly lists the handler, but `MessageTypes()` comes back **empty**. Under `TypeLoadMode.Static`, this means the message can never be dispatched at runtime — every publish of that message type fails with `No known handler for <MessageType>`, even though the handler class and its `Handle` method are right there in the assembly.

I think this is the same underlying "generated type loader doesn't cross assembly boundaries correctly" class of bug as #2640, hit via a different path: not `IncludeAssembly()` / the `Wolverine.SourceGeneration` analyzer, but the `codegen write` CLI command plus `TypeLoadMode.Static`, where message manifest capture (added in #2928 / PR for #2906, released in 6.36.0) doesn't seem to look outside the assembly being scanned.

## Environment

- WolverineFx 6.36.0, WolverineFx.RabbitMQ 6.36.0
- .NET 10 (net10.0)
- macOS (arm64), also reproduced under `dotnet publish -c Release -p:PublishAot=true` (Native AOT)

## Repro

Two projects:

- **`Contracts`** (class library) — defines the message:
  ```csharp
  namespace Contracts;
  public record DoThing(Guid Id);
  ```
- **`Host`** (worker service, references `Contracts`) — defines the handler and configures Wolverine:
  ```csharp
  namespace Host.Handlers;
  public class DoThingHandler
  {
      public Task Handle(Contracts.DoThing message) => Task.CompletedTask;
  }
  ```
  ```csharp
  builder.UseWolverine(opts =>
  {
      opts.CodeGeneration.TypeLoadMode = TypeLoadMode.Static;
      // ... transport config ...
  });
  ```

Steps:
1. `dotnet run --project Host -- codegen write`
2. Inspect `Host/Internal/Generated/WolverineHandlers/GeneratedHandlerRegistry.cs`
3. Publish a `DoThing` message and run `Host`

## Expected

`GeneratedHandlerRegistry.MessageTypes()` includes `typeof(Contracts.DoThing)`, and the message dispatches to `DoThingHandler.Handle` normally.

## Actual

Generated registry looks like this (real output, only names changed):

```csharp
[...]

Issue #4287, opened 2026-09-04
[Native AOT on Linux] Several startup crashes in JasperFx & Wolverine (`PlatformNotSupportedException` in `GetReferencedAssemblies`, unhandled stack walk, and `MissingMethodException` in generic reflection)


RaySinner:
## First of all, thanks for the most powerful library for .The NET that exists!!

Essentially bugs

### Description

When publishing a .NET 10 / .NET 9 service targeting **Linux Native AOT** (`<PublishAot>true</PublishAot>`, `linux-x64`, single-file ELF executable), Wolverine 6.32.0 and JasperFx 2.59.0 currently fail to start due to several assumptions around reflection and call stack inspection that do not hold under Native AOT on Linux.

Below is an in-depth breakdown of the root causes, call stacks, and concrete fix recommendations.

---

### Environment
- **Framework:** .NET 10.0 (also applies to .NET 8 / 9 Native AOT)
- **OS:** Linux x64 (Docker container)
- **Settings:** `<PublishAot>true</PublishAot>`, `<StripSymbols>true</StripSymbols>`, `<TypeLoadMode>Static</TypeLoadMode>`
- **Packages:** `WolverineFx 6.32.0`, `JasperFx 2.59.0`

---

#### Call Stack
```text
[Fatal] Startup error: System.PlatformNotSupportedException: Operation is not supported on this platform.
   at System.Reflection.Runtime.Assemblies.RuntimeAssemblyInfo.GetReferencedAssemblies() + 0x20
   at JasperFx.JasperFxOptions.HasReferenceToJasperFxTool(Assembly assembly) + 0xb
   at JasperFx.JasperFxOptions.DetermineCallingAssembly() + 0x26
   at JasperFx.JasperFxServiceCollectionExtensions.AddJasperFx(IServiceCollection services, Action`1 configure) + 0x2a
   at Wolverine.HostBuilderExtensions.AddWolverine(IServiceCollection services, WolverineOptions options, ...)
```

#### Cause
In `JasperFxServiceCollectionExtensions.AddJasperFx`:
```csharp
public static IServiceCollection AddJasperFx(this IServiceCollection services, Action<JasperFxOptions>? configure = null)
{
    // BUG: Called unconditionally BEFORE checking if RememberedApplicationAssembly is already set!
    var registrationAssembly = JasperFxOptions.DetermineCallingAssembly();
    if (JasperFxOptions.RememberedApplicationAssembly == null)
    {
[...]

PR #4430, opened 2026-09-13
GH-4426: emit the Native AOT [DynamicDependency] roots from codegen write

### LINKED ISSUE #4426: Native AOT: emit [DynamicDependency] rooting from codegen write (Wolverine-side follow-up to #4287 / jasperfx#743)
#4287 fixed most of Wolverine's Native AOT startup story (shipped in #4298 / JasperFx 2.63.1): the `Assembly.ExportedTypes`-based registry lookup, and reflective router construction for framework-internal message types (`Envelope`, `Acknowledgement`, `FailureAcknowledgement`, `IAgentCommand`).

It explicitly did not cover arbitrary user-defined message types. Per the closing comment on #4287, a real Native AOT app today still needs hand-written `[DynamicDependency(All, ...)]` roots for the generated registry, every generated handler type, every handler class, every message type, and the closed `MessageRouter<T>`/`EmptyMessageRouter<T>` per message type — see `src/Testing/Wolverine.AotSmoke.Publish/Program.cs`'s `AotRoots.Pin()` for the manually-written reference shape. Without these, `PublishAsync`/routing warm-up (`WolverineRuntime.PrepopulateRoutingCache`) throws `MissingMethodException` on startup for any message type that isn't one of the framework ones #4287 special-cased.

jasperfx#743 scoped and shipped the codegen primitives needed to automate this away (`GeneratedAttribute`/`AttributeArg`, `GeneratedType.AddStaticVoidMethod`, and an `AotRootsCompanionExtensions.AddAotRoots(...)` helper that emits exactly the `[ModuleInitializer]` + `[DynamicDependency]` companion) — released in JasperFx 2.65.0, already pinned here at 2.66.1. That issue's design doc calls out a "Wolverine follow-up" to actually consume these primitives (its W1–W4 work items) but that follow-up issue was never filed. This is it.

**Proposed implementation (W1):**

1. In `HandlerGraph.GeneratesCode.cs` (`explodeAllFiles`), while walking `Chains`, also collect each chain's generated type name (`chain.TypeName`, prefixed with `Internal.Generated.WolverineHandlers.`) and its `MessageType`.
2. Pass these into `HandlerRegistryCodeFile`; in `AssembleTypes`, call `assembly.AddAotRoots("AotRoots", roots)` — guarded by the same `DynamicCodeBuilder.WithinCodegenCommand` check already used for `MessageTypes()` — rooting: the registry type, every generated handler type (as sibling names via `AttributeArg.TypeNamed`, since they don't exist as `Type` yet during `codegen write`), every handler class, every message type, and `MessageRouter<T>`/`EmptyMessageRouter<T>` closed over each message type (neither has generic constraints, so this closes for any message type).
3. `AddAotRoots` already no-ops correctly for `--language fsharp`, so no extra branching needed.

**Verification:** extend `Wolverine.AotSmoke.Publish` (or add a sibling project) with a two-assembly repro (message type and handler in separate assemblies — the shape from #4420), do a real `dotnet publish -p:PublishAot=true`, run the native binary, and assert the handler fires. That closes the coverage gap noted in #4287's fix (the existing AOT smoke lanes are trim-analysis-only or CoreCLR runs, not real `PublishAot`), and would let `AotRoots.Pin()` in the existing smoke project be deleted as proof the rooting is now automatic.

Originally surfaced via #4420, which mis-diagnosed the symptom (an empty `MessageTypes()` manifest) — the actual mechanism blocking that reporter's Native AOT publish is this one.

jeremydmiller:
Makes `codegen write` emit the Native AOT rooting that every Native AOT application has had to hand-write, and proves it by deleting the hand-written copy.

Closes #4426.

## The problem

#4287 fixed Wolverine's Native AOT startup for the framework's own message types and explicitly left user-defined ones alone. Until now a real Native AOT app needed a hand-written `[DynamicDependency(All, ...)]` block covering the generated registry, every generated handler type, every handler class, every message type, and `MessageRouter<T>` / `EmptyMessageRouter<T>` closed over each one — because all of it is reached reflectively (assembly scan + `Activator.CreateInstance`), so ILC trims it and `TypeLoadMode.Static` silently degrades to a scan that finds nothing. Miss one message type and `PublishAsync` or the routing warm-up throws `MissingMethodException` at startup.

jasperfx#743 shipped the codegen primitives to automate this (JasperFx 2.65.0) and called for a Wolverine follow-up that was never filed. This is that W1 work item.

## What changed

`HandlerGraph.explodeAllFiles` now also collects the **names** of the handler types it is about to emit plus their message types, and `HandlerRegistryCodeFile` feeds them to `GeneratedAssembly.AddAotRoots`. That emits an `AotRoots` companion whose `Pin()` carries the rooting block under a `[ModuleInitializer]` — an unconditional ILC root, so **nothing has to call it**. There is no app-side code at all any more.

Details worth knowing:

- **Sibling generated types are rooted by name** (`AttributeArg.TypeNamed`), because they have no runtime `Type` while codegen is running. They are qualified with `GeneratedAssembly.Namespace`, which is already `{GeneratedNamespace}.{ChildNamespace}`.
- **Guarded by `DynamicCodeBuilder.WithinCodegenCommand`**, the same guard the conventional message-type scan uses. During Static *attach* the committed file already carries the companion, and adding another generated type then would only confuse the loader.
- **`AddAotRoots` no-ops for `--language fsharp`** (F# does not honour `ModuleInitializerAttribute`), so there is no language branching here.
- **Two analyzer suppressions, on their own non-iterator helper.** `MakeGenericType` trips IL3050 *and* IL2055 in core, where analysis is strictest and warnings are errors. An attribute on `buildAotRoots` would not reach its generated `MoveNext`, hence the helper. Both are honest: the call only runs during `codegen write` on CoreCLR, and neither router declares a generic constraint, so there are no requirements to guarantee.
- **The emitted block is ordered**, because this output is byte-compared by the drift gate from #4421.

## Verification — including the real-`PublishAot` gap #4287 left open

The hand-written `AotRoots` class **and its `Pin()` call are deleted** from `Wolverine.AotSmoke.Publish`. That project now proves the emitted rooting by having none of its own, which is the strongest assertion available here.

| | Result |
|---|---|
| Real `dotnet publish -p:PublishAot=true` + **run the native binary** | **exit 0** — `Mach-O 64-bit arm64`, no managed assemblies beside it. Logged *"Using pre-generated Wolverine HandlerRegistry (1 handler types); skipping assembly scan"*, dispatched a message, `OK: Native AOT boot + dispatch smoke passed`. The registry survived the trimmer with nothing hand-written. |
| `codegen write` twice in a row | byte-identical |
| CoreTests | **2944 passed, 0 failed, 2 skipped** — unchanged from baseline |
| Full solution build (`wolverine.slnx -c Release -f net9.0`) | **0 errors, 0 warnings**, 214/214 projects |
| `CICodegenDrift` (the #4421 gate) on this branch | **green** — "the committed code generation output for all 6 projects matches" |

## About the second commit

#4426 changes what `codegen write` emits for **every** project, so the four projects #4427 just regenerated needed it again to pick up the companion — otherwise the new drift gate is red on this branch, which is the gate doing exactly its job. Roots emitted per project: ConsoleApp 27, StaticCodeGenDemonstrator 6, CodeGenTarget 1, DeepMiddlewareUsage 1. The two HTTP-only projects get a single root because they declare no message handlers, so the companion anchors just the registry — still the type Static mode must locate reflectively.

## Known gaps, deliberately not widened here

- **`GeneratedHttpEndpointRegistry` is not rooted.** This covers the handler side (the issue's W1). The HTTP endpoint registry is a separate `ICodeFile` in `Wolverine.Http` and wants the same treatment — worth its own follow-up rather than being smuggled in here.
- **`StartupStyleTarget` stays excluded**, for the reason recorded in #4427: regenerating it deletes its endpoint files and emits an empty registry, which is a separate bug.

https://claude.ai/code/session_01VDUrBeB4tTnKj4AExCS1nj

## Credit

**Co-authored by @derekchoate**, whose #4426 is not just a bug report — it is the design this PR implements. That issue laid out the W1 plan in full: collect each chain's generated type name and `MessageType` while walking `Chains` in `HandlerGraph.GeneratesCode.cs`, pass them into `HandlerRegistryCodeFile`, call `AddAotRoots` from `AssembleTypes` behind the existing `DynamicCodeBuilder.WithinCodegenCommand` guard, name sibling generated types via `AttributeArg.TypeNamed` because they have no runtime `Type` during `codegen write`, root the registry plus every generated handler type, handler class, message type and `MessageRouter<T>`/`EmptyMessageRouter<T>` closed over each, and rely on `AddAotRoots` already no-opping for `--language fsharp`. It also specified the verification: a real `dotnet publish -p:PublishAot=true`, run the native binary, and delete `AotRoots.Pin()` as the proof that rooting is now automatic.

The implementation here follows that plan closely, and the two places it diverged were both forced by the compiler rather than by design — the IL3050 **and** IL2055 suppressions needing their own non-iterator helper, and deferring the two-assembly repro.

@derekchoate also filed #4420, which surfaced this mechanism in the first place (reported as an empty `MessageTypes()` manifest, then correctly re-diagnosed in #4426).

One honest note: this credit was added after the squash merge. The merged commit `77349ebc4` carries no `Co-Authored-By` trailer for him, and changing that would mean rewriting `main`, which is not worth doing to a published branch — so the credit lives here instead, and the next commit in this area should carry the trailer.


jeremydmiller:
Author credit where it belongs: **@derekchoate** should be recorded as a co-author of this change.

#4426 was effectively the implementation spec — it named the collection point in `HandlerGraph.explodeAllFiles`, the hand-off into `HandlerRegistryCodeFile`, the `AddAotRoots` call behind the `WithinCodegenCommand` guard, `AttributeArg.TypeNamed` for sibling generated types, the exact set of things to root (registry, generated handlers, handler classes, message types, and `MessageRouter<T>`/`EmptyMessageRouter<T>` closed over each), the F# no-op, *and* the verification strategy of doing a real `PublishAot` run and deleting `AotRoots.Pin()` as proof. He also filed #4420, which surfaced the mechanism to begin with.

I've added a Credit section to the PR description. Being straight about the limitation: this landed after the squash merge, so the merged commit (`77349ebc4`) has no `Co-Authored-By` trailer for him. Adding one would require rewriting `main`, which isn't worth doing to an already-published branch — two open PRs are based on it. If a git-level trailer matters, the clean route is the follow-up this PR already flagged (rooting `GeneratedHttpEndpointRegistry` on the `Wolverine.Http` side), with `Co-Authored-By: Derek Choate <990568+derekchoate@users.noreply.github.com>` on that commit.

PR #4454, opened 2026-09-16
GH-4453: a SourceServiceFromHttpContext<T> service is declared once per HTTP chain

### LINKED ISSUE #4453: SourceServiceFromHttpContext<T> declares the service twice when two methods of a compound handler take it (CS0128)
**Describe the bug**

Hit while moving to 6 after JasperFx/jasperfx#838. We use `SourceServiceFromHttpContext<IUserContext>()` for the case the docs describe: a scoped `IUserContext` shared with AspNetCore middleware, endpoints take it as a parameter. It breaks on a compound handler where `LoadAsync` and the endpoint method both take `IUserContext`. The generated handler declares the local twice and doesn't compile:

```
error CS0128: A local variable or function named 'userContext' is already defined in this scope
```

`codegen preview` for the endpoint below:

```csharp
public override async System.Threading.Tasks.Task Handle(Microsoft.AspNetCore.Http.HttpContext httpContext)
{
    // This app is configured to source this service from the HttpContext.RequestServices
    var userContext = Microsoft.Extensions.DependencyInjection.ServiceProviderServiceExtensions.GetRequiredService<IUserContext>(httpContext.RequestServices);
    // This app is configured to source this service from the HttpContext.RequestServices
    var userContext = Microsoft.Extensions.DependencyInjection.ServiceProviderServiceExtensions.GetRequiredService<IUserContext>(httpContext.RequestServices);
    System.Diagnostics.Activity.Current?.SetTag("handler.type", "OrderEndpoint");
    ...
```

It's one line away in your own test for #2831: `WidgetEndpoint.Post` in `service_location_assertions.cs` takes `IThing thing`, the sourced type under the name codegen gives it. Add `public static void Before(IThing thing) { }` to that class and `can_use_service_locations_with_http` fails with the same CS0128. The shape itself is one you already test: `Bug3374LoadAsyncAsParametersEndpoints.cs` shares a `DbContext` between `LoadAsync` and `Put`.

Whether it breaks depends on the parameter name. `userContext` is what codegen calls the variable for `IUserContext`; if `LoadAsync` uses that name, any later method taking the type duplicates it, under any name:

| `LoadAsync` parameter | `Delete` parameter | `var userContext` declared |
| --- | --- | --- |
| `userContext` | `userContext` | 2, CS0128 |
| `userContext` | `user` | 2, CS0128 |
| `user` | `userContext` | 1 |
| `user` | `user` | 1 |

So the natural name is the landmine, and renaming one parameter defuses it without anything in the source saying why.

Cause, from the 6.38.0 source: `RequestServicesVariableSource.Create` in `src/Http/Wolverine.Http/CodeGen/RequestServicesFrame.cs` returns `new RequestServicesFrame(type).Service`, so every call is a new frame. JasperFx's `MethodFrameArranger.TryFindVariableByName` calls `Create` on every matching method source for each by-name parameter lookup and keeps the result when the name matches, but never records it in the by-type cache. The next method's lookup, by name or by type, gets another frame. Frame dedup is by reference and can't collapse them. `MessageBusSource` in the same folder already caches its frame for the same reason (#2958).

It first appears in 6.2.1, where the JasperFx 2.2.1 dependency-walk rewrite (JasperFx/jasperfx#390) stopped one same-named variable silently inheriting the other's frame. That rewrite is correct; it just stopped masking this. Reproduced on 6.38.0 / JasperFx 2.69.3.

**To Reproduce**

Web project, `net10.0`, single package `WolverineFx.Http 6.38.0`. No database.

```csharp
using System.ComponentModel.DataAnnotations;
using JasperFx;
using Wolverine;
using Wolverine.Http;

var builder = WebApplication.CreateBuilder(args);

builder.Host.UseWolverine();

// Shared with AspNetCore middleware in the real app
builder.Services.AddScoped<IUserContext, UserContext>();
builder.Services.AddWolverineHttp();

var app = builder.Build();

app.MapWolverineEndpoints(opts =>
{
    opts.SourceServiceFromHttpContext<IUserContext>();
});

return await app.RunJasperFxCommands(args);

public interface IUserContext
{
    string UserId { get; }
}

public class UserContext : IUserContext
{
    public string UserId { get; set; } = "";
}

public record Order(Guid Id);

public static class OrderEndpoint
{
    // Both methods take IUserContext as `userContext`, the name codegen gives the variable.
    public static Task<Order?> LoadAsync(Guid id, IUserContext userContext)
        => Task.FromResult<Order?>(new Order(id));

    [WolverineDelete("/orders/{id}")]
    public static IResult Delete([NotBody] [Required] Order order, IUserContext userContext)
        => Results.NoContent();
}
```

1. `dotnet run -- codegen preview` -> `DELETE_orders_id` has the two `var userContext` lines above.
2. `dotnet run -- codegen write`, then `dotnet build` -> the `CS0128`.
3. Rename the `LoadAsync` parameter to `IUserContext user`, delete `Internal/Generated` (the file from step 2 is now part of the build and blocks `codegen write`), repeat steps 1 and 2 -> one declaration, build succeeds.

With `TypeLoadMode.Dynamic` the same thing surfaces at startup as `InvalidOperationException: Compilation failures! CS0128 ...`.

**Expected behavior**

One `RequestServicesFrame` per chain per sourced type, whatever the parameters are called. `RequestServicesVariableSource` returning the same frame's variable on repeated `Create` calls does it, the way `MessageBusSource` does. I'll open a PR with that change and a regression test in `service_location_assertions.cs`.

AllainPL:
Fixes #4453.

## The bug

A compound HTTP handler whose `LoadAsync` and endpoint method both take a service registered with `SourceServiceFromHttpContext<T>()` gets two `RequestServicesFrame`s in one generated `Handle` method, so the local is declared twice and the chain fails with CS0128. The trigger is the first-resolved method naming the parameter the way codegen names the variable (`IUserContext userContext`); a later method then gets a second frame under any parameter name.

`RequestServicesVariableSource.Create` returned `new RequestServicesFrame(type).Service` on every call, and `MethodFrameArranger.TryFindVariableByName` calls `Create` on each matching method source for every by-name parameter lookup without caching the result by type.

Latent until JasperFx 2.2.1 (Wolverine 6.2.1). The old `DependencyGatherer` cached a variable's creator by `Variable`, whose equality is type plus name, so the second same-named variable inherited the first frame and its own was dropped. The 2.2.1 rewrite (JasperFx/jasperfx#390) reads each variable's own creator, which is correct, and surfaces the duplicate.

## The fix

Cache the frame inside `RequestServicesVariableSource`, the same shape as `MessageBusSource` after #2958. The source is registered per HTTP chain in `HttpChain.Codegen.cs`, so the cache is exactly one generated method. No API or option change.

## Verification

One test in `service_location_assertions.cs`, `compound_handler_sharing_an_http_context_sourced_service_compiles`, a `DELETE` through the real pipeline against a `LoadAsync` + endpoint pair that both take the sourced `IThing thing` and an `AlwaysUseServiceLocationFor` `IFlag flag`. Without the fix the route returns 500 with `CS0128 ... 'thing' is already defined`; with it, 204. CS0128 is a compile error, so the 204 is the one-declaration proof, and the `IFlag` shared the same way shows ordinary service location was never affected.

Full `Wolverine.Http.Tests` with the fix: **1049 passed, 0 failed, 10 skipped** (the pre-existing form-array skips). `dotnet build wolverine.slnx -c Release` succeeds.


jeremydmiller:
@AllainPL Okay, cool. I think this will be in the next Nuget later today

PR #4475, opened 2026-09-17
Only count DUE envelopes when judging the scheduled-job poller stuck

jeremydmiller:
## The problem

The stuck-scheduled-poller health signal reads `PersistedCounts.Scheduled`, which counts **every**
scheduled envelope regardless of whether its execution time has passed:

```csharp
if (counts.Scheduled > 0 && counts.Scheduled >= _previousCounts.Scheduled)
```

A scheduled queue holding envelopes that are not due yet is a scheduled queue doing its job. So any
deliberate delay longer than `check interval × HealthStuckPollCycleThreshold` reports the poller as
stuck, and stays that way. Ordinary retry scheduling has the same shape, which means **the signal
grows with correct usage.**

Measured on a production fleet of 512 sharded message databases (reported downstream by a
CritterWatch user): **254 of 271 active alerts — 94% — were this one check**, spread over 265
databases. 114 of those databases were "degraded" over a **single** envelope, scheduled 15 minutes
out by an application that deliberately waits for a quiet period before acting:

```csharp
new EvaluateAutomaticInsurerClaiming(practiceAgb).DelayedFor(settings.Value.QuietPeriod)
```

It is also a contradiction inside `DurabilityHealthSignals` itself. Signal 1 (persistence
reachability) is driven by what the poll loop **actually did** — `RecordPollSuccess` /
`RecordPollFailure` — while signal 3 infers stuckness from queue depth. The runtime could therefore
know the poller was completing clean cycles and report it stuck anyway.

## The change

`PersistedCounts.ScheduledDue` — scheduled envelopes already **past** their execution time, the ones
the poller should have moved by now. That is the only fact that separates "waiting" from "not
moving", and it is what the health signal now reads.

**`int?`, and null means NOT MEASURED rather than zero.** A store that does not report it makes this
signal stand down rather than falling back to the undifferentiated count — falling back is the
defect. PostgreSQL implements it as a `FILTER` clause on the existing `group by status` scan, so the
due count costs no extra query; every other provider reports null today and is simply silent here.
Silent beats wrong, and the remaining providers are a mechanical follow-up.

## One detail worth a reviewer's eye

The comparison is `execution_time <= now()`, **not** `now() at time zone 'utc'` as
`PostgresqlQueueListener` uses a few files away. `execution_time` is `timestamptz`
(`AddColumn<DateTimeOffset>`), so plain `now()` is an absolute-instant comparison, whereas
`now() at time zone 'utc'` produces a naive timestamp that is coerced back using the server's
`TimeZone` setting.

Demonstrated on PostgreSQL 16 with `timezone='America/Chicago'`:

```
select (timestamptz '2026-09-17 17:00:00+00' <= now())                      -- f  (correct: not due)
     , (timestamptz '2026-09-17 17:00:00+00' <= (now() at time zone 'utc')) -- t  (wrong: says due)
```

An envelope scheduled two hours in the **future** reads as due under the at-time-zone form on any
non-UTC server. That looks like a latent bug in the queue listener; I have deliberately left it
alone rather than change it inside an unrelated fix, but it is worth its own look.

## Tests

Four new tests in `durability_health_signals_tests`, and the pre-existing one renamed to say which
count it is about:

- due-and-not-draining still degrades, and the message says "past their execution time"
- **not due yet stays healthy over six consecutive checks** — the field case
- a store reporting `null` stands the signal down even at 5,000 scheduled
- a draining due backlog resets the consecutive-check counter

Verified by negative control: restoring the undifferentiated count fails exactly those three new
tests and leaves the pre-existing twelve passing. 15/15 green on net9.0 and net10.0.

## Compatibility

`ScheduledDue` is a new nullable property with a null default, so stores that do not set it compile
and behave exactly as before except that this one signal goes quiet for them. No existing behaviour
changes for a store that does report it, other than the false positives disappearing.


jeremydmiller:
Closing in favour of #4476, which carries the same single commit rebased onto main. This branch was cut from a local release-sweep branch, so the PR picked up two unrelated, unpushed commits (13 files instead of 4) — my mistake, and not something to review.

Issue #4480, opened 2026-09-17
Bug: Wolverine.Oracle DurabilityAgent exception during CheckRecoverableIncomingMessagesOperation


Trasvi:
**Describe the bug**
When using Wolverine.Oracle for persistence, the Durability agent throws an exception when trying to recover messages from the incoming table.
```
fail: Wolverine.Runtime.Agents.IAgentCommand[0]
      Invocation of Wolverine.RDBMS.Polling.DatabaseOperationBatch failed!
      Wolverine.RDBMS.Polling.DatabaseBatchCommandException: Database operation batch failure:
1. Wolverine.RDBMS.Durability.CheckRecoverableIncomingMessagesOperation
2. Wolverine.RDBMS.Durability.CheckRecoverableOutgoingMessagesOperation
3. Wolverine.RDBMS.Durability.MoveReplayableErrorMessagesToIncomingOperation
select received_at, count(*) from WOLVERINE.wolverine_incoming_envelopes where status = 'Incoming' and owner_id = 0 group by received_at
       ---> System.InvalidCastException: Unable to cast object of type 'System.Decimal' to type 'System.Int32'.
         at Oracle.ManagedDataAccess.Client.OracleDataReader.GetFieldValueAsyncHelper[T](Int32 ordinal, CancellationToken cancellationToken)
         at Wolverine.RDBMS.Durability.CheckRecoverableIncomingMessagesOperation.ReadResultsAsync(DbDataReader reader, IList`1 exceptions, CancellationToken token) in /home/runner/work/wolverine/wolverine/src/Persistence/Wolverine.RDBMS/Durability/CheckRecoverableIncomingMessagesOperation.cs:line 43
```

This occurs because Oracle returns `COUNT(*)` as a `decimal`.
This is avoided elsewhere because most count queries are executed as scalar and converted using `Convert.ToInt32`. However when the count is just one column of many and read with a reader, GetFieldValueAsync<int> fails.

**To Reproduce**
```
using Microsoft.Extensions.DependencyInjection;
using Microsoft.Extensions.Hosting;
using Wolverine;
using Wolverine.Oracle;
using Wolverine.Persistence.Durability;
using Wolverine.Tracking;
using Wolverine.Transports;
using Wolverine.Util;

var builder = Host.CreateApplicationBuilder(args);

var oracleConnectionString = builder.Configuration["ConnectionStrings:Oracle"];
var schemaName = builder.Configuration["Wolverine:SchemaName"];

[...]

PR #4483, opened 2026-09-18
GH-4480: allow Oracle durability agent to recover incoming messages

Trasvi:
This fixes a bug [4480](https://github.com/JasperFx/wolverine/issues/4480) the Oracle DurabilityAgent that prevented recovery of messages. Oracle returns COUNT(*) columns as a decimal but the datareader was trying to use the basic GetFieldValueAsync<int> method to retrieve the value. This caused an InvalidCastException.

Changed the reader in CheckRecoverableIncomingMessagesOperation to use the `Convert.ToInt32(reader.GetValue(1))` pattern which does not throw the exception.


jeremydmiller:
@Trasvi I'm going to get this into the next release, but also try to add more test coverage on the durability agent across the lesser used databases (oracle, sqlite, mysql) to prevent this kind of thing in the future

